Issue #144, opened 2014-03-15
File rename does not rename file but updates references


as3boyan:
When I confirm to update references it changes class name and it works as expected, but file itself sometimes remain doesn't renamed.

FD 4.5.3.7

Issue #518, opened 2014-12-05
Remove rename-confirmation-popup on rename/move empty folder in sourcepath


markknol:
It would be nice if renaming a folder that is just created does not asks for "You are renaming a directory. Do you also want to update the references?".

Also, if you rename a folder called "DATA" to "data", it gives an error that source and targetpath should differ.

markknol:
This is not fixed yet. When I rename empty folder, it still asks if I want to update references.

markknol:
Thanks!!

Issue #400, opened 2014-09-30
Source folder rename completely messes up classpaths/imports


markknol:
If you rename a source folder in a haxe project, it ask if you want to update references to. If you choose yes, it completely messes up all packages + imports, because it adds the full directory path (like _"C:\mark\projects\secret-game\source\appplication\nl\stroep\games\"_) as package.

I think you can reproduce it by creating a new project and try to rename "src" to "source".

After a while it also gives this error (that source folder contains about 80 classes)
![flashdevelop-rename-error](https://cloud.githubusercontent.com/assets/576184/4454669/7c8852fc-4873-11e4-880e-8f3bd1faf762.jpg)

I feel lucky I did this by accident after a SVN update, so I can safe revert. 

Specs: Windows 7 + FlashDevelop 4.7.0.25 developmnent#cf924f397a

Issue #187, opened 2014-04-13
 AS3 Update references


SMKS:
There seems to be a big problem when package names and imports get updated. Causes duplicates. Crashes the structure of the project. Please fix, very dangerous.

PR #541, opened 2014-12-21
Project tree refactoring fixes

Neverbirth:
Fixed tons of errors and unsupported cases, mainly in the Move command. Should fix issues #144 #162 #187 #400 #518 and several more issues not reported.
Rename command left modified files unsaved, while Move command saved them, the latter is the common approach taken, and I'd say seeing tons of files with unsaved changes may confuse the user, so made Rename to behave the same.
The Move command tried to keep modified files open, but the code was wrong, I fixed the initial check but left it commented, I think it pollutes the user interfaces when the refactoring affects many files. If this feature is wanted it would require some further changes.
When the Move and Rename commands finish, the affected files that were initially open are kept open, although the position is lost.
Removed calls to RefactoringHelper.GetSearchPatternFromLang and use Project.DefaultSearchExtension instead, that way everything is centralized, this allowed to see the default search filter of Haxe projects was missing the recently added .hxp files (confirmed to be an oversight), and made finding references able to find matches in code sections of MXML files (no extensively tested).
Made folder-package check case-sensitive, as it's the right thing to do.

### DIFF
--- a/External/Plugins/AS2Context/Context.cs
+++ b/External/Plugins/AS2Context/Context.cs
@@ -795,7 +795,7 @@ namespace AS2Context
                 {
                     string pathname = package.Replace('.', Path.DirectorySeparatorChar);
                     string fullpath = Path.GetDirectoryName(cFile.FileName);
-                    if (!fullpath.ToUpper().EndsWith(pathname.ToUpper()))
+                    if (!fullpath.EndsWith(pathname))
                     {
                         if (settings.FixPackageAutomatically && CurSciControl != null)
                         {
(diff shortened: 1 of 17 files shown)


elsassph:
Using overrides is a common practice when working with libraries at least - at work we are using it a lot. FD handles that by following the compiler rule, ie. the first match wins (excepted in Haxe where the class path is reversed because that's the last match).
However I wouldn't expect refactoring to rename all the copies - hopefully it just doesn't do bad things :)

elsassph:
Also happy new year! ;)

Neverbirth:
> However I wouldn't expect refactoring to rename all the copies - hopefully it just doesn't do bad things :)

Indeed, it shouldn't work that way, sadly it does heh. For someone that was overriding a class in a library it won't be a problem tho.

I've made a simple test, well, I can see a relatively easy way of supporting this bizarre case, at least for moving, I'd have to check renaming.

Happy New Year to you too!

PR #544, opened 2014-12-23
More debugging improvements

Neverbirth:
Fixed breakpoint grid, it was being displayed below the new actions toolstrip.
Fixed and simplified methods for activating documents and going to specific lines (for example, when going to some breakpoint or a stack frame).
Fixed getting files locations when using a source path not relative to the project file, and look for files in global classpaths as well.
New features for the stackframe panel: copy lines, hide external code and filtering.


External/Plugins/FlashDebugger/Debugger/DebuggerManager.cs
```diff
@@ -269,14 +268,13 @@ public String GetLocalPath(SourceFile file)
                     return localPath;
                 }
             }
-            Project project = PluginBase.CurrentProject as Project;
+            IProject project = PluginBase.CurrentProject;
             if (project != null)
             {
-                foreach (string cp in project.Classpaths)
+                foreach (string cp in project.SourcePaths.Concat(ProjectManager.PluginMain.Settings.GetGlobalClasspaths(project.Language))
```
elsassph: For readability please extract the collection in a variable.

External/Plugins/FlashDebugger/Controls/StackframeUI.cs
```diff
@@ -40,81 +166,346 @@ public StackframeUI(PluginMain pluginMain, ImageList imageList)
...
         {
+            wholeFrameStack.Clear();
             lv.Items.Clear();
+            toolStripTextBoxFilter.Enabled = lv.Enabled = false;
         }
 
         public void ActiveItem()
         {
-            foreach (ListViewItem item in lv.Items)
+            lv.Items[lastSelected].ImageIndex = -1;
```
elsassph: You may want to check that the index is valid.

External/Plugins/FlashDebugger/Controls/StackframeUI.cs
```diff
@@ -40,81 +166,346 @@ public StackframeUI(PluginMain pluginMain, ImageList imageList)
...
+            int index = PluginMain.debugManager.CurrentFrame;
+            var selectedItem = wholeFrameStack[index];
```
elsassph: Idem

External/Plugins/FlashDebugger/Controls/StackframeUI.cs
```diff
@@ -40,81 +166,346 @@ public StackframeUI(PluginMain pluginMain, ImageList imageList)
...
+                    if (itemData.Index < index) continue;
+                    if (itemData.Index == index)
+                    {
+                        lv.Items[i].ImageIndex = currentImageIndex;
+                        lastSelected = i;
+                        break;
+                    }
+                    if (itemData.Index > index)
+                    {
+                        lv.Items[i - 1].ImageIndex = currentImageIndex;
```
elsassph: Can `i-1 < 0`?

External/Plugins/FlashDebugger/Controls/StackframeUI.cs
```diff
@@ -40,81 +166,346 @@ public StackframeUI(PluginMain pluginMain, ImageList imageList)
...
+                    {
+                        title += " at " + sourceFile + ":" + item.getLocation().getLine();
+                        if (project != null)
+                        {
+                            foreach (string cp in project.AbsoluteClasspaths)
+                            {
+                                string pathBackSlash = cp.TrimEnd(new[]{Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar});
+                                pathBackSlash = pathBackSlash.IndexOf(Path.AltDirectorySeparatorChar.ToString()) > -1 ?
+                                    pathBackSlash + Path.AltDirectorySeparatorChar : pathBackSlash + Path.DirectorySeparatorChar;
+                                if (sourceFile.getFullPath().ToString().StartsWith(pathBackSlash))
```
elsassph: Maybe do a case insensitive comparison?

External/Plugins/FlashDebugger/Controls/StackframeUI.cs
```diff
@@ -40,81 +166,346 @@ public StackframeUI(PluginMain pluginMain, ImageList imageList)
...
+                                }
+                            }
+                        }
+                    }
+                    var listItem = new ListViewItem(new[] {string.Empty, title}, -1)
+                        {
+                            Tag = new ListItemData {Frame = item, Index = i++}
+                        };
+                    listItem.UseItemStyleForSubItems = false;
+                    // TODO: Apply proper theming colour
```
elsassph: @Meychi can you help?

External/Plugins/FlashDebugger/Controls/StackframeUI.cs
```diff
@@ -40,81 +166,346 @@ public StackframeUI(PluginMain pluginMain, ImageList imageList)
...
+
+        private void ContextMenuOpening(Object sender, System.ComponentModel.CancelEventArgs e)
+        {
+            copyContextMenuItem.Enabled = setFrameContextMenuItem.Enabled = gotoSourceContextMenuItem.Enabled = lv.SelectedItems.Count > 0;
+            copyAllContextMenuItem.Enabled = lv.Items.Count > 0;
+        }
+
+        private void CopyTextClick(Object sender, EventArgs e)
+        {
+            Clipboard.SetText(lv.SelectedItems[0].SubItems[1].Text);
```
elsassph: Is there a risk SelectedItems can be null/empty?

External/Plugins/FlashDebugger/Controls/StackframeUI.cs
```diff
@@ -40,81 +166,346 @@ public StackframeUI(PluginMain pluginMain, ImageList imageList)
...
+                        if (((ListItemData)item.Tag).Index == currentFrame)
+                        {
+                            item.ImageIndex = currentImageIndex;
+                            lastSelected = lv.Items.Count - 1;
+                        }
+                    }
+                }
+            }
+
+            wholeFrameStack[PluginMain.debugManager.CurrentFrame].ImageIndex = currentImageIndex;
```
elsassph: This kind of array access without bounds check make me nervous ;)

External/Plugins/FlashDebugger/Controls/StackframeUI.cs
Neverbirth: No, the control is not enabled if SelectedItems is empty.

External/Plugins/FlashDebugger/Controls/StackframeUI.cs
Neverbirth: At least as code currently stands there won't be indexes out of range, but I'll make more tests.

External/Plugins/FlashDebugger/Controls/StackframeUI.cs
Neverbirth: That's a bit subjective, since packages and paths should be case sensitive I'd say it shouldn't, also since we want to check for our own files, there shouldn't be casing problems. I guess that if an user adds some source path, and then renames it, the problem could pop up, but then the project would violate case sensitivity.

I know casing doesn't matter in Windows, but does in other platforms, so I like to take it into account.

@Meychi?

elsassph: Windows (and Mac by default) paths aren't case sensitive and classpaths can sometimes be configured with incorrect casing. That has been a recurring problem with the debugger.

PR #553, opened 2015-01-06
fixes: Displays the status of files under version control system

SlavaRa:
Before fix:

After:

### DIFF
--- a/External/Plugins/SourceControl/Actions/ProjectWatcher.cs
+++ b/External/Plugins/SourceControl/Actions/ProjectWatcher.cs
@@ -14,24 +14,17 @@ namespace SourceControl.Actions
 {
     public class ProjectWatcher
     {
-        private static bool initialized = false;
-
-        static internal readonly List<IVCManager> VCManagers = new List<IVCManager>();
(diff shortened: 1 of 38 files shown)


elsassph:
I'm sorry to repeat it, but can you keep code cleanup and fixes in separated PRs please?

SlavaRa:
Ok

elsassph:
I essentially don't agree with most of the stylistic changes you did: 

- I like whitespace; blank lines are here for code respiration,
- I like to move `return` on the next line as it makes it more "visible" in the code,
- I don't add `private` like in Haxe because that's the default http://stackoverflow.com/questions/3763612/default-visibility-for-c-sharp-classes-and-members-fields-methods-etc.

I'm ok with cleaning up `using` statements, fixing inconsistent whitespace, removing dead code, fixing incorrect `public` visibility.

Neverbirth:
I like adding the modifier, first, it makes clear the developer intention and second, it's clearer and faster for people reading the code, as the link shows, the default visibility varies (even between .NET languages, for example the default in these cases would be internal or public in other .NET languages).

PR #555, opened 2015-01-07
Fix for crash in some incorrectly defined metadata tags

Neverbirth:
Wouldn't have committed this to the development branch if it were not because it fixes some possible nasty crashes, and FD 4.7 is not released yet.

EDIT: For example the ones fixed here: https://github.com/apache/flex-flexunit/commit/1cf55d524f47affdb0c4e0e60d511d39e3d6e8b2

BTW, I was the one to add this crash while fixing other wrong cases and add some unsupported ones.


elsassph:
I don't see what crash it prevents.

Neverbirth:
It crashes in [Event(type="testsComplete", type="flash.events.Event")], because it tries to add two elements with the same "type" key, using the default property [key] = value, doesn't give a duplicate key exception like the Add method.

PR #557, opened 2015-01-09
Cleaned up some ugly code with LINQ

Neverbirth:


External/Plugins/CodeRefactor/Provider/RefactoringHelper.cs
```diff
@@ -369,25 +370,13 @@ private static List<String> GetAllProjectRelatedFiles(IProject project, bool onl
             }
             else
             {
-                // NOTE: Could be simplified in .NET 3.5 with LINQ .Concat.Select.Distinct
-                var visited = new Dictionary<string, byte>();
-                foreach (string path in project.SourcePaths)
+                foreach (string path in project.SourcePaths.
```
elsassph: For readability you should move the link part out of the loop.

elsassph:
We have zero usage of Linq so far - how much of an overhead are we adding?

Neverbirth:
That's pretty relative, some LINQ queries are insignificantly slower, others are several times slower, The readability and simplicity gains are generally well above performance, but in this case I'd say the overhead is minimal, plus it's code that executes only per user request, so there shouldn't be any need to worry.

External/Plugins/CodeRefactor/Provider/RefactoringHelper.cs
Neverbirth: I prefer it this way, when I see a variable I think "this is used in more the one place", but I don't believe it matters that much nor my reasoning is any better, so I'll change it.

PR #558, opened 2015-01-09
Check for file overwrite on moving files before doing any refactoring

Neverbirth:


elsassph:
What is the behavior if no class matching the file name is found in the file? Falls back to simple file move?

Neverbirth:
In that case the user's computer will explode, killing all the house residents in the process, all because of writing bad code.

elsassph:
Oh then that's ok ;)

PR #560, opened 2015-01-13
SourceControl cleanup

SlavaRa:
Only cleanup code


elsassph:
I don't understand... In the previous PR I answered that I did not agree with the code style changes, and here they are.

I'm going to merge it because I respect the contribution effort, but please do not repeat that on another plugin.

SlavaRa:
Apparently I misunderstood a comment, sorry((.
I only have to do PR with an error-correcting code?

SlavaRa:
I found explaining comment, remade

PR #561, opened 2015-01-13
SourceControl cleanup

SlavaRa:


External/Plugins/SourceControl/Sources/Git/Status.cs
```diff
@@ -90,10 +91,10 @@ override protected void runner_ProcessEnded(object sender, int exitCode)
             if (OnResult != null) OnResult(this);
         }
 
-        override protected void runner_Output(object sender, string line)
+        override protected void Runner_Output(object sender, string line)
         {
             int fileIndex = 3;
-            if (line.Length < fileIndex || line.Length < 3) return;
```
elsassph: Isn't checking the length preferable, considering the fact that we're accessing `line[1]`?

PR #568, opened 2015-01-18
Treeview flicker

Neverbirth:
Some more work trying to avoid flickering.
Also, optimized the creation and update of the outline.

### DIFF
--- a/External/Plugins/ASCompletion/PluginUI.cs
+++ b/External/Plugins/ASCompletion/PluginUI.cs
@@ -579,6 +579,8 @@ namespace ASCompletion
         {
             //TraceManager.Add("Outline refresh...");
             outlineTree.BeginStatefulUpdate();
+            if (prevChecksum.StartsWith(aFile.FileName))
+                aFile.OutlineState = outlineTree.State;
 
             try
             {
--- a/External/Plugins/ASCompletion/PluginUI.cs
+++ b/External/Plugins/ASCompletion/PluginUI.cs
@@ -83,6 +83,7 @@ namespace ASCompletion
 
         private GeneralSettings settings;
         private string prevChecksum;
+        private string prevLines;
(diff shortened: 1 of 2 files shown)


External/Plugins/ASCompletion/PluginUI.cs
```diff
@@ -673,6 +679,138 @@ private void RefreshView(FileModel aFile)
...
+        private void UpdateTree(FileModel aFile)
+        {
+            try
+            {
+                if (aFile == FileModel.Ignore)
+                    return;
+
+                var classes = new Dictionary<string, ClassModel>();
+                foreach (var classModel in aFile.Classes)
+                    classes[classModel.Name] = classModel;
```
elsassph: In principle it would be possible for a file to declare 2 classes with the same name. Especially in Haxe when you might be copy-pasting declarations in one file before doing some renaming.

Neverbirth:
I made some changes over the weekend, I don't know if you were noticed about them.

elsassph:
I merged after checking your changes, let's give a good try :)

Neverbirth:
BTW, although FileModel.OutlineState isn't used anywhere else, it has effect when switching files.

elsassph:
What effect? 
BTW I'd like to remove FileModel.OutlineState completely but I didn't want to delay this merge for that.

Neverbirth:
I meant that although FileModel.OutlineState isn't used anywhere else, and it seems it doesn't have any use, it turns out that when switching files their last outline tree state is now restored.

If you want to remove it I see two options:

1.- Forget about other files and care only for the current file. As far as I've seen in the GitHub history, state wasn't correctly restored until now, so there is no real loss.

2.- Use a Dictionary in PluginUI to keep the state of other files. This would require more code to manage the data kept in it.

elsassph:
Yes I understand that.

PR #574, opened 2015-01-21
ASClassWizard work

Neverbirth:
Some cleanup, and more importantly, fixed behaviour on new file conflicts.

### DIFF
--- a/External/Plugins/ASClassWizard/PluginMain.cs
+++ b/External/Plugins/ASClassWizard/PluginMain.cs
@@ -32,8 +32,8 @@ using System.Collections.Generic;
 
 namespace ASClassWizard
 {
-	public class PluginMain : IPlugin
-	{
+    public class PluginMain : IPlugin
+    {
         private String pluginName = "ASClassWizard";
         private String pluginGuid = "a2c159c1-7d21-4483-aeb1-38d9fdc4c7f3";
         private String pluginHelp = "www.flashdevelop.org/community/";
@@ -47,7 +47,7 @@ namespace ASClassWizard
         private String constructorArgs;
         private List<String> constructorArgTypes;
 
-	    #region Required Properties
+        #region Required Properties
         
         /// <summary>
         /// Api level of the plugin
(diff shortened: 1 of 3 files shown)


External/Plugins/ASClassWizard/PluginMain.cs
```diff
@@ -227,28 +227,37 @@ private void DisplayClassWizard(String inDirectory, String templateFile, String
...
+                    if (File.Exists(newFilePath))
+                    {
+                        string title = " " + TextHelper.GetString("FlashDevelop.Title.ConfirmDialog");
+                        string message = TextHelper.GetString("PluginCore.Info.FolderAlreadyContainsFile");
+                        conflictResult = MessageBox.Show(PluginBase.MainForm,
+                            string.Format(message, newFilePath, "\n"), title,
+                            MessageBoxButtons.YesNoCancel, MessageBoxIcon.Warning);
+                        if (conflictResult == DialogResult.No) return;
+                    }
+                } while (conflictResult == DialogResult.Cancel);
```
elsassph: What the trick? The messagebox is presented until the user chooses Yes or No only?

Meychi: The logic is kinda hard to pickup from this code.

Neverbirth: Before, the logic was:

Show wizard dialog.
If the dialog is accepted:
  - Get the new file name.
  - Check for conflict, if conflict:
       - Display a YesNoCancel alert, if the user pressed Cancel the function returned, otherwise the file was overwritten.

Since presenting a YesNoCancel message box is standard, and the cancel option shouldn't cause any data loss, selecting it displays the dialog again with all fields filled. I think the old option of selecting a YesNoCancel dialog was quite appropiate, just Yes and No would made the user to launch and fill the form again.

I wouldn't say it's doing anything fancy, it's just wrapping the minimal code needed in a do while loop, and also checking for the No option that was before ignored, the rest of the code is the same as before.

Meychi: Ah ok, that sounds good to me.

elsassph: Gotcha.

PR #605, opened 2015-03-02
Copy Tree command and combine inherited (#2)

hgupta9:
1. Add "Copy Tree" and "Copy ID" commands to DataTreeControl
2. Hide full class paths ("flash.display.MovieClip" shows as "MovieClip")
3. List inherited vars alongside main vars of a class instance
4. Use FD4 code styling (braces on new lines)
5. Add comments for String utilis (StringsHelper)

### DIFF
--- a/External/Plugins/FlashDebugger/Controls/DataTreeControl.cs
+++ b/External/Plugins/FlashDebugger/Controls/DataTreeControl.cs
@@ -284,7 +284,8 @@ namespace FlashDebugger.Controls
 				if (inherited.Count > 0)
 				{
 					// list inherited alongside main class members
-					foreach (DataNode item in inherited) {
+					foreach (DataNode item in inherited)
+					{
 						node.Nodes.Add(item);
 					}
 
@@ -459,8 +460,10 @@ namespace FlashDebugger.Controls
 		// ADDED
 		#region Copy ID & Tree
 
-		private void CopyItemIDClick(Object sender, System.EventArgs e) {
-			if (Tree.SelectedNode != null) {
+		private void CopyItemIDClick(Object sender, System.EventArgs e)
+		{
+			if (Tree.SelectedNode != null)
+			{
 
 				DataNode node = Tree.SelectedNode.Tag as DataNode;
 				Clipboard.SetText(node.ID);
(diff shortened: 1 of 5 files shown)


Meychi:
To me this looks good, needs to be localized though. Before doing anything else, let us review this relating to our plans and current works on FD5.

hgupta9:
I will re-submit this PR against the fd5 branch, after whatever changes are made.

External/Plugins/FlashDebugger/Controls/DataTreeControl.cs
```diff
@@ -20,6 +20,8 @@ public partial class DataTreeControl : UserControl, IToolTipProvider
         private ToolStripMenuItem copyMenuItem, viewerMenuItem, watchMenuItem;
 		private List<String> expandedList = new List<String>();
 		private bool watchMode;
+		private ToolStripMenuItem copyIDMenuItem;
```
elsassph: For consistency, keep `ToolStripMenuItem` vars declarations together.

External/Plugins/FlashDebugger/Controls/DataTreeControl.cs
```diff
@@ -243,120 +247,142 @@ void TreeExpanding(Object sender, TreeViewAdvEventArgs e)
...
-						node.Nodes.Add(staticNode);
+						nodes.Add(memberNode);
+					}
+				}
+
+				// inherited vars
+				if (inherited.Count > 0)
+				{
+					// list inherited alongside main class members
+					foreach (DataNode item in inherited)
```
elsassph: This is a change in behaviour - can we keep the `inherited` group?

hgupta9: Can we add an option in FD options to let the dev choose which he wants? I personally find listing inherited side-by-side very useful, since it lets me see properties I care about without going into [inherited] group.

External/Plugins/FlashDebugger/Controls/DataTreeControl.cs
hgupta9: Done.

External/Plugins/FlashDebugger/Controls/DataTreeControl.cs
elsassph: You should suggest that as another PR, ideally providing the setting property. It should be like a property that the user has to set to "true", like "Combine inherited properties in debugger inspectors" so the behaviour defaults to the current behaviour.

hgupta9: Done.

hgupta9:
Changes done. Please see PR https://github.com/fdorg/flashdevelop/pull/608

PR #607, opened 2015-03-03
SWC const value extraction fix

### LINKED ISSUE #606: Auto Complete for custom events with event metadata
**Issue:**
Instead of listing a custom event, auto complete offers up the string value of the event's constant.

**Steps to replicate: **
1. Create a custom class that dispatches a custom event
2. Add the correct event metadata for the custom events to the custom class
3. Build a SWC including the custom class/event including ASDocs
4. Add the SWC to a project's library
5. Attempt to list the custom events via auto complete on addEventListener

**Custom Event:**
```actionscript
public static const POLLER_UPDATED:String = "pollerUpdated";
public static const POLLER_ERROR:String = "pollerError";
```
**Event metadata:**
```actionscript
/**
 * mm.utils.load.PollerEvent.POLLER_UPDATED
 * @eventType mm.utils.load.PollerEvent
 */
[Event(name = "pollerUpdated", type = "mm.utils.load.PollerEvent")]
/**
 * mm.utils.load.PollerEvent.POLLER_ERROR
 * @eventType mm.utils.load.PollerEvent
 */
[Event(name="pollerError", type="mm.utils.load.PollerEvent")]
```
**Auto complete:**
Auto complete

Github doesn't support attaching SWCs, so here's a link to the sample SWC on google drive: https://drive.google.com/file/d/0B-E-NS9ouP-ANExTd0dleHJ1QkU/view?usp=sharing (need to click the download button because Google Drive recognizes the file as an archive).

elsassph:
Fixes #606 : force extract const values provided in fat SWCs even if no doc is specified.

### DIFF
--- a/External/Plugins/AS3Context/AbcConverter.cs
+++ b/External/Plugins/AS3Context/AbcConverter.cs
@@ -627,12 +627,14 @@ namespace AS3Context
     {
         public bool IsFinal = false;
         public bool IsDynamic = false;
+        public bool IsStatic = false;
 
         public string ShortDesc = null;
         public string LongDesc = null;
         public string Returns = null;
         public string Value = null;
         public string ApiType = null;
+        public string DeclType = null;
 
         public List<ASMetaData> Meta;
         public Dictionary<string, string> Params = new Dictionary<string, string>();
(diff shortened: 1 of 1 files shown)


Meychi:
Should this be added to FD4 too?

elsassph:
I forgot a little thing and we can apply to FD4 as well.

PR #608, opened 2015-03-03
Copy Tree command and combine inherited (#3)

hgupta9:
1. Add "Copy Tree" and "Copy ID" commands to DataTreeControl
2. Hide full class paths ("flash.display.MovieClip" shows as "MovieClip")
3. List inherited vars alongside main vars of a class instance
4. Add property in DataTreeControl to toggle this behavior
5. Use FD4 code styling (braces on new lines)
6. Add comments for String utilis (StringsHelper)

### DIFF
--- a/External/Plugins/FlashDebugger/Helpers/CopyAsTree.cs
+++ b/External/Plugins/FlashDebugger/Helpers/CopyAsTree.cs
@@ -5,8 +5,10 @@ using System.Text;
 using Aga.Controls.Tree;
 using FlashDebugger.Controls;
 
-namespace FlashDebugger {
-	public class CopyTreeHelper {
+namespace FlashDebugger
+{
+	public class CopyTreeHelper
+	{
 
 		public const int CopyTreeMaxRecursion = 10;
 		public const int CopyTreeMaxChars = 1000000;
@@ -14,7 +16,8 @@ namespace FlashDebugger {
 		private static Dictionary<string, int> CopiedObjects;
 
 
-		public static string GetTreeAsText(TreeNodeAdv treeNode, DataNode dataNode, string levelSep, DataTreeControl control, int levelLimit) {
+		public static string GetTreeAsText(TreeNodeAdv treeNode, DataNode dataNode, string levelSep, DataTreeControl control, int levelLimit)
+		{
 			StringBuilder sb = new StringBuilder();
 
 			// ensure expanded
@@ -27,17 +30,20 @@ namespace FlashDebugger {
 			return sb.ToString() ?? "";
 		}
 
-		private static void GetTreeItemsAsText(IList treeNodes, IList dataNodes, string levelSep, int level, StringBuilder sb, DataTreeControl control, int levelLimit) {
+		private static void GetTreeItemsAsText(IList treeNodes, IList dataNodes, string levelSep, int level, StringBuilder sb, DataTreeControl control, int levelLimit)
+		{
 
 			// per node
 			int len = dataNodes.Count;
-			for (int c = 0; c < len; c++) {
+			for (int c = 0; c < len; c++)
+			{
 				DataNode child = (DataNode)dataNodes[c];
 				TreeNodeAdv childTree = (TreeNodeAdv)treeNodes[c];
 
 
 				// skip if unwanted item
-				if (!IsWantedChild(child)) {
+				if (!IsWantedChild(child))
+				{
 					continue;
 				}
 
@@ -50,7 +56,8 @@ namespace FlashDebugger {
 				sb.Append(child.Text + " : " + child.Value);
 
 				// recurse for children .. but skip if unwanted items
-				if (child.Nodes.Count > 0 && IsWantedParent(child)) {
+				if (child.Nodes.Count > 0 && IsWantedParent(child))
+				{
 
 					// opening brace
 					sb.AppendLine("{");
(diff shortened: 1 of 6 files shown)


hgupta9:
For some reason Meychi's change also got added into this. Its a minor change so when this PR is being merged it can be ignored or whatever. 

@elsassph : I hope everything looks ok now?

Meychi:
There is now a conflict and we cannot merge this automaticly. Try to remove my commit from the PR: http://stackoverflow.com/questions/20765030/git-remove-a-commit-from-pull-request

hgupta9:
I'll make another PR, but just check if the code is ok first. I'll make all changes into one final PR.

Meychi:
To me it looks ok, we need to localize it and add a setting to enable it. Is this ok to merge folks (@elsassph, @Neverbirth)?

hgupta9:
Will "Copy Tree" be disabled by default? Or do you mean add a setting for the "combineInherited" property?

Meychi:
It will default to old behaviour but we will add a setting to the debugger to enable "combineInherited".

hgupta9:
Great. Thanks for that. I'll upload the final version in a new PR if everyone is ok with it. I had to modify it for FD5, eg `DataNode` is now `ValueNode`, among other things.

hgupta9:
@elsassph, @Neverbirth : Are you guys okay with this? Shall I submit the final PR? Any code changes?

Neverbirth:
I think you should close this PR and make the new one from fd5, there are always chances something slips away when porting to fd5 and that makes us having to review twice the changes (which also rises the chances of us missing something).

Out of the things I saw, although some may be a bit subjective:

* Node Value getter behaviour has changed, I think the previous one was better as it showed more information (some of it arguably useful, but all of it had its purpose).
* I know having to update each instance of a class may sometimes be a bit problematic, but I don't feel that DataTreeControl.CombineInherited should be static. If that's a problem, the setting could always be accessed by PluginMain. Anyway I'm aware the setting doesn't exist yet.
* CopyTreeHelper could also see some improvements, I don't think having that static CopiedObjects Dictionary is good, and some of the filtering made by looking for some string values could clash with custom classes, although the chances are slim, they are there.

hgupta9:
@Neverbirth : Good points, all of them. As I said before I have already edited the code for FD5.

I'll answer each point:

- Node Value
 - Functions which showed a meaningless int like "234324" now show "Function"
 - Classpaths now are hidden so "flash.display.MovieClip" shows "MovieClip"
 - No other changes (the pointer ID after @ is still there, eg. "MovieClip@fab342")
- CombineInherited should be static - Whatever, can be changed by the person who adds the setting. I didn't know how settings work
- You'll have to elaborate on the clashing point. I couldn't see any clashes.

Meychi:
All that int info and classpaths might be needed info even though i agree it looks crap.
Static values are shared by all nodes even if some nodes should not use it.
The setting and menu item additions are not trivial, they need localization too.

I will close this PR, please create an issue where the details can be discussed before reimplementation as this PR needs to be reworked from FD5 branch.

Please do not take this badly, we do appreciate all the work you've put into this. We just want important changes to be done well and very well thought out and agreed on.

Neverbirth:
I don't agree on removing the classpath, I find it informative, and it helps removing ambiguities, plus a lot of IDEs display it as well: Visual Studio, IntelliJ IDEA, Flash Builder... so it may not be that undesirable.

External/Plugins/FlashDebugger/Helpers/CopyAsTree.cs
```diff
@@ -0,0 +1,212 @@
...
+
+namespace FlashDebugger
+{
+	public class CopyTreeHelper
+	{
+
+		public const int CopyTreeMaxRecursion = 10;
+		public const int CopyTreeMaxChars = 1000000;
+
+		private static Dictionary<string, int> CopiedObjects;
```
Neverbirth: CopyTreeHelper.CopiedObject is static, and right now after the function exits it isn't freed, I think it would be better to just pass it as an argument to GetTreeItemsAsText, it's only used there.

External/Plugins/FlashDebugger/Helpers/CopyAsTree.cs
```diff
@@ -0,0 +1,212 @@
...
+		private static bool IsWantedParent(DataNode parent)
+		{
+			try
+			{
+
+				// if is an empty array []
+				// then skip it from opening and closing { } the output
+				if (parent.Nodes.Count == 1 || parent.Nodes.Count == 2)
+				{
+					if (parent.Value.StartsWith("Array"))
```
Neverbirth: What if the user has some class with the name starting as Array?

External/Plugins/FlashDebugger/Helpers/CopyAsTree.cs
```diff
@@ -0,0 +1,212 @@
...
+					{
+						if (Array.IndexOf(as3DisabledProps, child.Text) > -1)
+						{
+							return false;
+						}
+					}
+
+					// if is an AS2 display object,
+					// don't go upward (_parent)
+					if (parent.Value == "MovieClip" || parent.Value == "Video")
```
Neverbirth: As before, the user could have a custom class named MovieClip, Video, Main, define classes inside flash.display, etc and have wanted variables named like the ones being filtered out.

As I said, chances are slim, but they are there.

hgupta9:
I hope you guys are okay with this PR https://github.com/fdorg/flashdevelop/pull/616

External/Plugins/FlashDebugger/Helpers/CopyAsTree.cs
hgupta9: Fixed using full classpaths. (`ValueNode.ClassPath`)

External/Plugins/FlashDebugger/Helpers/CopyAsTree.cs
hgupta9: Fixed using full classpaths. (`ValueNode.ClassPath`)

External/Plugins/FlashDebugger/Helpers/CopyAsTree.cs
hgupta9: CopiedObject is freed at the end of the copying. If you know a way to forcibly dispose it, write back.

Issue #563, opened 2015-01-14
Colored highlighting of lines with keywords in Output panel.


grapefrukt:
![screenshot 2015-01-14 16 19 25](https://cloud.githubusercontent.com/assets/158901/5741138/b7fcda76-9c09-11e4-92d6-b36a852f841f.png)
*(red and green lines are added by me as an example)*

Add a list of configurable keywords that will be highlighted in the Output panel. 

 As best I can tell, there's support for colored lines already, so this shouldn't be much of a stretch. 
Either it could be a list of keyword/color pairs, or more easily, make use of the five categories already present (Info, Debug, Warning, Error, Fatal).

Sensibly, the presence of such a word would highlight that entire line. 

I've started working on adding this myself but figured I'd make an issue for discussing if this is a good idea at all. This could also be a solution to issue https://github.com/fdorg/flashdevelop/issues/530.

PR #610, opened 2015-03-03
Colored highlighting of lines with keywords in Output panel. 

grapefrukt:
As described in https://github.com/fdorg/flashdevelop/issues/563


Meychi:
Thank you grapefrukt!

We like the idea behind this feat but we think another implementation would be better and and the same time get rid of the current int coloring as it sometimes causes undesired partial removals of traces among another issues.

We need to think this thru little more but we will address this in FD5.

Meychi:
Done to FD5.

PR #632, opened 2015-03-23
Exclude paths under SDK from CodeRefactor...

Meychi:
This works as a good base filter but with Haxe we should also filter out custom haxelibs.

### DIFF
--- a/External/Plugins/CodeRefactor/PluginMain.cs
+++ b/External/Plugins/CodeRefactor/PluginMain.cs
@@ -317,9 +317,9 @@ namespace CodeRefactor
                 ASResult result = isValid ? resolved.Result : null;
                 if (result != null && !result.IsNull())
                 {
-                    bool isRenameable = (result.Member != null && result.Member.InFile != null && File.Exists(result.Member.InFile.FileName) && !RefactoringHelper.IsUnderSDKPath(result.Member.InFile.FileName))
-                        || (result.Type != null && result.Type.InFile != null && File.Exists(result.Type.InFile.FileName) && !RefactoringHelper.IsUnderSDKPath(result.Type.InFile.FileName))
-                        || (result.InFile != null && File.Exists(result.InFile.FileName) && !RefactoringHelper.IsUnderSDKPath(result.InFile.FileName))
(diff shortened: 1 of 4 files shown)


External/Plugins/CodeRefactor/PluginMain.cs
```diff
@@ -322,15 +311,15 @@ private void UpdateMenuItems()
...
                 bool isValid = langIsValid && resolved != null && resolved.Position >= 0;
                 ASResult result = isValid ? resolved.Result : null;
                 if (result != null && !result.IsNull())
                 {
-                    bool isRenameable = (result.Member != null && result.Member.InFile != null && File.Exists(result.Member.InFile.FileName))
-                        || (result.Type != null && result.Type.InFile != null && File.Exists(result.Type.InFile.FileName))
-                        || (result.InFile != null && File.Exists(result.InFile.FileName))
+                    bool isRenameable = (result.Member != null && result.Member.InFile != null && File.Exists(result.Member.InFile.FileName) && !RefactoringHelper.IsUnderSDKPath(result.Member.InFile.FileName))
+                        || (result.Type != null && result.Type.InFile != null && File.Exists(result.Type.InFile.FileName) && !RefactoringHelper.IsUnderSDKPath(result.Type.InFile.FileName))
+                        || (result.InFile != null && File.Exists(result.InFile.FileName) && !RefactoringHelper.IsUnderSDKPath(result.InFile.FileName))
```
elsassph: `IsUnderSDKPath` could take care of verifying if the value isn't null and exists - that would be more readable. Also you could have a little function that accepts a `FileModel`

Meychi: Is it better now?

Meychi:
I just messed up my libs by renaming builtin types, i had to try to see it in action :) I'm so merging this right now!

elsassph:
Hehe ok, but I was suggesting to regroup this kind of long test `(RefactoringHelper.ModelFileExists(result.InFile) && !RefactoringHelper.IsUnderSDKPath(result.InFile))` into one function. Merge it wh

PR #635, opened 2015-03-30
Added Plugins to AppMan.xml

JoeRobich:
Adds NavigationBar, EditorMiniMap, CustomizeToolbar, QuickLaunch, and TrackOpenFile extensions to AppMan.


Meychi:
Great, thanks!

Plugins need to be put into the commented extras section which is only for FD5. Older FD4 releases may not work correctly so these will be supported only in FD5. I can do it myself as I think I need to rename the section comments to be clearer. Also I will enable multiple extra groups for the future.

JoeRobich:
I moved my plugin entries into the Extra's section.

PR #636, opened 2015-03-30
Offer to save the syntax for all files with this extension when changing the syntax, closes #627

### LINKED ISSUE #627: Automatically Remember syntax coloring association
For example, my .hxt files are actually regular xml. I think any user would benefit that when they switch one .xml alike file, all other files from the same extensions gets colored the same way.

This way when i switche my .hxt  to xml syntax coloring once, it gets retained for all my .hxt.

This would change my experience forever :)

Gama11:


Meychi:
Great, thanks! I will check this later today.

PR #638, opened 2015-03-31
Update appman.xml to remove TrackOpenFile.

JoeRobich:
TrackOpenFile extension is no longer needed now that ProjectManager has the Track Active File option.

### DIFF
--- a/appman.xml
+++ b/appman.xml
@@ -324,18 +324,5 @@
 			<Url>https://github.com/JoeRobich/fd-quicklaunch/releases/download/v0.10/QuickLaunch_0.10.fdz</Url>
 		</Urls>
 	</Entry>
-	<Entry>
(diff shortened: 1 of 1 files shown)


Gama11:
Interesting, I didn't know this was a thing - quite hidden in fact. Why not add a toggle button to the ProjectManager for this (@Meychi)? Both InteliJ and Eclipse have that, for example.

Meychi:
Feel free to do a PR :)

Gama11:
I'm wondering what the best option here is... There already is a "Locate Active Document" button, could we turn that one into a toggle button that switches this setting? Or is it important to have a dedicated button for Locating the active document *a single time*?

There are probably enough buttons as is there.

Meychi:
I think changing that would be best option.

PR #639, opened 2015-03-31
ProjectManager: turn "Locate Active Document" into "Track Active Document", see #638

Gama11:


Neverbirth:
I've got some complaints about this change. It caused FD to lose the ability to locate the current active document with a single click or by keyboard, and after giving it some thought I think they are right, one thing is the setting to track the active document, and another one is the behavior offered by the old button. You can get it by clicking twice, but that isn't very nice.

One of the initial comments for this change was that IDEs like IntelliJ work like this, but that's not entirely true:

intellij

The track documents toggle is under the gear button, together with other settings in a dropdown menu:

intellij2

That Autoscroll from source option, but the first button shown in the first picture is like our old Locate button.

I was thinking about making some tristate option like "Track: Never, Automatic, Manually", but in the end that would be stupid: in Automatic the button is still a toggle button or how we do switch between the other modes? in Never we disable the button? So I think the best option would be to revert this back.

Comments @Meychi @elsassph @Gama11 ?

PR #640, opened 2015-03-31
Fix a crash when drag- and dropping a project file

Gama11:


Meychi:
Thanks. Could you please describe this a little bit so i could check what causes this? Because i think the doc should not be null.

Gama11:
Like the title says, this happens when I drag- and drop a FD project file into the editor to open it. `OpenEditableDocument()` returns null in that case - which I suppose makes sense since project files are not opened in the editor? Haven't investigated further than that, as a simple null check seems to fix the issue.

Meychi:
Ah, FD project file, not project related file. Ok, thanks!

PR #644, opened 2015-04-01
Fixed DockPanel and Scintilla Marker images in high DPI.

JoeRobich:
Certain images scale poorly with the HighQuality setting. Added alternate
method to the ScaleHelper to stretch these images using NearestNeighbor.

Before:
image

After:
image


JoeRobich:
I dug a little deeper and figured out DockPanel images, like pin and close, where using color swapping to change from White to Black. I updated the InertButtonBase to use an inversion matrix so that the interpolated colors get swapped as well. I was then able to change those images back to using ScaleHelper.Scale instead of the Stretch method.

After:
image

Meychi:
Great stuff. I think maybe we should have new bigger markers instead of scaling them. What do you think? 

Also i need to think if i could custom draw some of the controls to improve the theming.

JoeRobich:
I think I can fix the issue with scaling the marker images. Basically I am going to reduce the number of colors to 50 using the quantizer from this blog (https://msdn.microsoft.com/en-us/library/aa479306.aspx) and make them pngs with transparency.

Meychi:
Sounds good to me. Go for it.

JoeRobich:
Replaced Marker bmps with pngs. Added a color quantizer so that XPM can reduce color depth before converting image to string. Changed Markers to be Scaled not Stretched. Removed ScaleHelper.Stretch.

After
image

Meychi:
These don't work correctly in default setup all that good. Try these with clean FD at 100% scaling to see what i mean.

JoeRobich:
Sorry about that. I missed the part where InertButtonBase recolored to the Foreground color. I thought it swapped between black and white based on Background darkness. I have updated the code to recolor while maintaining the alpha channel. But it seems this PR won't update anymore. I'll start a new one and reference this one.

I reverted to the default theme and the Visual Studio debugger seems to simulate 100% scaling. Other than the color being wrong for DockPanel images, did you see any other issues?

After:
image

Meychi:
Now it looks better. I reverted the incorrect merge so please make a new PR. I will test the new PR at 100% before merging it. Thanks!

PR #646, opened 2015-04-01
Fix DockPanel and Marker image scaling issues 2

JoeRobich:
This is a continuation of an earlier PR https://github.com/fdorg/flashdevelop/pull/644

@100%
image

@200
image


Meychi:
Hmm... i'm not satisfied how the icons end up looking. Is there anyway to improve the looks when they scale? Or should we create a bigger set of icons for the high dpi mode? Also in the latest commit the close icons (x) are scaled incorrectly at 100%.

This PR is ballooning atm so i might end up manually picking the changes.

JoeRobich:
I am guessing the Marker icons don't come out looking great because XPM doesn't support partial transparency. 

As for unintentional image scaling, I looked harder at the Quantizer code since that was something I was trusting to work properly. I found the following

```
// Draw the source image onto the copy bitmap,
// which will effect a widening as appropriate.
g.DrawImageUnscaled(source, bounds);
```

> The DrawImageUnscaled method draws an image using its physical size, so the image will have its correct size in inches regardless of the resolution (dots per inch) of the display device.

I squashed all the commits and changed that line to a `g.DrawImage`. I hope that resolves things. =)

PR #647, opened 2015-04-03
Move the caret / selection in SelectionDuplicate

Gama11:
Like FD, IntelliJ IDEA also has a Ctrl+D shortcut to duplicate the current selection (or line if the selection is empty), but it also
- moves the caret down one line if a line is duplicated
- moves the selection to the duplicated text if the selection is duplicated

I think this is convenient, so I implemented the same behavior here.

It does seem a little counter-intuitive when you undo the duplication, because the selection doesn't go back to what it was before. I tried `Begin` / `EndUndoAction()`, but Scintilla doesn't care about selection changes in the undo history.
The [docs](http://www.scintilla.org/ScintillaDoc.html) say:

>Scintilla saves actions that change the document. Scintilla does not save caret and selection movements, view scrolling and the like.

I'm not sure how this behavior could be emulated either, there are undo / redo event listeners, but I don't know how you would be able to tell that what's being undone / redone is a SelectionDuplicate action.


Meychi:
Instead of modifying the builtin method, please create a new one so that we don't break any existing code that use this method.

Gama11:
I don't think it's used anyhwere else (VS says 0 references, though it doesn't detect the one in mainmenu.xml of course).

Meychi:
3rd party plugins. :)

Gama11:
Done.

PR #649, opened 2015-04-03
Update HaxeProjects when the classpath is rebuilt

Gama11:
Previously, FD would not recognize any version change of haxelibs (e.g. when doing a haxelib set on the command line or installing a new version), even when manually refreshing the project - the project had to be reloaded entirely.
There might be a better solution that automatically detects changes, like a FileSystemWatcher for the .current files.

### DIFF
--- a/External/Plugins/HaXeContext/Context.cs
+++ b/External/Plugins/HaXeContext/Context.cs
@@ -180,7 +180,7 @@ namespace HaXeContext
                 string path = "";
                 do { 
                     path = p.StandardOutput.ReadLine();
-                    if (path != null && path.Length > 0 && !path.StartsWith("-") && Directory.Exists(path))
+                    if (!string.IsNullOrEmpty(path) && !path.StartsWith("-") && Directory.Exists(path))
                     {
                         path = NormalizePath(path).TrimEnd(Path.DirectorySeparatorChar);
                         paths.Add(path);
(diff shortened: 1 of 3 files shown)


External/Plugins/HaXeContext/Context.cs
```diff
@@ -388,6 +388,8 @@ public override void BuildClassPath()
             // add haxe libraries
             if (proj != null)
             {
+                proj.UpdateVars(false);
```
elsassph: This can have serious consequences, especially with OpenFL-like projects types.

We should just make sure that pressing the 'Refresh' button in the project panel does resolve the haxelibs properly - it normally should be the case.

elsassph:
We should not detect haxelib versions changes, but make sure that you can request FD to resolve the haxelibs on demand.

External/Plugins/HaXeContext/Context.cs
Gama11: So basically, this call should only happen when Refresh is pressed?

elsassph: This is what happens currently (unless there's a bug somewhere):
When you press the Refresh button in the project panel it sends a `ProjectManager.UserRefreshTree` event which hints HaxeContext to clear the haxelib version cache. Then `BuildClasspath` is called again and it should pick the correct haxelib versions.

Gama11: FD definitely doesn't pick up on haxelib version changes right now when pressing "Refresh"  - for OpenFL projects at least.

The haxelibCache is indeed cleared when "Refresh" is pressed, but it's empty to begin with for those projects. It seems to be crucial that display is run again and that hxml is parsed for the new versions, which happens in `ExternalToolchain::UpdateProject()` (which my `UpdateVars()`-call leads to).

elsassph: Well then maybe we need `UpdateVars` to be called after Refresh is pressed. But it just shouldn't happen in `BuildClasspath`.

PR #651, opened 2015-04-03
Use ToolStripEx for the ProjectManager's TreeBar to enable clickthrough, closes #648

### LINKED ISSUE #648: ProjectManager TreeBar buttons aren't clickable when FD doesn't have focus
If FD is in the background, then these buttons still appear to be clickable, however it doesn't work. This is confusing because visually, it still looks like the button was clicked.

Initially I thought that this could be a `ToolStripButton` thing, as the main menu buttons are `ToolStripMenuItem`s and replacing the Project Manager buttons with `ToolStripMenuItem`s fixes the issue (not really a great workaround though because the `Checked` property behaves differently) - but `ToolStripButton`s are used in other places like QuickFind and they don't have this issue.

Gama11:


Meychi:
I could not reply earlier to your issue but i knew this was the issue, thanks.

PR #667, opened 2015-04-14
Project GetRelativePath improvement

### LINKED ISSUE #655: Breakpoints can lead to path too long exception
As mentioned here:
http://www.flashdevelop.org/community/viewtopic.php?f=6&t=12191

Recently I started up FlashDevelop and got this bug:

```
The specified path, file name, or both are too long. The fully qualified file name must be less than 260 characters, and the directory name must be less than 248 characters.

   at System.IO.Path.SafeSetStackPointerValue(Char* buffer, Int32 index, Char value)
   at System.IO.Path.NormalizePathFast(String path, Boolean fullCheck)
   at System.IO.Path.NormalizePath(String path, Boolean fullCheck)
   at System.IO.Path.GetFullPathInternal(String path)
   at System.IO.Path.GetFullPath(String path)
   at ProjectManager.Projects.ProjectPaths.GetAbsolutePath(String baseDirectory, String path)
   at ProjectManager.Projects.Project.GetAbsolutePath(String path)
   at FlashDebugger.BreakPointManager.Load()
   at FlashDebugger.PluginMain.HandleEvent(Object sender, NotifyEvent e, HandlingPriority prority)
```

Which was weird because it used to be fine; I hadn't just installed a new version or done anything weird. At first I though my project might have recently added really long, deeply nested folders or something, so I even ran Path Length Checker to be sure. I was well within the safe margins for everything.

But I knew the bug was caused by a file path somewhere being too long or else malformed. 

Looking through the stack I noticed it was the breakpoint manager. So I dug up the location of the breakpoints data in the settings, which turned out to be:

```
C:\Users\YOUR_USER_NAME\AppData\Local\FlashDevelop\Data\FlashDebugger\Breakpoints\
```

And deleted all the xml files. Now when I start up FlashDevelop it works perfectly!

I wish I would have saved some of those xml files, but even so I can almost certainly guess the issue is that I put breakpoints in some weird obscure places that ultimately generated fully resolved file paths that were too long for FD to handle.

elsassph:
- Optimize GetRelativePath for trivial cases (very intensive usage in PM tree)
- Do not consider path sharing the same drive letter as relative (ie. on C:\ common) - should fix #655

### DIFF
--- a/External/Plugins/ProjectManager/Projects/ProjectPaths.cs
+++ b/External/Plugins/ProjectManager/Projects/ProjectPaths.cs
@@ -13,8 +13,20 @@ namespace ProjectManager.Projects
                 throw new ArgumentException("The path is already relative.");
 
             char slash = Path.DirectorySeparatorChar;
-            string[] a = baseDirectory.Trim(slash).Split(slash);
-            string[] b = path.Trim(slash).Split(slash);
+            path = path.TrimEnd(slash);
+            baseDirectory = baseDirectory.TrimEnd(slash);
+
+            // trivial cases
+            if (path == baseDirectory) 
+                return "";
+            if (path[1] == ':' && path[0] != baseDirectory[0]) // drive
+                return path; 
+            if (path.Length > baseDirectory.Length && path.StartsWith(baseDirectory + slash))
+                return path.Substring(baseDirectory.Length + 1);
+
+            // resolve relative path
+            string[] a = baseDirectory.Split(slash);
+            string[] b = path.Split(slash);
 
             ArrayList relPath = new ArrayList();
             int i = 0;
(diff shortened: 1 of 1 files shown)


Meychi:
I could test this if you could me few examples of a bad paths.

elsassph:
I was thinking this would be worth starting adding unit tests :)

The tests should be: 
- faster: paths are equals, 
- faster: path is contained in base directory (typically relative paths of files within the project folder),
- faster not relative: paths have different drive letter,
- consider not relative: paths only share the drive letter (typically classpath in very different location, like a haxelib VS your project's folder).

The last case is one reason we can have very long relative paths because the paths share nothing but the drive letter so I think it makes sense to not consider them as relative. For example: 
- relative path of `C:\g\h\i\j\k\l` from `C:\a\b\c\d\e\f` is `..\..\..\..\..\..\g\h\i\j\k\l`
- recombined as absolute path becomes `C:\a\b\c\d\e\f\..\..\..\..\..\..\g\h\i\j\k\l`
- this essentially doubles the path length can lead to the exception.

Neverbirth:
> I was thinking this would be worth starting adding unit tests :)

If so, please, consider XUnit or NUnit before MSTest.

elsassph:
@Neverbirth feel free to recommend your preference.

PR #668, opened 2015-04-15
Fix localization name of HighlightMatchingWordsDelay

Gama11:
The description wouldn't show up / produce an error in the output panel.


waneck:
Brb

Issue #444, opened 2014-10-16
Functions with default values string including comas


elsassph:
http://www.flashdevelop.org/community/viewtopic.php?f=6&t=11969

PR #679, opened 2015-05-02
Better string parsing. Fixes Issue #444

Neverbirth:
Do not skip \ and the next character when parsing strings. All IDEs I've tried leave them in hints, and when there are escaped Unicode or ASCII characters it results in strange output. Also, skipping it could result in other wrong results, for example some bad formed include statement.

ASComplete.FindNearSymbolInFunctDef has been rewritten to better detect parameters. The new function is larger, but is faster, more inline with file parsing, and I find it to be easier to get what the function actually does than the previous one, but the latter may be just something subjective. Are ({[ really possible? I left them, but sincerely, I don't know when they can appear (aside from trying to use objects and arrays in default values, which is wrong).

This adds multi line string support in Haxe in the native parser as well. Having multiline strings as default value to optional arguments could result in several issues, for example incomplete models and some completion issues. It also fixes some completion issues inside functions if there are multiline strings.

Last, this fixes pressing , inside a string as a parameter of a function when the call tip is visible.

BTW, some FB versions also hint parameters incorrectly if there are commas inside a string used as the default value.

EDIT: Forgot to comment that pressing , inside a string with the call tip visible does not work with Haxe if we are inside some secondary line in a multiline string. But this is because Scintilla styling fails as you know.

### DIFF
--- a/External/Plugins/ASCompletion/Completion/ASComplete.cs
+++ b/External/Plugins/ASCompletion/Completion/ASComplete.cs
@@ -1297,11 +1297,11 @@ namespace ASCompletion.Completion
             // measure highlighting
             int start = calltipDef.IndexOf('(');
             while ((start >= 0) && (paramIndex-- > 0))
-                start = FindNearSymbolInFunctDef(calltipDef, ",", start + 1);
+                start = FindNearSymbolInFunctDef(calltipDef, ',', start + 1);
 
-            int end = FindNearSymbolInFunctDef(calltipDef, ",", start + 1);
+            int end = FindNearSymbolInFunctDef(calltipDef, ',', start + 1);
             if (end < 0)
-                end = FindNearSymbolInFunctDef(calltipDef, ")", start + 1);
+                end = FindNearSymbolInFunctDef(calltipDef, ')', start + 1);
 
             // get parameter name
             string paramName = "";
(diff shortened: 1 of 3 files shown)


External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -1344,31 +1344,94 @@ static private void ShowCalltip(ScintillaControl Sci, int paramIndex, bool force
             else UITools.CallTip.CallTipSetHlt(start + 1, end, true);
         }
 
-        static string[] featStart = new string[] { "/*", "{", "<", "[", "(" };
-        static string[] featEnd = new string[] { "*/", "}", ">", "]", ")" };
-
         static private int FindNearSymbolInFunctDef(string defBody, string symbol, int startAt)
```
elsassph: It looks like this function is used only with single char symbols - would this be worth optimizing for that case?

Neverbirth: There are /* and */, so I thought that maybe in the future there may be needs for other strings and decided to make it more generic... but probably /* and */ could be treated separatedly and only care about chars.

elsassph: No I mean the only `symbol` that are looked up in the function definition are `,` and `)`.

Neverbirth: I see, yes, if wanted it could be changed, I don't know if there are chances of wanting to support something else.

Will try to commit the change later today.

Neverbirth: Done

PR #680, opened 2015-05-02
Improved regex handling.

Neverbirth:
- Parse Haxe regex literals correctly.
  - Take into account regex literals can follow an object literal property.
  - Take into account regexes in FRSearch.

It could cause some (minor) completion issues but more importantly, it's responsible for this:

https://twitter.com/_bigp/status/594314905878134786

EDIT: Surely regex parsing should take into account the possible flags as well.

### DIFF
--- a/PluginCore/PluginCore/FRService/FRSearch.cs
+++ b/PluginCore/PluginCore/FRService/FRSearch.cs
@@ -519,6 +519,7 @@ namespace PluginCore.FRService
                             if (c == '/' && pos < len - 1)
                                 if (src[pos + 1] == '*') commentMatch = 1;
                                 else if (src[pos + 1] == '/') commentMatch = 2;
+                                else LookupRegex(src, ref pos);
                         }
                         else if (commentMatch == 1)
                         {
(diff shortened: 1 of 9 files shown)


PluginCore/PluginCore/FRService/FRSearch.cs
```diff
@@ -608,6 +609,45 @@ private List<SearchMatch> SearchSource(string src, int startIndex, int startLine
             return results;
         }
 
+        private bool LookupRegex(string ba, ref int i)
+        {
+            if (PluginBase.CurrentProject == null ||
```
elsassph: I think we don't need to limit to AS3/Haxe - we don't have this condition when looking up /* ... */ comments as we don't even have the concept of code VS non-code files.

Neverbirth: But regex literals are a special case, supported only on a handful of languages, but that need to be handled like comments and string literals, otherwise when FRSearch finds a quote inside a regex literal it thinks it's a string literal, and cause bugs like the one above.

elsassph: We certainly need to lookup regexes, I just mean that we should not test the project language.

If you want to do Haxe-specific lookup it would make more sense to check the file extension (if that's an information we have).

Neverbirth: No, we don't have the file name, that was my first thought in fact, but since modifying FRSearch to pass that info meant modifying code in several places and adding more logic, I decided to take this approach.

Neverbirth: Of course I can make the change if you prefer, but it may be later today or already tomorrow, you caught me yesterday in the middle of another thing.

Neverbirth:
Modified FRSearch to get the source file name and use it to check for regex literals.

Neverbirth:
Forgot to add that the latest commit also checks for regex flags.

PluginCore/PluginCore/FRService/FRSearch.cs
```diff
@@ -608,6 +612,61 @@ private List<SearchMatch> SearchSource(string src, int startIndex, int startLine
...
 
+        private bool LookupRegex(string ba, ref int i)
+        {
+            if (SourceFile == null)
+                return false;
+
+            string ext = System.IO.Path.GetExtension(SourceFile).ToLowerInvariant();
+
+            // Haxe, ActionScript, and JavaScript support Regex literals
+            if (!Helpers.FileInspector.IsActionScript(SourceFile, ext) &&
```
elsassph: Worth moving this detection to the beginning of the parsing instead of resolving that every time the function is called.

elsassph: Actually, although passing `SourceFile` is one valid approach, I think it's not the right one as it doesn't support LoomScript files. Which means we have to hardcode more and more file extensions here.

Not sure what should be the right approach - we could check the Scintilla syntax configuration (does it have regexes?), but ultimately I think we should just look up any `/.../` as regexes no matter what the language and including Haxe where a valid regex has to start with `~`.

Neverbirth: But Scintilla syntax configuration also depends on file extensions, plus the Haxe configuration file doesn't have the entry set (I guess because since regexes start with ~ they aren't styled correctly).

I prefer how it is currently than assume all files equal. If PluginCore had access to the language syntax files maybe that could be used as you say, although I'd go for individual check for Haxe at any rate. IMHO, I'd leave it this way for now. If in the future FRSearch needs further language specific checks I'd revisit then how to handle it.

Neverbirth:
Fair enough

PR #683, opened 2015-05-13
Improved smart indenting in XML completion.

Neverbirth:
I've modified the behaviour when pressing enter on XML files. The current system has some problems and may result in files with some erratic indentation, especially with multi line tags.

The new indentation mostly mimics the style of some IDEs like Visual Studio, others like Eclipse-based ones are similar but have some slight differences (if there are multiple tags in the same line or if a tag spawning to multiple lines has different indentation on each one).

I'd say the only key difference with other IDEs is the last Enter shown in the gifs, while the new code always adds a new line, other IDEs may just reindent the following closing tag. I feel adding a new line is better.

xmlold

xmlnew

The GIF quality is not the best and there are some artifacts, but I think it shows the improvements well. Not all cases are shown.

One thing I don't like of the current indentation system is that it requires several undo steps for getting the original typed code, but that's not a problem of the new code (although maybe avoiding calls to SetLineIndentation when not needed may improve things just a bit), and decided to leave it as is right now.

### DIFF
--- a/External/Plugins/XMLCompletion/XMLComplete.cs
+++ b/External/Plugins/XMLCompletion/XMLComplete.cs
@@ -286,7 +286,8 @@ namespace XMLCompletion
                     }
                     if (PluginSettings.SmartIndenter)
                     {
-                        // Get last non-empty line
+                        // There is no standard for XML formatting, although most IDEs have similarities. We are mostly going with Visual Studio style with slight differences.
+                        // Get last non-empty line.
                         String text = "";
                         Int32 line2 = line - 1;
                         while (line2 >= 0 && text.Length == 0)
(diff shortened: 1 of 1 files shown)


elsassph:
@Neverbirth I think your formatting system would be a lot smarter if it would account to the fact that closing tags is optional in html5: ie `<meta ...>`, ``, etc. don't need the `/>`. This is an information that you should have access to as "leaf" tags are annotated in the HTML completion support.

Neverbirth:
I'll try to make some investigation and tests. Strictly speaking that shouldn't be here, as this is generic XML formatting and closing tags is mandatory, Scintilla has some different syntax for HTML , I don't know how FlashDevelop behaves in that way, is that one used instead of the XML syntax? if so it could be checked, but since FD uses an old version I wouldn't be surprised if there are some problems there. I won't be able to look into this today tho.

I've also noticed yesterday at night a slight problem in my code, nothing serious , in most cases it won't even appear, and thought about adding some extra comments for other people as well.

PR #685, opened 2015-05-15
Update appman.xml

mikicho:
Add "Find&Replace Advance" plugin for AppMan

### DIFF
--- a/appman.xml
+++ b/appman.xml
@@ -407,7 +407,7 @@
 		<Name>Find&ReplaceAdvance</Name>
 		<Version>1.1</Version>
 		<Build>1</Build>
-		<Checksum>a2f7af174c16655cb4b6fbcbf97d1611</Checksum>
+		<Checksum>52f824304662af369133fadd035eb8f1</Checksum>
 		<Desc>Find and Replace plugin Expanded.</Desc>
 		<Group>Plugins</Group>
 		<Info>http://www.flashdevelop.org/community/viewtopic.php?f=4&t=12200</Info>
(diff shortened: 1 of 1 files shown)


Meychi:
Great work! I finally had time to review this and here are my comments:

1. Do not include setting files with the FDZ installer
2. The controls should use consistent flat style and other properties
3. The controls should have adequate space around them and be aligned properly
4. Make sure the plugin works up to 200% text scaling properly

Please address these issues and we will add the plugin to AppMan.

mikicho:
Hi,
Thanks :)
Setting file included in FDZ because I added some settings so if someone already have the plugin he need to delete his setting file else it will crash.
I will fix the rest.

Meychi:
You need to make it so that it will not crash without settings file. :)

Also you will overwrite user settings if the user updates the plugin.

mikicho:
>
1.    Do not include setting files with the FDZ installer
2.    The controls should use consistent flat style and other properties
3.    The controls should have adequate space around them and be aligned properly
4.    Make sure the plugin works up to 200% text scaling properly

1. How I can do it in elegant way? I succeed only to check if one of my settings is 0 but it's don't work in any cases.
2. All controls should be flat style or only ComboBox?
3. Done
4. Done, except group header in results list. Do you have an idea how I can resize them too?

Thanks :)

Meychi:
Where is the updated FDZ? Also you need to update the MD5 to match file when ready.

mikicho:
I updated [v1.1 realease](https://github.com/mikicho/findandreplaceex/releases/tag/v1.1), I'll update the MD5 after you approve my plugin :)

Meychi:
Not a blocker but in the future please apply themes correctly if possible. You may need to use extended FD controls like ListViewEx and such to make it work correctly.

Im ok for merge, please update the MD5.

mikicho:
Ok, Done.
I'll apply those themes change in my nect release.

PR #691, opened 2015-05-21
Enhance ThemeManager to allow for Controls to inherit base styling

JoeRobich:
For instance if you currently wanted to set the ForeColor for all classes that inherit ToolStripItem you have to update ToolStripButton, ToolStripTextBox, TooStripComboBox, etc... with these changes you could simply set ToolStripItem.ForeColor.

To set default colors for all Controls you could now define Control.ForeColor and Control.BackColor.


Meychi:
Done.

Issue #694, opened 2015-05-23
Find And Replace In Files *.hxp extension filter includes .hxproj


Gama11:
Not sure why this is happening, maybe it's only checking the first three chars of the extension. Either way, *.hxp should not include FD's .hxproj files.

Removing the *.hxp als gets rid of the results in .hxproj files. 

![](http://i.imgur.com/JchWwGj.png)

PR #708, opened 2015-05-25
Fixes: Find And Replace In Files *.hxp extension filter includes .hproj #694

elsassph:
When extension mask is 3 chars (ie. not .as or .hx, but .hxp) it will match any extension of more characters (ie. .hxproj).

The fix adds a verification that the matched file extension really matches the mask.

### DIFF
--- a/PluginCore/PluginCore/Utilities/PathWalker.cs
+++ b/PluginCore/PluginCore/Utilities/PathWalker.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using PluginCore;
 using System.ComponentModel;
 using System.Windows.Forms;
+using System.Text.RegularExpressions;
 
 namespace PluginCore.Utilities
 {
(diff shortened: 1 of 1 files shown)


PluginCore/PluginCore/Utilities/PathWalker.cs
```diff
@@ -98,8 +98,13 @@ private void ExploreFolderWithMasks(String path, String[] masks)
             foreach (String mask in masks)
             {
                 String[] files = Directory.GetFiles(path, mask);
+                String control = mask.Length == 5 && mask[0] == '*' && mask[1] == '.' ? mask.Substring(1) : null;
```
Neverbirth: I'd say this won't work if mask is something like *.hx*, *.hx?, *f*.hx, etc. PluginCore has Helpers.FileHelper.FileMatchesSearchFilter, although I'd say it would be more costly for something like this where you only want to check the extension (the extension part could be extracted tho).

elsassph: Yeah it would fail with fancy matching after the initial `*.`. Not sure how to efficiently sove that, and if that's a use case we want to support at all.

elsassph: Well I guess I'll use a regex instead.

PR #711, opened 2015-05-26
appman.xml: update to Haxe 3.2.0

Gama11:


elsassph:
I think we should offer both 3.1.3 and 3.2 downloads for now.

Gama11:
Ok. They need a different `Id` I guess? Does it matter what those are? `haxe3setup1` and `haxe3setup2` would work?

Meychi:
Nope, but haxe3setup2 is better.

Gama11:
Updated.

PR #723, opened 2015-05-27
Adjustments in scrollbar calculations

elsassph:
Still slightly off but it looks decent.


Meychi:
Done is separate commit with extras. :)

PR #725, opened 2015-05-28
Apply Theme to DataGridView cells

JoeRobich:
BreakPointUI grid was still rendering cells with the default background color. These changes update the DefaultCellStyle when themes are applied.


Meychi:
Thanks!

PR #729, opened 2015-05-28
Implement Haxe position completion, closes #702

### LINKED ISSUE #702: [haxe] implement compiler position completion
http://haxe.org/manual/cr-completion-position.html

This can be used for more reliable "Goto Declaration" results and in cases where FD is unable to find a declaration at all.

Gama11:


External/Plugins/HaXeContext/Context.cs
```diff
@@ -1381,10 +1384,66 @@ internal void OnFunctionCompletionResult(HaxeComplete hc, HaxeCompleteStatus sta
...
+                        {
+                            HighlightDeclaration(sci, result);
+                        });
+                    }
+                    else HighlightDeclaration(sci, result);
+                    break;
+            }
+        }
+
+        internal void HighlightDeclaration(ScintillaControl sci, HaxePositionCompleteResult result)
```
elsassph: Not sure this should be part of HaxeContext - I think there should be 1 point where we open and highlight the document to show a declaration.

elsassph:
Background execution can end outside the UI thread so it's important to check for `InvokeRequired`.

About `REGULAR` I think it could be something like `HaxeCompilerService.COMPLETION`, `HaxeCompilerService.POSITION`.

elsassph:
Aside from the comments I think you did a fine job - although I have yet to try :)

Gama11:
Any idea why it's not always working correctly even with checking `InvokeRequired` then?

The Haxe Manual considers all of these services as "completion". I think it makes sense to stick to the "official" wording, even if position / usage completion isn't "completion" in the traditional sense.

External/Plugins/HaXeContext/Context.cs
Gama11: Yeah, this would be fixed if this was made consistent with FD's current goto declaration highlighting behavior like I mentioned, using `ASComplete.LocateMember()`. The field name is necessary for that.

elsassph:
I'm not sure why `sci.EnsureVisible` doesn't always works - maybe @Meychi will have an idea, otherwise I'll give it a try soon.

Even if the manual call all the services "completion", the "services" name is what stick when I discuss with TiVo engineers working on IntelliJ Haxe plugin.

Gama11:
Hm.. yeah, it does make more sense. Maybe we should change the Haxe Manual instead. :) (cc @Simn)

Simn:
I'm fine with changing it. It was the original term and I never bothered to think about it, but I can see why it might be confusing.

Gama11:
@elsassph I renamed the enum to `HaxeCompilerService`. Also, `ASComplete.LocateMember()` is now used for highlighting making it consistent with the current goto declaration highlighting.

I also haven't noticed the `EnsureVisible` issue anymore, not sure if it's gone.

Gama11:
@elsassph Any update on this?

elsassph:
I'll look into that now - I was wondering what happens when you have Haxe < 3.2.

Gama11:
It fails and outputs "No completion point found" to the output panel. Should we check that the Haxe version is at least 3.2 before attempting to use position completion? I don't think there's any mechanism to easily compare haxe versions right now..

elsassph:
That's the problem: we don't want to enable that for Haxe < 3.2 and there isn't an easy way to know it for now. Maybe the SDK version can be passed in the contexts `BuildClassPath` in the `ContextSetupInfos` object provided by Project Manager.

Gama11:
Updated. I'm retrieving the current version via the `hxSettings` object and the `currentSDK` path in `Context`.

Unfortunately, it seems like the versions in the `InstalledSDKs` array aren't too reliable. If I install version 3.2.0 over 3.1.3 or vice versa, FD doesn't detect any version changes even after restarting it until I manually delete the SDK.

External/Plugins/HaXeContext/Context.cs
```diff
@@ -1615,4 +1673,48 @@ public HaxeCompletionCache(ASContext context, MemberList elements, MemberList ot
             OtherElements = otherElements;
         }
     }
+
+    /// <summary>
+    /// Represents a semantic version, see http://semver.org/
+    /// </summary>
+    class SemVer
```
Gama11: This class is potentially useful outside of `Context.cs` or even the `HaXeContext` plugin, but I'm not sure where else to put it. `FlashDevelop.Utilities` maybe?

Gama11:
FD was actually trying to validate all installed Haxe SDKs on startup, but it wasn't working due to a bug. This should now be fixed.

External/Plugins/HaXeContext/Context.cs
elsassph: @Meychi maybe?

Gama11: Done.

External/Plugins/HaXeContext/Context.cs
```diff
@@ -1381,10 +1399,48 @@ internal void OnFunctionCompletionResult(HaxeComplete hc, HaxeCompleteStatus sta
...
+        internal void OnPositionCompletionResult(HaxeComplete hc, HaxePositionCompleteResult result, HaxeCompleteStatus status)
+        {
+            switch (status)
+            {
+                case HaxeCompleteStatus.ERROR:
+                    TraceManager.AddAsync(hc.Errors, -3);
+                    break;
+
+                case HaxeCompleteStatus.POSITION:
+                    ASComplete.SaveLastLookupPosition(hc.Sci);
```
elsassph: Is this isn't thread safe it's not good - the entire case POSITION should be execute after texting for InvokeRequired.

External/Plugins/HaXeContext/Context.cs
```diff
@@ -577,6 +580,21 @@ public override void RemoveClassCompilerCache()
         }
         #endregion
 
+        #region SDK
+        private InstalledSDK GetCurrentSDK()
+        {
+            return hxsettings.InstalledSDKs.FirstOrDefault(sdk => sdk.Path == currentSDK);
```
elsassph: `FirstOrDefault` sounds super useful but it sounds like it will never return null - or can it?

Gama11: It can return null - for example when the list of installed SDKs is empty. The "Default" in the name refers to the fact that it returns `default(TSource)` (see [here](https://msdn.microsoft.com/en-us/library/vstudio/bb340482(v=vs.100).aspx)), which is `null` in this case.

Fun fact, this used to be a 5-line `foreach`-loop before Resharper suggested to turn it into a LINQ-expression. ;)

External/Plugins/HaXeContext/Context.cs
Gama11: Updated. Guess two invokes are needed here for the two different `ScintillaControl`s.

External/Plugins/HaXeContext/Context.cs
elsassph: And I thought you were some C# god! ;)
Maybe I'll give another look at Resharper but I really didn't like it last time I tried.

External/Plugins/HaXeContext/Context.cs
```diff
@@ -1381,10 +1399,60 @@ internal void OnFunctionCompletionResult(HaxeComplete hc, HaxeCompleteStatus sta
...
+
+        private void HandlePositionCompletionResult(HaxeComplete hc, HaxePositionCompleteResult result)
+        {
+            ASComplete.SaveLastLookupPosition(hc.Sci);
+
+            PluginBase.MainForm.OpenEditableDocument(result.Path, false);
+            ScintillaControl sci = PluginBase.MainForm.CurrentDocument.SciControl;
+            const string keywords = "(function|var|[,(])";
+
+            if (sci.InvokeRequired)
```
elsassph: This `InvokeRequired` test isn't needed anymore.

Gama11: Even though it's a different ScintillaControl?

External/Plugins/HaXeContext/Context.cs
```diff
@@ -1381,10 +1399,60 @@ internal void OnFunctionCompletionResult(HaxeComplete hc, HaxeCompleteStatus sta
...
+                return false;
+
+            var hc = new HaxeComplete(sci, expression, false, completionModeHandler, HaxeCompilerService.POSITION);
+            hc.GetPosition(OnPositionCompletionResult);
+            return true;
+        }
+
+        internal void OnPositionCompletionResult(HaxeComplete hc, HaxePositionCompleteResult result, HaxeCompleteStatus status)
+        {
+            switch (status)
```
elsassph: It's ok, but you could have wrapped it all with `InvokeRequired` test and use `TraceManager.Add()` instead of `AddAsync` (that does the same thing).

FlashDevelop/Utilities/SemVer.cs
```diff
@@ -0,0 +1,51 @@
+﻿namespace FlashDevelop.Utilities
```
elsassph: I believe it should be added to PluginCore and not into FlashDevelop. Plugins aren't supposed to reference the FlashDevelop project itself.

External/Plugins/HaXeContext/Context.cs
Gama11: I quite like it. I'm a big fan of IntelliJ and it allows you to easily configure a lot of shortcuts the same way. The code analysis is really helpful too (convert to LINQ expressions like here, warns about potential null errors, unused usings / variables / functions etc).

Some inspections can be a bit annoying (the naming convention rules especially, since FD doesn't match most of those, but they can be turned off or adjusted).

External/Plugins/HaXeContext/Context.cs
elsassph: It doesn't matter: we use `InvokeRequired` to join the UI thread, it can be done with any UI element you have a reference to.

Gama11: I see, makes sense. :)

External/Plugins/HaXeContext/Context.cs
elsassph: I think that's the problem - I don't like IntelliJ's "Christmas tree" style where you have to spend your time toying with the warnings settings. Also it felt an absolute CPU hog.

External/Plugins/HaXeContext/Context.cs
elsassph: That's called "thread affinity" :)

Gama11: Good to know!

Gama11:
Updated with all the mentioned improvements.

External/Plugins/HaXeContext/HaXeContext.csproj
```diff
@@ -85,6 +85,10 @@
     <Reference Include="System.XML" />
   </ItemGroup>
   <ItemGroup>
+    <ProjectReference Include="..\..\..\FlashDevelop\FlashDevelop.csproj">
```
elsassph: Remove reference.

Gama11: Fixed!

PR #732, opened 2015-05-28
Native 64bits support

Neverbirth:
Useful for people that needs more memory, surely rare for FD itself (maybe with some plugins), but may be useful for Java. Also some people on locked environment has only Java 64bits, or prefer to have it and not install more than one version, etc.

The 64 bits version is actually AnyCPU, on x86 it would run as the current FD. Even if it's AnyCPU, when running in a 64bits PC, you cannot use a 32bits Java, this limitation comes from JNI4NET itself.

Compile FD in x86 and AnyCPU, the order doesn't matter, at the end the output folder will have both a FlashDevelop.exe and FlashDevelopx64.exe binaries. All the built-in plugins and external libraries are compatible in both modes, but there may be third party plugins compiled only for x86, I've found only one, but there are surely more out there, I haven't tested many of them, anyway, if someone needs some x86-only plugin and the autor doesn't update it, the user could run FlashDevelop.exe as always (there are tools to change the CPU type target, but if the plugin uses native API there are always chances that the author didn't correctly made it CPU agnostic).

I extensively tested it, but since I often swap my FD versions to work on other features there may be some problem I didn't notice, I'd vote for releasing test versions together with FD5.


elsassph:
Probably useful for Flash/AIR and with Java defaulting to 64bits.
What about 32bits system support? Wine/CrossOver?

Neverbirth:
FlashDevelop.exe is x86 as always, so there's no compatibility loss in that regard. Just in case I tested this in 32 bits systems back in the day with no problems as expected.

I don't know how the AnyCPU version would work on Wine/Crossover. I should look into installing that some day.

elsassph:
Oh sorry I didn't notice that there would be both 32 and 64bits FD.exe. What about Scintilla? Does it use a 64bits build of Scintilla?

Neverbirth:
Yes, on x86 it uses the same current version and on x64 uses a 64 bits build, based on the same version. I thought of using newer scintilla but didn't want to break anything or convert this on more than 64 bits support.

PR #755, opened 2015-06-04
Sci.IsVScrollBar and IsHScrollBar defer to active scrollbar

JoeRobich:
When custom scrollbars are in use then IsVScrollBar and IsHScrollBar should get/set their visibility instead of the native scrollbar visibility. 

- Separate properties for custom/native scrollbar visibility are added.
- The IsVScrollBar/IsHScrollBar defer to the active scrollbar's visibility property.
- Scrollbar visibility is synced when custom scrollbars are added or removed.
- OnResize doesn't include custom scrollbar size when their visibility is set to false.

### DIFF
--- a/PluginCore/ScintillaNet/ScintillaControl.cs
+++ b/PluginCore/ScintillaNet/ScintillaControl.cs
@@ -44,6 +44,8 @@ namespace ScintillaNet
 
         private ScrollBarEx vScrollBar;
         private ScrollBarEx hScrollBar;
+        private bool useCustomScrollBars = false;
+
 
         /// <summary>
         /// Gets the reference to the custom vertical scrollbar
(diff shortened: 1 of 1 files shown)


Meychi:
I just started thinking this yesterday. :) With this implementation we would need to expose and sync multiple properties between split sci controls. I'll try to come up with a simpler approach.

Meychi:
Ok, i did my implementation and based on my tests it works correctly, comments?

https://github.com/fdorg/flashdevelop/commit/e5bce38e42fa686727b80a7c4423f58e92bdc0c9

JoeRobich:
My only thought is that Resize needs to be called anytime the custom scrollbar's visibility changes.

Meychi:
Ok, i will address and test that.

Meychi:
Should be good now. Other comments or ideas?

JoeRobich:
No. I think that will work great. Thanks!

PR #757, opened 2015-06-05
Changed scintilla marker images from using XPM to RGBA.

JoeRobich:
Since RGBA supports an alpha channel marker images look better when scaled for high dpi.

- Added MarkerDefineRGBAImage method.
- Added RGBA static class for converting a Bitmap into a byte[].
- Updated the breakpoint, current line, and bookmark markers to use
  DefineRGBA method.


Meychi:
Great!

PR #758, opened 2015-06-05
ProjectTreeView refreshes colors on ApplyTheme

### LINKED ISSUE #738: Project Manager: node background colors aren't reset on theme switch
- Switch from Default to Thyleus:

- Switching back:

The background colors from the old theme stick around. They reset after selecting and deselecting each node.

JoeRobich:
Resolves #738.

- ProjectTreeView handles ApplyTheme event.
- On ApplyTheme all nodes are updated with the new ForeColor and BackColor.

### DIFF
--- a/External/Plugins/ProjectManager/Controls/TreeView/ProjectTreeView.cs
+++ b/External/Plugins/ProjectManager/Controls/TreeView/ProjectTreeView.cs
@@ -9,6 +9,8 @@ using System.Collections.Generic;
 using System.Collections.Specialized;
 using ProjectManager.Projects.AS3;
 using PluginCore;
+using PluginCore.Managers;
+using System.Linq;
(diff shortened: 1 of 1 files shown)


Meychi:
This code has compile errors. And is the ThemeControls call needed as this is done from MainForm before the event comes?

JoeRobich:
Sorry about the build errors. I should have know better than copying my changes into the github editor.

I call ThemeControls because the ApplyTheme event fires before the ProjectTreeView is themed and I didn't want to look up the foreground/background colors myself. =)

JoeRobich:
Of course I would be happy to update the code to ask for the ThemeColors.

Meychi:
The event in my opinion should come after. I think your code worked without that too.

PR #761, opened 2015-06-10
Update Settings for FP 18

SlavaRa:


elsassph:
Looks like we stopped including the corresponding `playerglobal.swc`s in `Tools/flexlibs` since FP14. Should we fix that?

FlashDevelop/Bin/Debug/Projects/310 Haxe - Flash Project/Project.hxproj
```diff
@@ -8,7 +8,7 @@
     <movie fps="30" />
     <movie width="800" />
     <movie height="600" />
-    <movie version="17" />
+    <movie version="18" />
```
elsassph: We need to check what FP version is supported by the Haxe compiler - Haxe 3.1.3 didn't even accept FP17.

Meychi:
We do patch the appman installed SDKs. The FD bundled SWCs are if we patch any used SDK. They add the size of FD. What should we do?

FlashDevelop/Bin/Debug/Projects/310 Haxe - Flash Project/Project.hxproj
Meychi: Last supported version for haxe 3.2 is 17 = 28. 3.1.3 supports also up to 12.5 -> 28 but swf-version is labeled incorrectly. Last correct version is 12.0 -> 23. What should we do?

Meychi:
I will merge this and lets fix the Haxe versions before release.

elsassph:
My opinion is that we should use the most commonly supported Flash version in the Haxe (and possibly Flash) templates. Someone who really needs a specific more recent Flash version will (probably) know what he is doing.

Meychi:
Then that would be 12.

PR #769, opened 2015-06-14
FlashDevelop startup optimization

elsassph:
- Delayed FlashConnect connection
- Delayed DockContent creation for rarely used UIs - applied to Profiler, FlashDebugger, FlashLogs Viewer.
- Dock panels now default to hidden.
- Show Project panel on first run (better first run experience)

### DIFF
--- a/FlashDevelop/Docking/DockablePanel.cs
+++ b/FlashDevelop/Docking/DockablePanel.cs
@@ -22,7 +22,6 @@ namespace FlashDevelop.Docking
             this.HideOnClose = true;
             this.Controls.Add(ctrl);
             Globals.MainForm.ThemeControls(this);
-            //this.Show();
         }
 
         /// <summary>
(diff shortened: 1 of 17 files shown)


FlashDevelop/Docking/DockablePanel.cs
```diff
@@ -22,7 +22,7 @@ public DockablePanel(Control ctrl, String pluginGuid)
             this.HideOnClose = true;
             this.Controls.Add(ctrl);
             Globals.MainForm.ThemeControls(this);
-            this.Show();
+            //this.Show();
```
elsassph: Seem to work fine with NOT showing panels by default and letting the layout handle it.

Meychi:
Should this be still worked on?

elsassph:
I proposed to do this optimization to try reducing startup time. I also wanted to hide rarely needed panels so that the first run experience looks cleaner.

At the very least, FlashConnect startup should be delayed significantly.

Meychi:
I think also "Show ProjectPanel on firstRun" is important. I think the debugger panels should be docked together. Otherwise i think the initial ui is ok. I suggest we do:

"Show ProjectPanel on firstRun"  & FlashConnect startup delayed & debugger panels docked together.

elsassph:
Ok, are you up to extracting it or should I do it?

Meychi:
I can do this.

Meychi:
Done.

PR #780, opened 2015-06-23
Automatically add the semicolon after imports

Gama11:
semicolon

Having to type the `;` manually every time when you already selected the item from the completion list is somewhat tedious.

### DIFF
--- a/External/Plugins/ASCompletion/Completion/ASComplete.cs
+++ b/External/Plugins/ASCompletion/Completion/ASComplete.cs
@@ -3958,7 +3958,8 @@ namespace ASCompletion.Completion
                 ContextFeatures features = ASContext.Context.Features;
 
                 // add ; for imports
-                if (expr.WordBefore == features.importKey || expr.WordBefore == features.importKeyAlt)
+                if (trigger != ';' && (expr.WordBefore == features.importKey ||
+                    expr.WordBefore == features.importKeyAlt))
                 {
                     sci.InsertText(sci.CurrentPos, ";");
                     sci.SetSel(sci.CurrentPos + 1, sci.CurrentPos + 1);
--- a/External/Plugins/ASCompletion/Completion/ASComplete.cs
+++ b/External/Plugins/ASCompletion/Completion/ASComplete.cs
@@ -184,10 +184,9 @@ namespace ASCompletion.Completion
                         if (features.hasGenerics && position > 2)
                         {
                             char c0 = (char)Sci.CharAt(position - 2);
-                            bool result = false;
                             if (c0 == '.' /*|| Char.IsLetterOrDigit(c0)*/)
                                 return HandleColonCompletion(Sci, "", autoHide);
-                            return result;
+                            return false;
                         }
                         else break;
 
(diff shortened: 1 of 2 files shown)


External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -184,10 +184,9 @@ static public bool OnChar(ScintillaControl Sci, int Value, bool autoHide)
                         if (features.hasGenerics && position > 2)
                         {
                             char c0 = (char)Sci.CharAt(position - 2);
-                            bool result = false;
                             if (c0 == '.' /*|| Char.IsLetterOrDigit(c0)*/)
                                 return HandleColonCompletion(Sci, "", autoHide);
-                            return result;
+                            return false;
```
Gama11: @elsassph Resharper noticed that `result` is always `false`. ;)

External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -4052,7 +4060,7 @@ private static bool SmartInsertion(ScintillaControl sci, int position, ASExpr ex
             if (inFile == null || import == null)
                 return false;
 
-            if (expr.Separator == ' ' && expr.WordBefore != null && expr.WordBefore != "")
+            if (expr.Separator == ' ' && !string.IsNullOrEmpty(expr.WordBefore))
```
Gama11: Another nice Resharper inspection here.

elsassph:
That's an interesting idea but I'd prefer that we avoid adding too many little things like that. Adding imports manually is such a rare operation that we're adding needless complexity and "code noise" to ASComplete.

Also what if you press semicolon to insert the import?

Gama11:
Adding **6 lines** of code in a **4500-line file** adds code noise? Really?

If `ASComplete` isn't readable enough, it's not because of small things like this, it's because of poor organization / a refactor being overdue. Some functions like `HandleDotCompletion()` are so big you can barely tell what's going on anymore.

How rare manually inserting imports is depends on your workflow. I do it quite often actually. Mabye because I didn't know that `Ctrl+Shift+1` could resolve imports. I'd imagine that's the case for a lot of users, as I've been using FD for years now.

I didn't realize pressing semicolon could trigger completion insertion. I guess that would result in two semicolons, but that should be a simple fix.

Gama11:
Fixed the double-semicolon bug.

elsassph:
Well this class certainly grew out of propotions... it's begging for a big refactoring, like many other giant classes of FD. 
One day...

PluginCore/PluginCore/Controls/CompletionList.cs
```diff
@@ -840,17 +839,11 @@ static public bool ReplaceText(ScintillaControl sci, String tail, char trigger)
         
         #region Event Handling
         
-        /// <summary>
```
elsassph: Haha, that's @Meychi 's doc placeholders, begging you to comment your code :D

elsassph:
Awesome this was inserting a ; everywhere in Flash, because ImportKeyAlt is null here.

PR #785, opened 2015-06-27
Project: add LanguageDisplayName

Gama11:
Language.ToUpper works for as2 and as3, but looks strange for haxe (HAXE) and loom (LOOM).

### DIFF
--- a/External/Plugins/LoomContext/Projects/LoomProject.cs
+++ b/External/Plugins/LoomContext/Projects/LoomProject.cs
@@ -23,7 +23,7 @@ namespace LoomContext.Projects
         }
 
         public override string Language { get { return "loom"; } }
-        public override string LanguageDisplayString { get { return "Loom"; } }
+        public override string LanguageDisplayName { get { return "Loom"; } }
         public override bool IsCompilable { get { return true; } }
         public override bool ReadOnly { get { return false; } }
         public override bool HasLibraries { get { return OutputType == OutputType.Application; } }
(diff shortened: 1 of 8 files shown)


External/Plugins/ProjectManager/Projects/Project.cs
```diff
@@ -76,6 +75,7 @@ public Project(string path, CompilerOptions compilerOptions)
         }
 
         public abstract string Language { get; }
+        public abstract string LanguageDisplayString { get; }
```
Gama11: I guess adding an `abstract` property here is a breaking API change.. Are there any third-party plugins that depend on this API?

elsassph: No, it's an addition so I believe it's compatible.

External/Plugins/LoomContext/Projects/LoomProject.cs
```diff
@@ -27,6 +23,7 @@ public override string Name
         }
 
         public override string Language { get { return "loom"; } }
+        public override string LanguageDisplayString { get { return "Loom"; } }
```
elsassph: What about `LanguageDisplayName`?

External/Plugins/ProjectManager/Projects/Project.cs
Gama11: But classes that inherit from `Project` need to implement it, right?

External/Plugins/LoomContext/Projects/LoomProject.cs
Gama11: Renamed. That's actually a much better name! :+1:

External/Plugins/ProjectManager/Projects/Project.cs
elsassph: I don't think anybody to have created a plugin adding a new project type, so that should be ok.

PR #786, opened 2015-06-27
Update config

SlavaRa:


elsassph:
Yeah I'm not sure about this list of tools, arguably it should include SourceTree and msysgit as well.

Gama11:
+1 for SourceTree. :)

Issue #799, opened 2015-07-12
After renaming, FlashDevelop keeps old path in memory


elsassph:
Example: 

```
/src/character/Example.as
```

Rename folder as:

```
/src/Character/Example.as
```

When opening `Example.as` (during the same session), FlashDevelop reuses the `FileModel` or the original path with lowercase `character` because models are matched in non-case sensitive way.

PR #800, opened 2015-07-12
Fixes issue #799 with FileModels path casing

elsassph:
the model's path is "refreshed" every time it is resolved to be used for the current document.


Meychi:
Seems all good.

PR #802, opened 2015-07-17
Do not generate super calls with Haxe's '?' marker (optional parameter).

### LINKED ISSUE #801: Bug: [Haxe] - Overriding method with optional parameters generates incorrect super() call
If you override a haxe method that has optional parameters, the '?' character gets added to the super() method call.

So:
```
function foo( ?a:Bool)
{
}

The generated override will become

override function foo(?a:Bool)
{
    super.foo(?a); //This will generate a compile time error.  
}
```

elsassph:
Fixes #801: Bug: [Haxe] - Overriding method with optional parameters generates incorrect super() call


Meychi:
Looks good to my vacation eyes :)

elsassph:
Well merge if you're happy :)

PR #813, opened 2015-08-01
fixes: organize imports from one or more

SlavaRa:
In release(5.0.1):

After fix:


elsassph:
It seems that Organize Imports was disabled on purpose when you only have one import. Why re-enabling it?

SlavaRa:
You can see from the screenshots that flash.display.Sprite is not used in the code, but to remove the import does not work through refactoring.

PR #840, opened 2015-08-28
Add "In Classpaths only" checkbox on the "Open Resource" dialog

yar3333:


Gama11:
If you look at the "Files changed" tab, you'll notice you're using a different indent in some places (tabs instead of spaces).

yar3333:
OK, fixed.

PR #859, opened 2015-09-06
quickbuild bug fix

S-Aoi:
Now `@mxmlc` can receive `+configname=air` argument. And paths can be quoted with `"`.

### DIFF
--- a/External/Plugins/AS3Context/Compiler/FlexShells.cs
+++ b/External/Plugins/AS3Context/Compiler/FlexShells.cs
@@ -29,7 +29,7 @@ namespace AS3Context.Compiler
         static public event SyntaxErrorHandler SyntaxError;
 
         static readonly public Regex re_SplitParams = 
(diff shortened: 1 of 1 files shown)


External/Plugins/AS3Context/Compiler/FlexShells.cs
```diff
@@ -321,6 +321,10 @@ public void QuickBuild(FileModel theFile, string flex2Path, bool requireTag, boo
                                             }
                                         }
                                         isPath = true;
+                                        // remove quotes
+                                        if (arg.StartsWith("\"") && arg.StartsWith("\""))
```
Gama11: Isn't the second condition here supposed to use `EndsWith()`?

elsassph:
Why removing quotes exactly?

S-Aoi:
Some paths may contain spaces. When I quote the path with `"`, FD show me an error saying `This file is not a valid AS3 class or does not include a valid @mxmlc build command.`.

elsassph:
Oh sorry, you're right there was a problem when your mxmlc command has quotes already.

PR #876, opened 2015-09-23
Fixed behavior after click of on the target build selector

SlavaRa:
Click of on "target build selector" before fix:
bug_click_on_targetbuildselector

Click after fix:
fixes_click_ob_targetbuildselector


Gama11:
I think that behavior is intentional so you can use unlisted targets.

SlavaRa:
For this is the configuration file

elsassph:
This is intentional.

SlavaRa:
ok.

PR #898, opened 2015-10-06
Show more path if several same-named references exists

yar3333:
project


SlavaRa:
nice )

elsassph:
Nice, but could we just add 'library' to the tokens ignored (like 'src')? That is using the mechanism which already exists.

yar3333:
@elsassph: May be yes, how to do that ("library" is duplicated only in my projects - it is not a common practice; other developers can have other duplicated names)?

elsassph:
Well 'library' seen a fairly generic term, like 'src' so it would be also be worth not counting as a useful folder name.
Your solution is somehow smarter but it works only if you have two projects using 'library' as the last folder name.

yar3333:
Many "library" - it is a problem for me, one "library" - is not! :)))

Meychi:
This list can be managed from ProjectManager settings -> Filtered Directory Names. If you want to improve this, please modify the current logic in ClasspathNode.

Issue #851, opened 2015-09-02
Haxe completion error


kevinresol:
```
--macro:1: character 0 : Could not parse `snow.system.module.Module.set(\"Audio\",\"snow.core.web.audio.Audio\")`
--macro:1: character 0 : Invalid character '\'
```

Was using Luxe and got this error during completion. Probably related to https://github.com/fdorg/flashdevelop/commit/a739113e801536528515abdd395c4183715f84b9

ref: https://github.com/underscorediscovery/snow/blob/master/framework.flow#L185

elsassph:
Hmm, how is configured your project? Is it FD flow integration  or something manual?

kevinresol:
using this template https://github.com/Chman/Snowkit-FD

kevinresol:
it is strange though: if I clone the repo and build it, it works. but it fails if I download the build from build server, both at c9e1bcc

kevinresol:
This error still exist at latest build 6e32087b2d91
Are there anything I can do to provide more info?

kevinresol:
OK, i double checked, building FD from source does not solve the problem. What solve the problem is not to use completion server

PR #900, opened 2015-10-10
Should not escape quotes for --macro statements

kevinresol:
#851


Meychi:
@elsassph: Is this correct?

elsassph:
Seems correct but didn't have a chance to try. Might well merge.

Meychi:
For next minor or major release?

PR #907, opened 2015-10-12
Native 64bits support 

Neverbirth:
Updated the branch to latest changes, and targets development branch instead of the, back then required, fd5 one.

See PR #732 for further details.


elsassph:
@Meychi it's good for me, but it may have an impact in terms of setup. Both versions need to be built by the CI.
On my computer, `x86` target builds the 32bits version and `Any CPU` target builds the 64bits version.

Issue #894, opened 2015-10-05
[AS2Context][Context][CheckModel] Package 'xxx' did not match file path doesn't work for root (empty) package


gene-pavlovsky:
I really like the automatic package name fixing feature, but it doesn't work for "package {" classes.
If I have a class in the root of my source path (e.g. I made something quick-and-dirty, then decided it's good enough to actually reuse), so I decide to move it to package "com.blah.blah.blah", when I open the file, the automatic package name fixer doesn't work. If I change the first line from "package {" to "package whatever {", and save, then the name gets fixed to the correct one.

PR #927, opened 2015-10-18
Better package and class name check

Neverbirth:
Modified the package and class name validation to be more aggressive. Fixes issue #894 and some others I've found while testing.

I've also seen that the code tries to display the messagebar with two lines of text, but only one is displayed, but haven't looked into this.


Neverbirth:
I'm sure AS2 package replacement is broken, I'll check it today.

elsassph:
What's the state of this PR?

Neverbirth:
After working on this I think I have discovered a bug in another part of FD, but still didn't check it nor the AS2 case.

Meychi:
?

Neverbirth:
I haven't looked into this yet, I've been busy with more important tasks, then I plan to come back to this, as I think it's also important.

PR #942, opened 2015-10-23
OrganizeImports, ASGenerator: sort imports by package then by name

### LINKED ISSUE #882: Organize imports - improve sort order
Right now Code refactor "Organize imports" feature has a weird sort order.
```
import com.razorscript.math.ArithmeticError;
import com.razorscript.math.expression.supportClasses.MathFunBase;
import com.razorscript.math.MathContext;
import com.razorscript.math.RoundingMode;
```

I think it would make more sense for this to be
```
import com.razorscript.math.ArithmeticError;
import com.razorscript.math.MathContext;
import com.razorscript.math.RoundingMode;
import com.razorscript.math.expression.supportClasses.MathFunBase;
```

The current sort order especially doesn't work well with Code refactor "Separate packages" option.

gene-pavlovsky:
Closes #882

### DIFF
--- a/External/Plugins/ASCompletion/Model/MemberModel.cs
+++ b/External/Plugins/ASCompletion/Model/MemberModel.cs
@@ -538,4 +538,40 @@ namespace ASCompletion.Model
         }
 
     }
+
+    /// <summary>
+    /// Compare members based on import name
+    /// </summary>
+    public class ByImportTypeMemberComparer : IComparer<MemberModel>
+    {
+        public static Int32 CompareImports(String import1, String import2)
+        {
+            // Use case-insensitive comparison.
+            //IComparer cmp = StringComparer.OrdinalIgnoreCase;
+            // Use case-sensitive comparison.
+            IComparer cmp = StringComparer.Ordinal;
+            String[] parts1 = import1.Split('.');
+            String[] parts2 = import2.Split('.');
+            int len1 = parts1.Length;
+            int len2 = parts2.Length;
+            // If the imports are at the same depth, compare them alphabetically.
+            if (len1 == len2)
+                return cmp.Compare(import1, import2);
+            int minPackageLen = ((len1 <= len2) ? len1 : len2) - 1;
+            // Alphabetically compare import packages part by part.
+            for (int i = 0; i < minPackageLen; ++i)
+            {
+                int cmpResult = cmp.Compare(parts1[i], parts2[i]);
+                if (cmpResult != 0)
+                    return cmpResult;
+            }
+            // One of the packages is a sub-package of the other one. Consider the parent package to compare as less than the sub-package.
+            return len1 - len2;
+        }
+
+        public Int32 Compare(MemberModel item1, MemberModel item2)
+        {
+            return CompareImports(item1.Type, item2.Type);
+        }
+    }
 }
--- a/External/Plugins/CodeRefactor/Commands/OrganizeImports.cs
+++ b/External/Plugins/CodeRefactor/Commands/OrganizeImports.cs
@@ -149,7 +149,7 @@ namespace CodeRefactor.Commands
         {
             String eol = LineEndDetector.GetNewLineMarker(sci.EOLMode);
             Int32 line = imports[0].LineFrom - DeletedImportsCompensation;
-            imports.Sort(new ImportsComparerType());
+            imports.Sort(new ByImportTypeMemberComparer());
             sci.GotoLine(line);
             Int32 curLine = 0;
             List<String> uniques = this.GetUniqueImports(imports, searchInText, sci.FileName);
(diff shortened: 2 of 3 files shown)


gene-pavlovsky:
**Should we use Comparer instead of CaseInsensitiveComparer to make string comparisons case-sensitive?**

External/Plugins/CodeRefactor/Commands/OrganizeImports.cs
```diff
@@ -226,7 +225,24 @@ class ImportsComparerType : IComparer<MemberModel>
     {
         public Int32 Compare(MemberModel item1, MemberModel item2)
```
SlavaRa: It may be worth it to make the code in one place?

elsassph:
The logic should be the same for auto-imports.
Non-case sensitive sorting it's an interesting idea, but I'd rather follow a rule enforced in other IDE so we should look what IntelliJ does.

External/Plugins/CodeRefactor/Commands/OrganizeImports.cs
gene-pavlovsky: I'd like that, unfortunately I have very little idea about C# in general and structure of FD in particular. (This being my first use of C#, actually). When I asked Mika how can I use the compare function from one plug-in in the other plug-in, he advised me to copy&paste it to avoid creating references between plug-ins...

Neverbirth: The comparer could be in ASCompletion and be reused by CodeRefactor without any problem.

gene-pavlovsky: Having never used C# before, for some reason I couldn't make the class from one plug-in visible in the other. But on second try it worked, maybe I've made a typo or something. I've moved the comparison code to a new ByImportTypeMemberComparer class in MemberModel.cs. OrganizeImports uses this member comparer for sorting, however ASGenerator.cs does regex-based matching so it needs to compare strings, not members, so I exposed the string comparison function as a static function ByImportTypeMemberComparer.CompareImports
I'm not sure it's the best place for this class, feel free to move it somewhere else.

gene-pavlovsky:
@elsassph Flash Builder uses case-sensitive comparison (e.g. in my example it sorts MathComplex MathReal Mathematics). Don't have another IDE to test.

Neverbirth:
Just checked on IntelliJ, it's also case sensitive in there.

elsassph:
Does the sorting require this complex logic? Doesn't a direct 'case sensitive sort' of the imports provide identical results?

gene-pavlovsky:
Not that complex :)
I think there's only one difference. With current logic, a package-level function will be sorted before any deeper level stuff, direct sort might not.
Example.
`
import aaa.function1;
import aaa.bbb.Class1;
import aaa.zzz.Class2;
`

Direct sort would produce:
`
import aaa.bbb.Class1;
import aaa.function1;
import aaa.zzz.Class2;
`

It makes a different for identifiers starting with a lowercase letter, such as package-level functions or namespaces (I think that's all), or if some developer uses an uncommon-for-AS naming convention where classes can start with lowercase or packages can start with uppercase.

elsassph:
I still wonder if there could be an implementation not doing the string.split...

gene-pavlovsky:
Well, split itself is not required, the depth of the package path could be determined by going through all the strings' chars and counting the number of '.'. I used split since I don't program in C# and it was an quick way to implement it. I wonder if you have a way to profile how long would sorting a thousand package names would take? If it's still fast enough, then maybe there's no need for rewrite?
I see, I didn't consider the completion lists...

elsassph:
Let's think about it, I'm interested in making FD more consistent.
It will be another good candidate to unit tests coverage which we are trying to put in place with @neverbirth.

Neverbirth:
IMHO that this and completion list use a different sorting method shouldn't be a blocker of any type, as they are different use cases.

wise0704:
image
Just a note that VS doesn't follow this rule.

External/Plugins/ASCompletion/Model/MemberModel.cs
```diff
@@ -538,4 +538,40 @@ public int Compare(MemberModel a, MemberModel b)
...
+    /// Compare members based on import name
+    /// </summary>
+    public class ByImportTypeMemberComparer : IComparer<MemberModel>
+    {
+        public static Int32 CompareImports(String import1, String import2)
+        {
+            // Use case-insensitive comparison.
+            //IComparer cmp = StringComparer.OrdinalIgnoreCase;
+            // Use case-sensitive comparison.
+            IComparer cmp = StringComparer.Ordinal;
```
elsassph: You should make the code more efficient by finding index of `.` separating package and type and doing (up to) 2 comparisons for package and type.

There are a few edge cases to consider for Haxe, here are valid imports:

    import foo.bar.ModuleName;
    import foo.bar.ModuleName.TypeName;

As algorithm I'd do: 

- starting from 0, get next index of `.`
- if index < 0 previous index is separation
- if char after `.` is uppercase (I think there's `Char.IsUpperCase(c)`) then stop (this takes care of Haxe modules)
- else try getting next `.` and repeat

Gama11: In Haxe, there's also `import ... in` / `import ... as`, is that being taken into account?

```haxe
import foo.bar.ModuleName as Name;
import foo.bar.ModuleName in Name;
```

What about `using`?

elsassph: Good reminder: we generally don't properly support those, just kind of ignore this part of the import declaration. Normal insertions will work fine, but operations like Organize Imports will drop them.

Issue #943, opened 2015-10-24
Debugger shows 1D Vector as 2D Vector


hgupta9:
The debugger value tree shows a simple `Vector.<Object>` as a `Vector.<Vector.<Object>`, yes, with the closing `>` missing. Here is a screenshot:

![fd5_vector_bug](https://cloud.githubusercontent.com/assets/6797866/10710966/7dbc8dec-7a8a-11e5-846e-48ae9fd72208.png)

I may be responsible for this behavior since I updated the debugger rendering recently. I just don't have time right now to fix it, but if no one does soon, I will. I think the code responsible for this is [ValueNode.cs#L65][1] and/or.[ValueNode.cs#L42][2].
  [1]:https://github.com/fdorg/flashdevelop/blob/development/External/Plugins/FlashDebugger/Controls/DataTree/ValueNode.cs#L65 
[2]:https://github.com/fdorg/flashdevelop/blob/development/External/Plugins/FlashDebugger/Controls/DataTree/ValueNode.cs#L42

Neverbirth:
The fix for this was merged in the official branch yesterday.

hgupta9:
Thank you! That's cool.

Issue #806, opened 2015-07-29
Expressions in Watch Panel disappear after stopping debugging


dleanjeans:
It's really annoying when you add a lot of expressions into the Watch Panel and they disappear after stopping debugging. If I recall correctly, Visual Studio keeps them after that.

ortizgames:
I second this; this feature would be really useful for the heavy development I've been doing lately.

Neverbirth:
I´ll look into this today.

Neverbirth:
Just FYI, this was done yesterday, but before committing I want to also persist the expressions between application sessions like in other IDEs.

ortizgames:
Neverbirth, you are awesome! :)

Andres Ortiz
Game Designer & Programmer
http://www.ortizgames.com

Neverbirth:
In case you are not aware, this was merged yesterday, so you can download a development build for the feature.

ortizgames:
Neverbirth,




I had been watching your pull request, but I hadn't checked in a couple of days - thank you!! I'll be back at my desk on Sunday and trying this out immediately.



—
Andres

ortizgames:
Watch entries persist and it's a miracle!! :)

Issue is finally resolved; thank you, Neverbirth! The day when this hits the release channel can't come soon enough!



—
Andres

PR #944, opened 2015-10-25
Store watch expressions between sessions

Neverbirth:
FD now stores the watch expressions for each Project like other IDEs. This change also means that the watch panel behaviour is now a bit different, previously, the watch panel was disabled unless inside a debugging sesión, now it's all the time enabled in order to allow the user to manage the expressions.

This fixes issue #806 and also fixes some problems with the watch panel while a debugging session was live.

Also, used this PR to fix issue #943.


Neverbirth:
Changed the behaviour while running: the watch panel isn't disabled to avoid theming issues, but we do not try to resolve any data in order to avoid problems.

Some IDEs like IntelliJ are like this.

External/Plugins/FlashDebugger/Controls/DataTree/ValueNode.cs
```diff
@@ -274,6 +249,51 @@ public ValueNode(string text, Value value)
...
+            text = text.Replace("\n", "\\n");
+            text = text.Replace("\r", "\\r");
+            text = text.Replace("\t", "\\t");
+            text = text.Replace("\v", "\\v");
+            if (text.Length > 65533)
+                text = text.Substring(0, 65533 - 5) + "[...]";
+            return text;
+        }
+
+        private string CleanTypeClassPaths(string qualifiedName)
```
elsassph: This is such a good candidate for unit tests - we really have to add something.

External/Plugins/FlashDebugger/Controls/DataTree/ValueNode.cs
```diff
@@ -34,20 +35,18 @@ public override string Value
                 string temp = null;
                 if (type == VariableType_.MOVIECLIP || type == VariableType_.OBJECT)
                 {
-                    string typeStr = "";
+                    string typeStr = m_Value.getTypeName().ToString();
+                    // rename vector
+                    typeStr = typeStr.Replace("__AS3__.vec.Vector.<", "Vector.<");
```
elsassph: The type doesn't include `::` anymore for AS3 Vector?

External/Plugins/FlashDebugger/Controls/DataTree/ValueNode.cs
Neverbirth: Well, that code has been there "forever". I only moved it to another place since I didn't like it to be in the middle of properties.

External/Plugins/FlashDebugger/Controls/DataTree/ValueNode.cs
Neverbirth: You are completely right, copy & paste error. Well, if you decide everything else is right maybe you can merge it and then change the value to "__AS3__.vec::Vector.<". I cannot atm.

Neverbirth:
OK, fixed the Vector replacing. It should be ready.

External/Plugins/FlashDebugger/Controls/WatchUI.cs
```diff
@@ -35,43 +41,55 @@ private void TreeControlResize(Object sender, EventArgs e)
...
+        }
+
+        private void WatchManager_ExpressionRemoved(Object sender, WatchExpressionArgs e)
+        {
+            UpdateElements();
+        }
+
+        private void WatchManager_ExpressionReplaced(Object sender, WatchExpressionReplaceArgs e)
+        {
+            treeControl.Nodes[e.Position] = GetExpressionNode(e.NewExpression);
```
elsassph: Verify that position is valid.

External/Plugins/FlashDebugger/Debugger/WatchManager.cs
```diff
@@ -0,0 +1,206 @@
...
+using PluginCore;
+using PluginCore.Helpers;
+using PluginCore.Utilities;
+
+namespace FlashDebugger.Debugger
+{
+    public class WatchManager
+    {
+        private IProject m_Project;
+        private string m_SaveFileFullPath;
```
elsassph: Is `m_` the coding convention for this plugin? That's certainly not what we usually follow.

Neverbirth: Yes, most of the plugin is using it.

External/Plugins/FlashDebugger/Controls/DataTree/ValueNode.cs
elsassph: It seems that this function `CleanTypeClassPaths` is completely new.

After looking hard at the code I'm starting to understand what it's doing (am-I correct that it supports Haxe's fancy generics and function types?) but it really should have some comments (if not unit tests) showing the kind of cleanup it does because it's far from obvious.

External/Plugins/FlashDebugger/Controls/WatchUI.cs
Neverbirth: Well, the logic is centralized in the WatchManager instance, and I'd say it's not possible to break this part of the code unless some plugin developer makes some nasty stuff like accessing the private tree control through reflection, which AFAIK would not make sense having both the WatchUI and WatchManager instances accessible.

But I'd add some check.

External/Plugins/FlashDebugger/Controls/DataTree/ValueNode.cs
Neverbirth: You are correct in your assumption. I'll add some comments.

And a lot of us will wait for the fully documented ASFileParser, ASComplete, etc ;P.

External/Plugins/FlashDebugger/Controls/WatchUI.cs
elsassph: Fair enough.

External/Plugins/FlashDebugger/Controls/DataTree/ValueNode.cs
elsassph: Ahem :) Next time I rewrite the parser or completion I'll add comments :P

Neverbirth: Nice, will it be ready today? :P

External/Plugins/FlashDebugger/Controls/WatchUI.cs
Neverbirth: Added some comments.

External/Plugins/FlashDebugger/Debugger/WatchManager.cs
Neverbirth: I'll try not to forget for the new debugger version

Neverbirth:
Thank you!

PR #946, opened 2015-10-28
Add a conditional output path argument: "{Build}"

wise0704:
The argument "{Build}" should resolve to the current build configuration to enable better support for conditional output path.

### DIFF
--- a/External/Plugins/ProjectManager/Projects/Project.cs
+++ b/External/Plugins/ProjectManager/Projects/Project.cs
@@ -348,9 +348,9 @@ namespace ProjectManager.Projects
         public string FixDebugReleasePath(string path)
         {
             if (!TraceEnabled)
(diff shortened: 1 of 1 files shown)


elsassph:
This doesn't follow the regular arguments processing convention: at the very least it should be replacing `$(BuildConfig)` and not `{Build}`.

wise0704:
I thought of using $(BuildConfig), but I came across two problems:
First, sometimes having a "$" in the path screw things up. I thought of instead using (BuildConfig) but thought that people might actually use that in the path name. Braces seemed like a better option.
Also, the current compiler argument $(BuildConfig) resolves to all lowercase "debug" and "release". I decided that to distinguish the difference, I should suggest a different name.

However, I don't really mind much :) as long as there is a way to support the conditional arguments in the output path.

elsassph:
Using Pascal case `Debug`/`Release` is rather specific to your use case. While I applaud your effort in figuring what to change in FD build system, this is a situation which should be solved by a post-build script or a C# macro.

PR #947, opened 2015-10-28
Improved AddClosingBraces feature

### LINKED ISSUE #370: Support generics / <> for add closing braces
This one might be a bit tricky, as it requires some special checks (is the word to the left a type with generics, or are we just trying to use the `<` operator?).

wise0704:
fix for nested braces not adding extra closing braces
fix for closing single/double quote not overwriting existing closing quote
plus a small typo fixed

Edit: closes #370.


elsassph:
Can you should example codes demonstrating the change?

wise0704:
you mean can I show*?? :)

>fix for nested braces not adding extra closing braces

inserting a left parenthesis after another: 
Before: capture 1 After: capture 2
Closing of nesting braces also work perfectly.
capture 5

>fix for closing single/double quote not overwriting existing closing quote

Before: capture 3 After capture 4

External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -414,27 +416,42 @@ public static void HandleAddClosingBraces(ScintillaControl sci, char c, bool add
...
+                if (c == braces.closing && sci.CurrentChar == braces.closing)
                 {
-                    // already an opening brace?
-                    if ((char)sci.CharAt(sci.CurrentPos - 2) != braces.opening)
-                    {
-                        sci.InsertText(sci.CurrentPos, braces.closing.ToString());
-                    }
+                    // move the condition to above so that it handles opening properly
+                    //// already a closing brace?
+                    //if (sci.CurrentChar == braces.closing)
```
elsassph: Can you remove commented code here and there?

External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -414,27 +416,42 @@ public static void HandleAddClosingBraces(ScintillaControl sci, char c, bool add
                         HandleRemoveBrace(sci, c, braces);
                 }
             }
+            // If the current style is a string literal in double/single quotes and the matching char is entered
+            else if (c == '"' && IsDoubleQuoteStyle(style) || c == '\'' && IsSingleQuoteStyle(style))
+            {
+                // If the next char is the corresponding closing char
+                if (addedChar && c == sci.CurrentChar)
```
elsassph: What happens if you want to type an escaped quote (eg. `\"`)?

wise0704: Just like Visual Studio, the closing quote still is overridden. I would want it to not override for escaped quotes, but I figured that the process can get complicated. One case I could think of was when the leading backslash is also an escaped backslash, and in the process of checking if there is a backslash before the backslash, gotta check if it is escaped...

wise0704: Or actually, would chaning the line 420 (haha 420 sorry) to
```
else
{
    style = sci.BaseStyleAt(sci.CurrentPos - 1);
    if (c == '"' && IsDoubleQuoteStyle(style) || c == '\'' && IsSingleQuoteStyle(style))
    ...
```
work? Since if the entered char is still the double/single quote style, it means the string is not terminated, right?

wise0704:
I found a bug while trying to add support for escaped quotes not overriding.

`ASComplete.OnShortcut()` is sometimes not called when `key == Keys.Back`.
It seems completely random, and it is quite frustrating because sometimes I can't tell whether it's the bug or error in my code that is not deleting what should be deleted. Moving the caret up and down several times before pressing backspace seems to be the only solution.

External/Plugins/ASCompletion/Completion/ASComplete.cs
elsassph: The styles are in fact defined in `PluginCore.ScintillaNet.Lexers`. AS/Haxe typically use the Cpp lexer.

elsassph:
Maybe something (main app or a plugin or Scintilla itself) captures the key in some conditions.

Neverbirth:
Yep, I know about this, the problem is how FD is currently handling the key strokes for the completion list and the call tip. Currently the call tip is consuming the key. I have it fixed in a branch... and it would allow for more new features, so if you could go to [this board topic](http://flashdevelop.org/community/viewtopic.php?f=9&t=12472) and help with the 64 bits testing so it gets merged soon, I'll commit the work for the refactored completion list, that fixes this issue along with some others and have some usability improvements among other things.

elsassph:
Another improvement I'd suggest for auto-closing braces: do not insert a closing brace when it's likely to be undesirable. Like: `var foo = |bar;` I'd like to wrap `bar` with parenthesis but I have to delete the closing brace inserted.

wise0704:
@Neverbirth That's a good news! However, I'm sorry I can't help you with the 64 bit testing; I'm one of those who still has 32 bit computers. 

@elsassph I'm already working on that :)
Right now, I'm trying to sort out the style values that would most likely to be the ones that should be wrapped around. However the trouble is that some braces have different meanings and uses than others - `()` and `[]`  are very similar, but they are quite different from `{}` and `""` in many cases. 
I feel like I'll end up modifying the `Braces` class or other things, but shouldn't be too hard.

Neverbirth:
Even if you don't have 64bits you can help by reporting that the normal version is working as always without any problem.

I'll try to update another of the branches to the latest development commit and upload more builds soon.

wise0704:
About 45c9cee: please ignore this. I'm very new to GitHub (this is the first time ever using it) and I just downloaded the GitHub Desktop whatever and was playing around with it when this merge thing happened.

I added an option for the user to choose when to automatically add closing braces for each set of braces.
This turned out to be quite cool, because it allowed me to support the generic type braces `<>` as well as enable users to add custom 'braces', such as the the tertiary operator `?:`. To try this out just adding a new brace with `Opening = '?' Closing = ':' AfterStyle = DEFAULT` will do. 
Users can also manually add&remove default braces or their automatic closing options. 

I ended up adding a new resource string to PluginCore, and if it will be used they would need to be translated for languages other than en_US.

Neverbirth:
Not everyone has Visual Studio 2015 ;).

Neverbirth:
Also, I'd flag your fix for the backspace as not 100% valid, because there are cases where the backspace would stop working, for example if the calltip is available, but the underlying scintilla control isn't using ASComplete or if the setting ASContext.CommonSettings.AddClosingBraces is false.

wise0704:
@Neverbirth All I actually did in the fix is to make sure that `CompletionList` does not handle backspace itself (not calling `sci.DeleteBack()` and not setting `e.Handled = true`) - just as what would happen when the completion list is not available. 
I tested and it does seem to still work when AddClosingBraces is false or ASComplete is disabled.

wise0704:
Can anyone test this and give me some feedback on how this works?

SlavaRa:
I can test the night

Neverbirth:
I'll also test it asap.

wise0704:
Thanks :)

wise0704:
Is there anything I have to do about the "conflicts"?
image

SlavaRa:
I think that you need to pull fdorg/development into your branch, and resolve conflict after.

ps
Please write me an email ratdg@ya.ru.
I can not find any other way to contact you

elsassph:
So did anyone give it a try?

wise0704:
@elsassph doesn't seem like it... lol

wise0704:
Possibly closes #370.

elsassph:
It looks like the feature is specialized for the cpp-lexer. What happens when you open an XML/HTML/python/etc. file?

wise0704:
It seems like ASCompletion doesn't work for XML/HTML/Python/JavaScript.
This feature doesn't work for those and fortunately doesn't screw them up :)

elsassph:
Hey I'm giving a try and it's pretty good.
One rather annoying thing is the insertion of closing quote when it's really not desired:

    var foo = |bar  -> var  foo = "|"bar
    a = { |foo:"value" } -> a = { "|"foo:"value" }

So generally quotes shouldn't auto-close if there is a non-whitespace character at the cursor position.

wise0704:
Yea, string were quite difficult to manage, because the style value after an opening quote is always `StringEOL`, so with the default setting of `BeforeStyles` set to `{ StringEOL }`, string always auto-closes, except at the end of file. 
If there is a way to bypass this, quotes will be as configurable as the other braces. Got any method you know?

For now, try setting the `"` and `'` to have `BeforeStyles` empty, and `BeforeChars` to `ABCDEF...wxyz([{<...` and `BeforeStyleMode` to `Exclusive`. A cheap way but still would work...
Speaking of whitespaces, I should probably enable users to enter escaped characters.

elsassph:
The configuration suggestion didn't really work.

Meychi:
Progress on this?

Neverbirth:
From my side I'd like to wait a bit to see it merged (even if the conflicts are resolved I mean).

wise0704:
Ah yes, I'm still working on this.
Strings like below: (`|` is a caret)
```ActionScript
var a:String = "| + "b";
```
screws everything up by Scintilla reporting `" + "` is `STRING` and `b";` is `STRINGEOL`.

Simple solution I thought was to delete the inserted char and get the base style value afterwards, but then it seems like the style values don't get updated immediately T^T.

Neverbirth:
Yes, style values are not updated immediately depending of when you check them. You'd have to use sci.Colourise method.

wise0704:
Oh, so that method is like an update for the styles? Will try.

Meychi:
@wise0704: Let us know when you can complete this.

wise0704:
> Hey I'm giving a try and it's pretty good.
> One rather annoying thing is the insertion of closing quote when it's really not desired:
> 
> ```
> var foo = |bar  -> var  foo = "|"bar
> a = { |foo:"value" } -> a = { "|"foo:"value" }
> ```
> So generally quotes shouldn't auto-close if there is a non-whitespace character at the cursor position.

I am currently working on this, but the examples you provided aren't really likely to occur in real practice. Like, if they wrote `... = bar` already, chances are that `bar` is already an identifier, and they might want to do something like `"value = " + bar`.
This will be customizable by the user using AddClosingBracesData, but we would need to decide what to use as the default setting though.

External/Plugins/AS3Context/AS3Context.csproj
```diff
@@ -79,23 +79,17 @@
...
     <Compile Include="Controls\ProfilerUI.Designer.cs">
       <DependentUpon>ProfilerUI.cs</DependentUpon>
     </Compile>
     <Compile Include="Controls\ObjectRefsGrid.cs">
       <SubType>Component</SubType>
     </Compile>
-    <Compile Include="Controls\Win32.cs">
-      <SubType>Component</SubType>
-    </Compile>
+    <Compile Include="Controls\Win32.cs" />
```
wise0704: what's happening here...

External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -396,60 +394,101 @@ private static void AutoStartCompletion(ScintillaControl Sci, int position)
...
+            if (!ASContext.CommonSettings.AddClosingBraces) return;
+            
+            if (addedChar)
+            {
+                // Get the before & after style values unaffected by the entered char
+                sci.DeleteBack();
+                sci.Colourise(0, -1);
+                byte styleAfter = (byte) sci.BaseStyleAt(sci.CurrentPos);
+                byte styleBefore = (byte) sci.BaseStyleAt(sci.CurrentPos - 1);
+                sci.AddText(1, c.ToString());
```
wise0704: What's the cost of `sci.Colourise()`?
Previously (not the commit) I had the `Colourise` method to only refresh the current line, but all other usages of the method colours the whole document.
image
I haven't tested if there is any problem occurring when styling a line only, but if it doesn't cost too much to just colourise the whole document (seeing how `ASComplete.OnChar()` calls it every time), I'd just leave it as is.

elsassph:
I think it's ok to colourise with `(0, -1)`.

I really think the behaviour should be to not insert the closing brace if you type a brace at the beginning of a word: it's better to have to add the closing brace if it wasn't inserted automatically than to remove a useless one.

wise0704:
By "word" I assume you mean `Style.Identifier` and `Style.Keyword`?

wise0704:
I see those two are already excluded from the "" and '' braces. Can you try them out?
Also, are the other braces closures working as you would expect them to?

wise0704:
The configuration can be removed from setting by just removing it from the settings object, but that doesn't do anything performance-wise or code structure-wise. Also this configuration is really just a suggestion and depending on the programming language each user uses and their taste, many users would want to modify a few options. E.g. In AS3, generics ~~not even~~ are written as `Vector.<T>`, thus after `'.'`. In Haxe, they are written as `Array<T>`, thus after `Type`. (I did try to compromise by setting it to after `Type` or `Operator`.)

elsassph:
Alright, I like the defaults now, but the rules would require an entire WIKI to make sense :D
The setting may be called "Add Closing Brace Options" though.
Good to merge.

elsassph:
Oh BTW I had to add a reference to `System.Design` in ASCompletion.

wise0704:
In ASCompletion?
I already moved the `DescriptiveCollectionEditor<T>` class to PluginCore, so ASCompletion shouldn't be using `System.Design`. I can compile without problems?

Meychi:
No compile issues here. This looks good to me for merge.

Meychi:
This has now conflicts.

wise0704:
Seems like git is not familiar with .resx files.

Meychi:
Conflicts again.

wise0704:
Git obviously doesn't like .resx files...

wise0704:
Fixed conflicts.

Neverbirth:
This was marked good to merge, so if nobody complains I'll merge it before the weekend.

Meychi:
This has conflicts again. Soon as fixed, i will merge this.

wise0704:
Something is wrong with the resource files and causes +/-6k lines from each. Probably line endings.
Sec.

wise0704:
Fixed conflicts.

PR #950, opened 2015-11-09
Include available Refactor menus in Contextual Code Generation list

wise0704:
I had a feeling that the Refactor menus are very useful, but they are not easily accessible due to the fact that every option needs a separate shortcut assigned in order to access using keyboard only (unless you press the right-click button on your keyboard which is cheating). So, inspired by Visual Studio, I thought it would be nice if the available options were to show up in the existing code generator menu (assign the shortcut to `Ctrl+OemPeriod` and boom). 
capturecapture2capture3
Currently
- Rename
- Extract Method
- Extract Local Variable
- Generate Delegate Methods
- Organise Imports
- Truncate Imports

are available in the completion list, but more can be added with just 2 lines of code for each. (Edit: 1 line of code after 54bdb08)

Plus, extended the support of `Convert to Constant` to also work for strings.
capture4

### DIFF
--- a/External/Plugins/ASCompletion/Completion/ASGenerator.cs
+++ b/External/Plugins/ASCompletion/Completion/ASGenerator.cs
@@ -6,6 +6,7 @@ using System.Globalization;
 using System.IO;
 using System.Text;
 using System.Text.RegularExpressions;
+using System.Windows.Forms;
 using ASCompletion.Context;
 using ASCompletion.Model;
 using ASCompletion.Settings;
@@ -41,6 +42,10 @@ namespace ASCompletion.Completion
         static private MemberModel contextMember;
         static private bool firstVar;
 
+        static List<ICompletionListItem> known;
+
+        public static List<ICompletionListItem> KnownList => known;
+
         static private bool isHaxe
         {
             get { return ASContext.Context.CurrentModel.haXe; }
@@ -56,13 +61,17 @@ namespace ASCompletion.Completion
 
         static public void ContextualGenerator(ScintillaControl Sci)
         {
+            known = new List<ICompletionListItem>();
+
             if (ASContext.Context is ASContext) (ASContext.Context as ASContext).UpdateCurrentFile(false); // update model
             if ((ASContext.Context.CurrentClass.Flags & (FlagType.Enum | FlagType.TypeDef)) > 0) return;
 
             lookupPosition = -1;
             int position = Sci.CurrentPos;
-            if (Sci.BaseStyleAt(position) == 19) // on keyword
+            int style = Sci.BaseStyleAt(position);
+            if (style == 19) // on keyword
                 return;
+
             bool isNotInterface = (ASContext.Context.CurrentClass.Flags & FlagType.Interface) == 0;
             int line = Sci.LineFromPosition(position);
             contextToken = Sci.GetWordFromPosition(position);
@@ -72,10 +81,13 @@ namespace ASCompletion.Completion
             string text = Sci.GetLine(line);
             bool suggestItemDeclaration = false;
 
-            if (isNotInterface && !String.IsNullOrEmpty(contextToken) && Char.IsDigit(contextToken[0]))
+            if (isNotInterface)
             {
-                ShowConvertToConst(found);
-                return;
+                if (style == 4 || style == 6 || style == 7)
+                {
+                    ShowConvertToConst(found, style);
+                    return;
+                }
             }
 
             ASResult resolve = ASComplete.GetExpressionType(Sci, Sci.WordEndPosition(position, true));
@@ -567,8 +579,7 @@ namespace ASCompletion.Completion
                 GenerateJob(GeneratorJobType.AddImport, matches[0], null, null, null);
                 return;
             }
-
-            List<ICompletionListItem> known = new List<ICompletionListItem>();
+            
             foreach (MemberModel member in matches)
             {
                 if ((member.Flags & FlagType.Class) > 0)
@@ -576,35 +587,28 @@ namespace ASCompletion.Completion
                 else if (member.IsPackageLevel)
                     known.Add(new GeneratorItem("import " + member.Name, GeneratorJobType.AddImport, member, null));
             }
-            CompletionList.Show(known, false);
         }
 
         private static void ShowPromoteLocalAndAddParameter(FoundDeclaration found)
         {
-            List<ICompletionListItem> known = new List<ICompletionListItem>();
             string label = TextHelper.GetString("ASCompletion.Label.PromoteLocal");
             string labelMove = TextHelper.GetString("ASCompletion.Label.MoveDeclarationOnTop");
             string labelParam = TextHelper.GetString("ASCompletion.Label.AddAsParameter");
             known.Add(new GeneratorItem(label, GeneratorJobType.PromoteLocal, found.member, found.inClass));
             known.Add(new GeneratorItem(labelMove, GeneratorJobType.MoveLocalUp, found.member, found.inClass));
             known.Add(new GeneratorItem(labelParam, GeneratorJobType.AddAsParameter, found.member, found.inClass));
-            CompletionList.Show(known, false);
         }
 
-        private static void ShowConvertToConst(FoundDeclaration found)
+        private static void ShowConvertToConst(FoundDeclaration found, int style)
         {
-            List<ICompletionListItem> known = new List<ICompletionListItem>();
             string label = TextHelper.GetString("ASCompletion.Label.ConvertToConst");
-            known.Add(new GeneratorItem(label, GeneratorJobType.ConvertToConst, found.member, found.inClass));
-            CompletionList.Show(known, false);
+            known.Add(new GeneratorItem(label, GeneratorJobType.ConvertToConst, found.member, found.inClass, style));
         }
 
         private static void ShowImplementInterface(FoundDeclaration found)
         {
-            List<ICompletionListItem> known = new List<ICompletionListItem>();
             string label = TextHelper.GetString("ASCompletion.Label.ImplementInterface");
             known.Add(new GeneratorItem(label, GeneratorJobType.ImplementInterface, null, found.inClass));
-            CompletionList.Show(known, false);
         }
 
         private static void ShowNewVarList(FoundDeclaration found)
@@ -642,7 +646,6 @@ namespace ASCompletion.Completion
                     }
                 }
             }
-            List<ICompletionListItem> known = new List<ICompletionListItem>();
             string label;
             if ((exprAtCursor != null && exprAtCursor.RelClass != null && (exprAtCursor.RelClass.Flags & FlagType.Interface) > 0)
                 || (found.inClass != null && (found.inClass.Flags & FlagType.Interface) > 0))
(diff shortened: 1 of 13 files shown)


Gama11:
Great idea! :+1:

External/Plugins/ASCompletion/Completion/ASGenerator.cs
```diff
@@ -1395,7 +1363,7 @@ private static void EventMetatag(ClassModel inClass, ScintillaControl Sci, Membe
             InsertCode(position, template);
         }
 
-        private static void ConvertToConst(ClassModel inClass, ScintillaControl Sci, MemberModel member, bool detach)
+        private static void ConvertToConst(ClassModel inClass, ScintillaControl Sci, MemberModel member, bool detach, int style)
```
elsassph: You could probably re-evaluate the style at cursor position instead of doing all these changes to pass `style`.

External/Plugins/ASCompletion/Completion/ASGenerator.cs
```diff
@@ -1415,21 +1383,12 @@ private static void ConvertToConst(ClassModel inClass, ScintillaControl Sci, Mem
...
-                cr = (char)Sci.CharAt(wordPosStart - 1);
-                if (cr == '.')
-                {
-                    wordPosStart = Sci.WordStartPosition(wordPosStart - 1, true);
-                }
-            }
+            int wordPosEnd = position + 1;
+            int wordPosStart = position;
+
+            while (Sci.BaseStyleAt(wordPosEnd) == style) wordPosEnd++;
```
elsassph: Is `Sci.BaseStyleAt` safe when reaching the boundaries of the text buffer?

External/Plugins/ASCompletion/Completion/ASGenerator.cs
```diff
@@ -1459,7 +1418,23 @@ private static void ConvertToConst(ClassModel inClass, ScintillaControl Sci, Mem
...
+            {
+                case 4:
+                    m.Type = features.numberKey;
+                    break;
+                case 6:
+                case 7:
+                    m.Type = features.stringKey;
+                    break;
+                default:
+                    m.Type = features.dynamicKey ?? features.objectKey;
```
elsassph: In Haxe, if the type is unknown it's better to not generate any type annotation.

External/Plugins/ASCompletion/PluginMain.cs
```diff
@@ -446,104 +446,107 @@ public void HandleEvent(Object sender, NotifyEvent e, HandlingPriority priority)
...
+                                else if (command == "ASCompletion.CtrlAltSpace")
+                                {
+                                    ASComplete.OnShortcut(Keys.Control | Keys.Alt | Keys.Space, ASContext.CurSciControl);
+                                    e.Handled = true;
+                                }
+                                else if (command == "ASCompletion.ContextualGenerator")
+                                {
+                                    if (ASContext.HasContext && ASContext.Context.IsFileValid)
+                                    {
+                                        ASGenerator.ContextualGenerator(ASContext.CurSciControl);
```
elsassph: Might be better to avoid making `KnownList` public (although refactoring to use a static prop is fine) and instead doing something like:
```
var options = ASGenerator.ContextualGenerator(ASContext.CurSciControl);
options.AddRange(GetRefactorMenus(de.Data as ToolStripMenuItem));
CompletionList.Show(options, false);
```

External/Plugins/ASCompletion/PluginMain.cs
```diff
@@ -960,6 +963,59 @@ private void ContextChanged()
...
         #endregion
+
+        #region Contextual Code Generation Utilities
+
+        void AddRefactorMenus(List<ICompletionListItem> list, ToolStripMenuItem menu)
+        {
+            if (list == null)
+                return;
+
+            ASComplete.NotifyContextChanged();
```
elsassph: Isn't the context already updated when getting the generators?

External/Plugins/ASCompletion/PluginMain.cs
```diff
@@ -446,104 +446,107 @@ public void HandleEvent(Object sender, NotifyEvent e, HandlingPriority priority)
...
+                                {
+                                    ASComplete.OnShortcut(Keys.Control | Keys.Alt | Keys.Space, ASContext.CurSciControl);
+                                    e.Handled = true;
+                                }
+                                else if (command == "ASCompletion.ContextualGenerator")
+                                {
+                                    if (ASContext.HasContext && ASContext.Context.IsFileValid)
+                                    {
+                                        ASGenerator.ContextualGenerator(ASContext.CurSciControl);
+                                        AddRefactorMenus(ASGenerator.KnownList, de.Data as ToolStripMenuItem);
```
elsassph: It's kind of ok like that, but maybe this should be properly decoupled and using a plugin event message to share the options list, allowing effectively any plugin to add relevant items here. And so the code adding the refactor menu items could stay in the refactor plugin.

elsassph:
Really neat contribution!

External/Plugins/ASCompletion/Completion/ASGenerator.cs
```diff
@@ -41,6 +42,10 @@ public class ASGenerator
         static private MemberModel contextMember;
         static private bool firstVar;
 
+        static List<ICompletionListItem> known;
+
+        public static List<ICompletionListItem> KnownList => known;
```
elsassph: You can do THAT in C# now? Sooo neat.

Neverbirth: However it requires you having VS2015, which is not possible for everyone out there. The same as I said for using the null conditional operator.

Gama11: Why is it not possible for everyone to have?

Neverbirth: Well, there can be several reasons: not fully owning the PC being used, or not administrative rights to install it or some other locked environment, lack of available space to install it, not easy access to download it, not supported OS, etc.

elsassph: Ah I'm still running VS2013 myself.

External/Plugins/ASCompletion/Completion/ASGenerator.cs
wise0704: When the cursor is at the end of file, it seems like it returns 0 (`DEFAULT`) rather than throwing an exception. I figured it out while working on the auto closing braces.

External/Plugins/ASCompletion/PluginMain.cs
wise0704: I was halfway through doing that but then I realised there are 3.7 million `return` statements in the `ContextualGenerator` method, so I kinda went "nah cbb", which is a bad practice.
I will fix it in the next commit with the power of `Ctrl+H` however.

External/Plugins/ASCompletion/PluginMain.cs
wise0704: I'm not sure, because following the event handler list when `CodeRefactor.PluginMain` dispatches `"ASCompletion.ContextualGenerator"`, it didn't seem like there was any code for updates. The only reason of calling `NotifyContextChanged` is because it calls `ASComplete.OnResolvedContextChanged`, which calls `CodeRefactor.PluginMain.OnResolvedContextChanged` which then calls `CodeRefactor.PluginMain.UpdateMenuItems`, which is what I want before checking for the `Enabled` of each option. 
I will check if it does get validated automatically without the struggles.

However, as you suggested below, using event mechanism will clearly solve this issue.

External/Plugins/ASCompletion/PluginMain.cs
wise0704: Didn't think of that, thank you!
I couldn't really be bothered to look into the event mechanism thinking it would be very complicated, but it seems like it is very similar to AS3 style of event handling with event types as enums and optional data. 
Using `new DataEvent(EventType.Command, "CodeRefactor.AddRefactorOptions", options)` would be alright? I have no idea about the rules for `DataEvent.Action` string.

External/Plugins/ASCompletion/Completion/ASGenerator.cs
wise0704: @Neverbirth Oh lol. It seems like I have to revert a few things... also in the #947.
The power of C# 6.0 is being wasted...

External/Plugins/ASCompletion/Completion/ASGenerator.cs
wise0704: Fixed in 54bdb08.

External/Plugins/ASCompletion/Completion/ASGenerator.cs
wise0704: Fixed in 54bdb08.

External/Plugins/ASCompletion/PluginMain.cs
```diff
@@ -446,104 +446,108 @@ public void HandleEvent(Object sender, NotifyEvent e, HandlingPriority priority)
...
                             }
+                            break;
 
-                            // build the SWF using MTASC
-                            else if (command == "ASCompletion.QuickBuild")
+                        // menu commands
+                        case EventType.Command:
+                            de = e as DataEvent;
+                            string command = de.Action ?? "";
+                            if (command.StartsWith("ASCompletion.", StringComparison.Ordinal))
```
wise0704: I'm pretty sure you must explicitly define `StringComparison.Ordinal` as the second parameter for better performance.
If I'm not mistaken, by default `StringComparison.CurrentCulture` is used to compare strings which compares them in the forms of words, not in character correspondence.

External/Plugins/CodeRefactor/PluginMain.cs
```diff
@@ -655,7 +661,79 @@ public void SaveSettings()
...
+        public RefactorItem(ToolStripItem item)
+        {
+            this.item = item;
+            label = Regex.Replace(item.Text, "[&.]", string.Empty);
+            icon = new Bitmap(item.Image ?? PluginBase.MainForm.FindImage("452")); //452, 473
+        }
+
+        public string Description
+        {
+            get { return label; /*TextHelper.GetString("ASCompletion.Info.GeneratorTemplate");*/ }
```
wise0704: I don't know which is more appropriate between `label` and `TextHelper.GetString("ASCompletion.Info.GeneratorTemplate")` as the description. I decided that the refactor items would be better to have different descriptions than the original context generation options, so I used `label`.

External/Plugins/ASCompletion/Completion/ASGenerator.cs
Neverbirth: Well, such are things when it comes to new software. I even know of companies that mainly work with .NET and their main VS version is 2010.

External/Plugins/ASCompletion/PluginMain.cs
elsassph: You're right, we should be using `StringComparison.Ordinal` in all these comparisons.

External/Plugins/CodeRefactor/PluginMain.cs
elsassph: Maybe `TextHelper.GetString("Label.Refactor").Replace("&","")`?

External/Plugins/ASCompletion/PluginMain.cs
wise0704: Can I make a string extension class in `PluginCore.Utilities` with `StartsWithOrdinal`, `EndsWithOrdinal`, `IndexOfOrdinal` and `LastIndexOfOrdinal`? It would make the code much more readable.

elsassph: Sounds good, should be tidier.

wise0704: In a new commit, or this? Sorry I'm just casually noobing.

elsassph: Let's discuss that separately - merged!

PR #952, opened 2015-11-11
Inline renaming support

wise0704:
Just another small stuff I've been playing with today...
I kinda liked the new inline renaming in VS2015 so I decided I should give it a go at implementing it in FD as well (no I'm not trying to implement everything from VS. I swear...)

Some current issues: I don't know how to highlight text / set background with all the complicated Scintilla stuff.

What this does: renaming will (maybe optionally by uncommenting some lines and adding a settings option somewhere) not pop up a Rename window, but instead, the word can be edited directly from the editor. Pressing Enter will accept the changes, pressing ESC or switching to another document will cancel the renaming. 

I'm not aware of any bugs as of now, but there could be some. Idk.


External/Plugins/CodeRefactor/Commands/InlineRename.cs
```diff
@@ -0,0 +1,191 @@
...
+
+        void AddHighlight()
+        {
+            //sci.AddHighlight(8, 0x00FF00, start, end - start);
+            ////if error: sci.AddHighlight(6, 0xFF0000, start, end - start);
+        }
+
+        void RemoveHighlight()
+        {
+            //sci.RemoveHighlight(start, end - start);
```
wise0704: `AddHighlight` and `RemoveHighlight` are supposed to paint the background of the text being edited to green.
But I have no idea how to do so, even after looking though all the members of `ScintillaControl`.
Help.

External/Plugins/CodeRefactor/PluginMain.cs
```diff
@@ -418,8 +418,10 @@ private void RenameClicked(Object sender, EventArgs e)
         {
             try
             {
-                Rename command = new Rename(true);
-                command.Execute();
+                //InlineRename.CancelExisting();
+                //new Rename(useInline);
+                ////or
+                InlineRename.Start();
```
wise0704: Uncommenting the top 2 lines and deleting the last line could be a better option...
`useInline` can just be a constant value or it can be from settings.

wise0704:
Further improvement on this is to handle `InlineRename.OnUpdate(string)` and update all references in the current document as the word is being edited (rename preview).

Even further improvement would be to check for duplicated definition, that if the current word is the same as any other variable in the same scope, a red highlight box should be displayed instead of green. (Of course, the highlighting needs to be sorted out for this. Also I don't even know if this is possible anyways.)

~~h e l p~~

External/Plugins/CodeRefactor/Commands/InlineRename.cs
SlavaRa: mb it help https://github.com/SlavaRa/fdplugin-highlight-selection/blob/master/PluginMain.cs#L195-L201

SlavaRa: bm it help https://github.com/SlavaRa/fdplugin-highlight-selection/blob/master/PluginMain.cs#L178

wise0704: thanks it really helped
for my case however, adding `sci.RemoveHighlights(0)` at the beginning was the solution

wise0704:
Part 1: Added option for including comments and strings in search for renaming

Part 2: Updated the Popup Rename Dialogue to a new one
capture

Part 3: Supports Preview Changes for Inline Renaming (references are stored at the beginning of renaming, therefore does not randomly "pickup" new references during the process)
animated

There is an option to switch between PopupRenameDialog and InlineRenameDialog (popup is default).

wise0704:
A quick question: anyone know how to get the width of the scrollbar?

wise0704:
Should the option "Include comments" in Rename and FindAllReferences only include comment documentation (`///` and `/** */`) or all comments?

External/Plugins/CodeRefactor/Commands/InlineRename.cs
```diff
@@ -0,0 +1,852 @@
...
+        #region Shortcut Handlers
+
+        void PerformUndo()
+        {
+            
+        }
+
+        void PerformRedo()
+        {
+
```
wise0704: Oh lol. Fail.

External/Plugins/CodeRefactor/Commands/InlineRename.cs
```diff
@@ -0,0 +1,852 @@
...
+        void PerformUndo()
+        {
+            
+        }
+
+        void PerformRedo()
+        {
+            
+        }
```
wise0704: Oh lol. Fail.

wise0704:
Added the ~~missing~~ undo/redo actions to the inline renaming.
Also, to be compatible with #955, added support for shortcuts for `PopupRenameDialog`.
I see an improvement available to the `MainForm.GetShortcutItemKeys()` and `ShortcutManager.RegisteredItems`, but I will make changes to them in a different commit (probs after this pull request is closed, due to conflicts).

Meychi:
Has conflicts now.

wise0704:
#972 caused a major conflict with the PR, as both were modifying pretty much the same part of the code. I have fixed all conflicts, but there may be a bug caused from them.

wise0704:
For some reason the syntax check doesn't start up for meee
image

Anyone want to try this feature out and tell me what you guys think of?

Meychi:
Conflicts :)

Meychi:
I see few issues here:

- You have a hardcoded font in the dialog
- No font scaling, so in highDPI this doesn't work
- Localization comments should have when they were added
- Make sure the dialog is in the same scale (design font size)

@elsassph, @SlavaRa: What do you think, is this good for merge when finished?

elsassph:
No objection.

SlavaRa:
I will answer in the evening

SlavaRa:
No objection.

Meychi:
@wise0704: Let us know when you complete this.

External/Plugins/ASCompletion/ASCompletion.csproj
```diff
@@ -45,13 +45,13 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
-    <DebugType>none</DebugType>
+    <DebugType>pdbonly</DebugType>
```
wise0704: pfffffff

PluginCore/PluginCore/Resources/de_DE.resX
```diff
@@ -26,36 +26,36 @@
...
+
+    There are any number of "resheader" rows that contain simple
     name/value pairs.
-    
-    Each data row contains a name, and value. The row also contains a 
-    type or mimetype. Type corresponds to a .NET class that support 
-    text/value conversion through the TypeConverter architecture. 
-    Classes that don't support this are serialized and stored with the 
+
+    Each data row contains a name, and value. The row also contains a
```
wise0704: I don't know why this has happened to every file...
Does this matter? This would reformat when someone edits them with VS right?

wise0704:
@Meychi I'm not sure what you mean by "Make sure the dialog is in the same scale (design font size)".

elsassph:
@Meychi can we merge that if the conflicts are solved by @wise0704 ?

wise0704:
This branch currently has some bugs from recent changes made to development branch. Fixes will be coming soon.

wise0704:
Bug fixed. Also now the haxe getter/setter method names also update during inline rename preview.
image

elsassph:
Not so friendly with high-dpi...
image

wise0704:
The highlight color, transparency and style can be changed easily (as a draft I just chose the green opaque highlight to match with VS).

wise0704:
"Include comments" and "Include strings" were added to match with VS, but to be honest I've never used those features in VS, and I don't really see any reason to use "Include strings" option at all.

"Include comments" can be more useful by changing `RefactoringHelper.IsCommentStyle()` to only return true on `CommentDoc` and `CommentLineDoc`.

External/Plugins/CodeRefactor/Commands/Rename.cs
```diff
@@ -5,119 +5,160 @@
...
+            return Create(RefactoringHelper.GetDefaultRefactorTarget(), inline, outputResults);
         }
 
         /// <summary>
         /// A new Rename refactoring command.
         /// </summary>
+        /// <param name="inline">Whether to use inline renaming.</param>
         /// <param name="target">The target declaration to find references to.</param>
         /// <param name="outputResults">If true, will send the found results to the trace log and results panel</param>
-        public Rename(ASResult target, Boolean outputResults)
```
SlavaRa: I think it's a bad idea, because all other commands look different. and I do not understand why it was necessary to change the constructors may be back as it was?

wise0704: I don't remember why I changed the constructors...
I'll check again and probs put them back.

elsassph:
Would it be possible to extract just the inline renaming (without in-comment and in-string options)? We could have a setting to enable inline VS regular (as it is now) refactor-rename.

wise0704:
Yep sure it's possible.

wise0704:
@elsassph Removed the dialogs. Also now you can edit on any highlighted references.

One minor bug is, sometimes after an inline rename, pressing contextual generator doesn't include Rename, until I move the caret to a different place.

Neverbirth:
IIRC then it means that after doing the inline rename you should refresh the context.

wise0704:
`ASContext.Context.UpdateCurrentFile(true)`?

Neverbirth:
Yes, although maybe passing false would be enough in this case.

External/Plugins/CodeRefactor/Provider/RenamingHelper.cs
```diff
@@ -1,141 +1,166 @@
...
+
+        private static void RestoreStartState()
+        {
+            int pos = startState.CursorPosition;
+            GetOffset(startState.Commands[0], ref pos);
+            GetOffset(startState.Commands[1], ref pos);
+            GetOffset(startState.Commands[2], ref pos);
+            ((ITabbedDocument) PluginBase.MainForm.OpenEditableDocument(startState.FileName)).SciControl.SetSel(pos, pos);
+
+            ASContext.Context.UpdateCurrentFile(true);
```
wise0704: @Neverbirth It's called here... and this method is the last being called during a rename operation. Any idea?

Neverbirth:
I see, then I would say the issue lays in the CodeRefactor plugin part. I guess OnResolvedContextChanged is not triggered in this case.

External/Plugins/CodeRefactor/Provider/RenamingHelper.cs
Neverbirth: I see, then I would say the issue lays in the CodeRefactor plugin part. I guess OnResolvedContextChanged is not triggered in this case.

wise0704: I tried to track the method calls with Go to Definition, but fell into the loop of doom...
I guess I could try to find a cleaner way for the inline rename, like disable all ASContext, ASCompletion and Scintilla functions during rename. 
Because how it currently works is that everything gets reverted back to how it was before inline rename, and then the normal process for Rename is done.

PR #953, opened 2015-11-12
[Haxe] Now "Find All References" and "Rename" works with String Interpolation

SlavaRa:
Result:
rename_expr_in_string_interpolation


elsassph:
We should not be adding this kind of feature in FD built-in engine and move to leveraging the Haxe compiler service to find references.

PR #954, opened 2015-11-17
Improved ScrollBarEx

wise0704:
[https://github.com/wise0704/flashdevelop/tree/ScrollBarEx]

Started on this branch while working on #952. I needed to get the references of the scroll bars of `ScintillaControl`, and ended up working on some more stuff.

- More event based coloring options for `ScrollBarEx`. (Independent arrow colors, color change on mouse over.)*
- Fixed the bug of the `ScrollBarEx` always snapping its centre to mouse regardless of the click position.
- Changed default colors for `ScrollBarEx` to match the scroll bars from Visual Studio.
- Modified `ScintillaControl` to only use `ScrollBarEx`, rather than switching back and forth from system scroll bar according to the presence of explicit theme color defined.

_'Ex' stands for Extended? Nah bro, it stands for Exclusive. Why only use ScrollBarEx when theme colors are explicitly defined, when it looks wayy cooler._

*It is backward compatible, as if a theme setting does not contain values for the newly introduced colors, they roll back to match as if there were no new color options - for example, if a setting does not define `HotForeColor`, which is the color on mouse over, it will be set to `ForeColor`, as before this commit there were no color change for mouse over. (_`ScrollBarEx.ValidateColors()` must be called however to achieve this._)

### DIFF
--- a/PluginCore/PluginCore/Controls/ScrollBarEx.cs
+++ b/PluginCore/PluginCore/Controls/ScrollBarEx.cs
@@ -32,7 +32,7 @@ namespace PluginCore.Controls
         private Color backColor/* = SystemColors.ActiveBorder*/;
         private Color backColorDisabled/* = SystemColors.ControlLight*/;
 
-        private bool colorsValidated;
+        private bool colorsInvalidated;
 
         /// <summary>
         /// Resets the component colors to default values.
@@ -50,18 +50,21 @@ namespace PluginCore.Controls
             arrowColorPressed = SystemColors.HotTrack;
             backColor = SystemColors.Control;
             backColorDisabled = SystemColors.ControlLight;
-            colorsValidated = true;
+            colorsInvalidated = false;
             Invalidate();
         }
 
         /// <summary>
-        /// Validates any unassigned colors to either default to associated colors.
+        /// Validates any unassigned colors to either default or to associated colors.
         /// Calling this method without assigning any new colors will set all colors to default.
+        /// <para/>
+        /// Call this method after the theme has been changed.
         /// </summary>
         public void ValidateColors()
         {
-            if (colorsValidated)
+            if (!colorsInvalidated)
             {
+                // No colors defined explicitly -> Reset colors to default.
                 InitializeColors();
                 return;
             }
@@ -73,20 +76,22 @@ namespace PluginCore.Controls
             if (foreColorPressed.IsEmpty) foreColorPressed = SystemColors.ControlDarkDark;
             if (backColor.IsEmpty) backColor = SystemColors.Control;
             if (backColorDisabled.IsEmpty) backColorDisabled = SystemColors.ControlLight;
-            // new colors
+            // Newly introduced color options - do not assign default colors. Instead fall back to associated colors
             if (foreColorHot.IsEmpty) foreColorHot = foreColor;
             if (arrowColor.IsEmpty) arrowColor = foreColor;
             if (arrowColorHot.IsEmpty) arrowColorHot = foreColorHot;
             if (arrowColorPressed.IsEmpty) arrowColorPressed = foreColorPressed;
 
-            colorsValidated = true;
+            colorsInvalidated = false;
         }
 
         /// <summary>
-        /// Resets all colors to null values.
+        /// Resets all validated colors to null values. Call this method before setting new colors.
         /// </summary>
         private void ResetColors()
         {
+            if (colorsInvalidated) return;
+
             curPosColor =
             borderColor =
             borderColorDisabled =
@@ -99,7 +104,7 @@ namespace PluginCore.Controls
             backColor =
             backColorDisabled =
             Color.Empty;
-            colorsValidated = false;
+            colorsInvalidated = true;
         }
 
         /// <summary>
@@ -237,11 +242,11 @@ namespace PluginCore.Controls
         }
 
         /// <summary>
-        /// Draws the thumb.
+        /// Draws the vertical thumb.
         /// </summary>
         /// <param name="g">The <see cref="Graphics"/> used to paint.</param>
         /// <param name="rect">The rectangle in which to paint.</param>
-        /// <param name="state">The <see cref="ScrollBarState"/> of the thumb.</param>
+        /// <param name="color">The color to draw the thumb with.</param>
         private void DrawThumbVertical(Graphics g, Rectangle rect, Color color)
         {
             var innerRect = new Rectangle(rect.Left + ScaleHelper.Scale(2), rect.Top, rect.Width - ScaleHelper.Scale(4), rect.Height);
@@ -253,11 +258,11 @@ namespace PluginCore.Controls
         }
 
         /// <summary>
-        /// Draws the thumb.
+        /// Draws the horizontal thumb.
         /// </summary>
         /// <param name="g">The <see cref="Graphics"/> used to paint.</param>
         /// <param name="rect">The rectangle in which to paint.</param>
-        /// <param name="state">The <see cref="ScrollBarState"/> of the thumb.</param>
+        /// <param name="color">The color to draw the thumb with.</param>
         private void DrawThumbHorizontal(Graphics g, Rectangle rect, Color color)
         {
             var innerRect = new Rectangle(rect.Left, rect.Top + ScaleHelper.Scale(2), rect.Width, rect.Height - ScaleHelper.Scale(4));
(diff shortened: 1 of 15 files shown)


elsassph:
I think @Meychi had kept the default scrollbars for light theme in case some people had issues with the fancy ones - hopefully now they are good to  apply all the time :)

wise0704:
Yep they should be. The new default colors are for light theme:
default_defaultdefault_hotdefault_presseddefault_arrowhotdefault_arrowpresseddefault_disabled
These are default, thumb mouse over, thumb down, arrow mouse over, arrow down and disabled in the order. Pretty much the same as Visual Studio. They all use `SystemColor` values.
If there aren't any bugs other than the one I fixed, I think this scroll bar is algoods to use all the time.

PluginCore/PluginCore/Controls/ScrollBarEx.cs
```diff
@@ -20,13 +20,92 @@ public class ScrollBarEx : Control
...
+            borderColorDisabled =
+            foreColor =
+            foreColorHot =
+            foreColorPressed =
+            arrowColor =
+            arrowColorHot =
+            arrowColorPressed =
+            backColor =
+            backColorDisabled =
+            Color.Empty;
```
wise0704: some weird as syntax... sure it does seem like assigning "null" values...

Meychi:
We need to keep system ones around as some people may prefer OS ones and performace reasons for example in CrossOver. 

The default theme is fully native with no theming but we need to think if we should default to a different theme. Can we agree which theme it should be? And we need to test it widely.

wise0704:
For keeping the system scroll bar, how about making a settings option "use custom scroll bar" with the values `On` `Off` and `Auto`, where `Auto` means use `ScrollBarEx` only when the current theme defines a colouring option for scroll bars?

wise0704:
About the default theme, I personally stick with the current default, or sometimes use DimGray or Obsidian. I think the current default is alright to keep, since changing theme is really easy with FlashDevelop for those who want to use different themes.

However, if a change of the default theme feels necessary, we could make a poll in the community forums to gather in other opinions and feedback.

Meychi:
My intention is to move all UI related settings to themes so that they have full control how things looks and behaves.

Meychi:
I still want to keep the previous functionality so that themes controls this. Could you make a new PR or revise this with the other improvements and fixes?

wise0704:
I get what you mean now :)

But I'm wondering _(my suggestion)_, since the function of `ScrollBarMode.Auto`, which is the default, is exactly the same as the previous functionality, wouldn't it be better to let users be able to turn on/off the custom scroll bars easily and temporarily through settings instead of locating and opening the theme files, and making changes to the files and having to refresh the theme? 

Considering the issues you mentioned such as performance; in case the scroll bars are lowering the performance only under specific conditions, you wouldn't want to make changes to the theme itself or make duplicates of the theme and having to switch between them just for the scroll bars.

wise0704:
Anyways, can I modify the default theme files to contain these values?
```
ScrollBar.UseCustom=False
#ScrollBar.HotForeColor=#888888
#ScrollBar.ArrowColor=#868999
#ScrollBar.HotArrowColor=#1C97EA
#ScrollBar.ActiveArrowColor=#007ACC
```
I think it would be better to judge whether to use the custom scroll bar based on the value of something like `ScrollBar.UseCustom` rather than `ScrollBar.ForeColor`.

Meychi:
I'm thinking that providing a simple dialog to adjust current theme values similar to syntax dialog would solve those issues.

Yes you can, i will adjust and finish the other themes and other things related to it.

Using one setting to enable scrollbars sounds good to me.

wise0704:
For `Obsidian.fdi` and `Thyleus.fdi`, I added the extra colors with values `#XXXXXX`.

I figured out that `SystemColor` values vary a lot depending on your machine, and that what I see on my computer may be quite different from others, so I decided to set the default color values of the scroll bar with explicit numeral values. 

I thought that it would not be necessary to add `"HotForeColor"`, `"ArrowColor"` and so on to `ThemeManager.ThemeControl()` method just for the scroll bar, so I made the scroll bar retrieve these values internally inside `ValidateColors()` methods.

PluginCore/PluginCore/Controls/ScrollBarEx.cs
```diff
@@ -18,15 +18,107 @@ namespace PluginCore.Controls
...
+        /// Calling this method without assigning any new colors will set all colors to default.
+        /// <para/>
+        /// Call this method after the theme has been changed.
+        /// </summary>
+        public void ValidateColors()
+        {
+            // Reset any unassigned colors to default.
+            InitializeColors();
+
+            // If no colors are explicitly defined, not fallback colors are required.
```
wise0704: no* fallback colors......

Meychi:
I meant that we should have one theme setting and not a real setting as I'm moving all ui settings to themes.

Meychi:
Do these changes fallback correctly with older themes not having all the values?

wise0704:
Yep, tested with the existing themes and also the ones I have (VsBlue, VsDark and VsLight).
[These lines](https://github.com/fdorg/flashdevelop/pull/954/files#diff-c1e38f09df3f1b3729ebacba8650c7a2R79) do the job which makes
 * 'fore color hover' falls back to 'fore color' as previously no color change on mouse over occurred
 * 'arrow color' falls back to 'fore color' as previously no color difference between thumb and arrow existed
 * 'arrow color hot' falls back to 'arrow color'
 * 'arrow color active' falls back to 'fore color active'.

Meychi:
Seems the fallbacks don't render properly. By removing:

if (!colorsInvalidated) return;

from ScollBarEx (around line 77) it works fine. Can you confirm?

Meychi:
Also i started thinking that could the Enum.Parse cause perf bottlenecks instead of validating strings directly?

Meychi:
Also the default fallbacks are now hardcoded in the ScrollBarEx, which is not great.

PluginCore/PluginCore/Controls/ScrollBarEx.cs
```diff
@@ -18,15 +18,95 @@ namespace PluginCore.Controls
...
+            if (!colorsInvalidated)          foreColorHot        = Color.FromArgb(136, 136, 136); //#888888
+            if (foreColorPressed.IsEmpty)    foreColorPressed    = Color.FromArgb(106, 106, 106); //#6A6A6A
+            if (!colorsInvalidated)          arrowColor          = Color.FromArgb(134, 137, 153); //#868999
+            if (!colorsInvalidated)          arrowColorHot       = Color.FromArgb( 28, 151, 234); //#1C97EA
+            if (!colorsInvalidated)          arrowColorPressed   = Color.FromArgb(  0, 122, 204); //#007ACC
+            if (backColor.IsEmpty)           backColor           = Color.FromArgb(232, 232, 236); //#E8E8EC
+            if (backColorDisabled.IsEmpty)   backColorDisabled   = Color.FromArgb(192, 192, 192); //#C0C0C0
+            if (borderColor.IsEmpty)         borderColor         = Color.FromArgb(232, 232, 236); //#E8E8EC
+            if (borderColorDisabled.IsEmpty) borderColorDisabled = Color.FromArgb(136, 136, 136); //#888888
+            colorsInvalidated = false;
```
wise0704: `colosInvalidated = false` causes the fallback colors not being assigned.
I remember changing the structure of `ValidateColors` after testing the themes :(
I'll remove this line.

wise0704:
The `Enum.Parse` is part of #974. 
In one of my previous comments I talked about how the `SystemColor` values could be very different depending on your OS and personal settings.
However, if you want them to be `KnownColor` values, I will try to find equivalent or similar from `Color` struct or revert back to `SystemColor` values.

Meychi:
I'm now working on this, no need to do anything.

wise0704:
Just a note.
The equivalent properties still have the old `SystemColor` values I assigned to them previously as parameters to `DefaultValueAttribute()`. You may want to use them if you like.

PluginCore/PluginCore/Controls/ScrollBarEx.cs
```diff
@@ -18,15 +18,95 @@ namespace PluginCore.Controls
     [DefaultProperty("Value")]
     public class ScrollBarEx : Control
     {
+        #region scroll bar settings
+
+        public static bool UseCustom
+        {
+            get { return PluginBase.MainForm.GetThemeValue("ScrollBar.UseCustom", "false").ToLower() == "true"; }
+        }
```
wise0704: This line can now be changed to:
```C#
get { return PluginBase.MainForm.GetThemeFlag("ScrollBar.UseCustom"); }
```

PR #955, opened 2015-11-17
[CodeRefactor] Support for hot keys are converting to uppercase, lowercase in rename form

SlavaRa:
Uses hot keys EditMenu.ToLowercase and EditMenu.ToUppercase as when editing code
Result:
coderefactor_tolowercase_touppercase_in_rename_form


elsassph:
Well that's a secret shortcut... If we were to add this shortcut then I see no reason not to add it the original LineRefactor class and not as a custom feature of the refactor dialog.

PR #963, opened 2015-11-30
Haxe type parameters improvements

Neverbirth:
When displaying classes check for type parameters and display them if existing.

typeparams1

When autocompleting after : check in functions for the existance of type parameters. Previously this case was breaking completion completely.

typeparams2

The first screenshot also shows this fix, before the completion list wouldn't be shown in there unless forced with Ctrl+Alt+Space.

### DIFF
--- a/External/Plugins/ASCompletion/Completion/ASComplete.cs
+++ b/External/Plugins/ASCompletion/Completion/ASComplete.cs
@@ -2331,32 +2331,35 @@ namespace ASCompletion.Completion
                     known = found;
                 }
             }
-            
-            if (ASContext.Context.Features.hasDelegates && !ASContext.Context.CurrentClass.IsVoid())
+
+            if (!ASContext.Context.CurrentClass.IsVoid())
             {
-                MemberList delegates = new MemberList();
-
-                foreach (MemberModel field in ASContext.Context.CurrentClass.Members)
-                    if ((field.Flags & FlagType.Delegate) > 0)
-                        delegates.Add(field);
-
-                if (delegates.Count > 0)
+                if (ASContext.Context.Features.hasDelegates)
                 {
-                    delegates.Sort();
-                    delegates.Merge(known);
-                    known = delegates;
+                    MemberList delegates = new MemberList();
+
+                    foreach (MemberModel field in ASContext.Context.CurrentClass.Members)
+                        if ((field.Flags & FlagType.Delegate) > 0)
+                            delegates.Add(field);
+
+                    if (delegates.Count > 0)
+                    {
+                        delegates.Sort();
+                        delegates.Merge(known);
+                        known = delegates;
+                    }
                 }
-            }
 
-            if (ASContext.Context.Features.hasGenerics && !ASContext.Context.CurrentClass.IsVoid())
-            {
-                var typeParams = GetVisibleTypeParameters();
-
-                if (typeParams != null && typeParams.Items.Count > 0)
+                if (ASContext.Context.Features.hasGenerics)
                 {
-                    typeParams.Add(known);
-                    typeParams.Sort();
-                    known = typeParams;
+                    var typeParams = GetVisibleTypeParameters();
+
+                    if (typeParams != null && typeParams.Items.Count > 0)
+                    {
+                        typeParams.Add(known);
+                        typeParams.Sort();
+                        known = typeParams;
+                    }
                 }
             }
 
--- a/External/Plugins/ASCompletion/Completion/ASComplete.cs
+++ b/External/Plugins/ASCompletion/Completion/ASComplete.cs
@@ -2334,16 +2334,25 @@ namespace ASCompletion.Completion
             
             if (ASContext.Context.Features.hasDelegates && !ASContext.Context.CurrentClass.IsVoid())
             {
+                MemberList delegates = new MemberList();
+
                 foreach (MemberModel field in ASContext.Context.CurrentClass.Members)
                     if ((field.Flags & FlagType.Delegate) > 0)
-                        known.Add(field);
+                        delegates.Add(field);
+
+                if (delegates.Count > 0)
+                {
+                    delegates.Sort();
+                    delegates.Merge(known);
+                    known = delegates;
+                }
             }
 
             if (ASContext.Context.Features.hasGenerics && !ASContext.Context.CurrentClass.IsVoid())
             {
                 var typeParams = GetVisibleTypeParameters();
 
-                if (typeParams.Items.Count > 0)
+                if (typeParams != null && typeParams.Items.Count > 0)
                 {
                     typeParams.Add(known);
                     typeParams.Sort();
@@ -3819,8 +3828,8 @@ namespace ASCompletion.Completion
 
         static private MemberList GetTypeParameters(MemberModel model)
         {
-            var retVal = new MemberList();
-            var template = model.Template;
+            MemberList retVal = null;
+            string template = model.Template;
             if (template != null && template.StartsWith("<"))
             {
                 var sb = new StringBuilder();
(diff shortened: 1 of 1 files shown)


External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -3777,6 +3817,94 @@ static public ASResult GetExpressionType(ScintillaControl sci, int position, boo
             }
         }
 
+        static private MemberList GetTypeParameters(MemberModel model)
+        {
+            var retVal = new MemberList();
```
elsassph: Maybe avoid this instantiation if the type isn't a generic type?

Neverbirth: It was mainly done that way for readability, but sure, no problem.

I also found a problem in other part of code. PR in a few minutes.

Neverbirth:
Done. Also, as said in the commit. I realized that if the class has delegates (I think this is Loom?) and there are no class vars, the cached list would be corrupted with delegates, and adding more entries over time.

External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -3777,6 +3826,100 @@ static public ASResult GetExpressionType(ScintillaControl sci, int position, boo
...
+        }
+
+        static private MemberList GetVisibleElements()
+        {
+            MemberList known = new MemberList();
+            known.Add(ASContext.Context.GetVisibleExternalElements());
+
+            if (ASContext.Context.Features.hasGenerics && !ASContext.Context.CurrentClass.IsVoid())
+            {
+                known.Merge(GetVisibleTypeParameters());
```
elsassph: `GetVisibleTypeParameters` can return `null`.
Maybe that was a bad idea to avoid this allocation - makes the code needlessly complicated.

External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -2318,12 +2331,33 @@ static public void HandleAllClassesCompletion(ScintillaControl Sci, string tail,
                     known = found;
                 }
             }
-
+            
             if (ASContext.Context.Features.hasDelegates && !ASContext.Context.CurrentClass.IsVoid())
```
elsassph: Maybe factorize `!ASContext.Context.CurrentClass.IsVoid()`?

External/Plugins/ASCompletion/Completion/ASComplete.cs
Neverbirth: Yes, but MemberList.Merge has already a check for null, so no need for checks there.

External/Plugins/ASCompletion/Completion/ASComplete.cs
Neverbirth: What do you mean by factorizing it?

External/Plugins/ASCompletion/Completion/ASComplete.cs
elsassph: Ah right.

External/Plugins/ASCompletion/Completion/ASComplete.cs
elsassph: Adding delegate info and generics info both depend on being inside a valid class.

elsassph:
BTW not sure to follow what you said about Loom delegates issue.

External/Plugins/ASCompletion/Completion/ASComplete.cs
Neverbirth: Right, I didn't do it just to mark a distinction between both things. But I'll do it.

Neverbirth:
About delegates:

            if (ASContext.Context.Features.hasDelegates && !ASContext.Context.CurrentClass.IsVoid())
            {
                foreach (MemberModel field in ASContext.Context.CurrentClass.Members)
                    if ((field.Flags & FlagType.Delegate) > 0)
                        known.Add(field);
            }

`known` comes from `ASContext.Context.GetAllProjectClasses()` which is a cached instance reused always it's needed (unless `showClassVars` is true and there are class vars), so if there are delegates, they are added to that cached list. And each time `HandleAllClassesCompletion` is called it adds again the delegates.

External/Plugins/ASCompletion/Completion/ASComplete.cs
Neverbirth: Done.

External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -2319,11 +2332,35 @@ static public void HandleAllClassesCompletion(ScintillaControl Sci, string tail,
...
+                    }
+                }
+
+                if (ASContext.Context.Features.hasGenerics)
+                {
+                    var typeParams = GetVisibleTypeParameters();
+
+                    if (typeParams != null && typeParams.Items.Count > 0)
+                    {
+                        typeParams.Add(known);
```
elsassph: Merge?

Also it looks like you avoided the duplication of items in the types collection so that's good.

Neverbirth: I don't know, did you ever make benchmarks between merge and a sort? I vote for Sort being faster, but I don't know at all. In the end the list should be equally sorted, shouldn't it?

elsassph: Merging 2 sorted lists should be the fastest - it does a single iteration of the lists and inserts the items where they belong.

External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -3777,6 +3829,100 @@ static public ASResult GetExpressionType(ScintillaControl sci, int position, boo
...
+                }
+            }
+
+            return retVal;
+        }
+
+        static private MemberList GetVisibleElements()
+        {
+            MemberList known = new MemberList();
+            known.Add(ASContext.Context.GetVisibleExternalElements());
```
elsassph: A bit wasteful to create a copy of the visible elements even when no generics are involved...

External/Plugins/ASCompletion/Completion/ASComplete.cs
Neverbirth: However there are Insert and Remove calls, so depending on the collection size, where the items need to be placed and which types of items I wouldn't be surprised if the Sort method was faster. But well, then there is the overhead of calling the comparer function and all of that. I'd be truly curious.

External/Plugins/ASCompletion/Completion/ASComplete.cs
Neverbirth: Since it uses AddRange it shouldn't be very noticeable, but will change it. It's also a readability thing.

Neverbirth: Made the change.

External/Plugins/ASCompletion/Completion/ASComplete.cs
Neverbirth: Changed.

External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -3777,6 +3829,107 @@ static public ASResult GetExpressionType(ScintillaControl sci, int position, boo
...
+                        else if (")}]>".IndexOf(c) > -1)
+                            groupCount--;
+                        sb.Append(c);
+                    }
+                }
+                if (sb.Length > 0)
+                {
+                    if (retVal == null) retVal = new MemberList();
+                    if (!inConstraint)
+                        retVal.Add(new MemberModel { Name = sb.ToString(), Type = sb.ToString(), Flags = FlagType.TypeDef });
```
elsassph: Is that VS 2013 compatible?

Neverbirth: Yes, IIRC even 2008.

elsassph: I'm so old school :D

PR #968, opened 2015-12-02
Small updates to BuildEventVars

wise0704:
[https://github.com/wise0704/flashdevelop/tree/BuildEventVars]

Added support for `$(UserAppDir)` in the Pre-/Post-Build Arguments. 
Fixed `$(FDBuild)` and `$(ToolsDir)` showing wrong values in Project Properties dialog.

**Before**
image
**After**
image

### DIFF
--- a/External/Plugins/ProjectManager/Projects/BuildEventVars.cs
+++ b/External/Plugins/ProjectManager/Projects/BuildEventVars.cs
@@ -44,6 +44,7 @@ namespace ProjectManager.Projects
             infos.Add(new BuildEventInfo("FDBuild", FDBuild));
             infos.Add(new BuildEventInfo("ToolsDir", ToolsDir));
             infos.Add(new BuildEventInfo("TimeStamp", DateTime.Now.ToString("g")));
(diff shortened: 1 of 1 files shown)


elsassph:
Seems correct but @Meychi should review it.

Meychi:
The concept is correct but we should provide $(BaseDir) instead. The user tools dir is incorrect if FD is running standalone mode. The correct base path for a template or FDZ would be $(BaseDir)/Tools/. Basicly prefering user tools dir but for standalone mode using app tools dir.

wise0704:
So `UserAppDir` (after being renamed to `BaseDir`) property should check for the existence of `.local` file, right (`.local` exists means `$(AppDir)`, otherwise `$(UserAppDir)`)?
Is the change to `FDBuild` still correct?

Meychi:
Hmm... Let me finish this, i'm not sure if these are consistent with the built in path expansions.

wise0704:
Alright, sweet as. Do I have to do something to hand over or whatever? I'm very, very new to GitHub. ~~noob~~

PR #847, opened 2015-09-01
Cleanup of dependencies and small refactorings

SlavaRa:
I understand that the PR does not bear functional fixes or improvements, but it makes the code more uniform and cleaner

PR #969, opened 2015-12-03
Ordinal String Comparison

wise0704:
[https://github.com/wise0704/flashdevelop/tree/StringComparison_Ordinal]

The default behaviours of 
```C#
string.StartsWith(string)
string.EndsWith(string)
string.IndexOf(string)
string.IndexOf(string, int)
string.IndexOf(string, int, int)
string.LastIndexOf(string)
string.LastIndexOf(string, int)
string.LastIndexOf(string, int, int)
```
are to call the equivalent overloaded methods with `StringComparison.CurrentCulture` as the last parameter. These are in fact much slower than the ordinal (character correspondence) string comparisons. 

The solution I've provided is to have a string extension class in `PluginCore` namespace that provides the extension methods corresponding to the ones above, with the suffix `Ordinal` added to them (e.g. `string.StartsWithOrdinal(string)`). 

I see no point of using culture-relevant comparisons in any case for FlashDevelop unless I've missed something, so I changed all ~~**589!!!!!**~~ occurrences of those above, with either the extension methods, or in case of `IndexOf` and `LastIndexOf` with length-1 strings, to `char`'s. ~~_I'm dying after that_~~.
E.g.
```C#
str.StartsWith(str2) -> str.StartsWithOrdinal(str2)
str.IndexOf("#") -> str.IndexOf('#')
```

From now on, can everyone please be careful not to use the implicit culture-specific methods of string? There are 2 ways to avoid them:
 * Explicitly specify `StringComparison.Ordinal` every time.
 * Use my extension ~~coz i need someone to appreciate me changing a total of 589 references~~.

EDIT: this all started from [this](https://github.com/fdorg/flashdevelop/pull/950#discussion_r44503699)...


wise0704:
https://msdn.microsoft.com/en-us/library/dd465121.aspx

Meychi:
Sounds good but i need review this in detail so that i doesn't cause issues with localization.

wise0704:
Yep, sure :) I went halfway ~~(not even close to halfway hahaha)~~ doing that but as I don't know that much about all parts of the code and since there are too many, I gave up. 
I got a feeling that this PR will beat the record of the longest PR being open for this repository.

elsassph:
@wise0704 you'll have a hard time beating this one https://github.com/fdorg/flashdevelop/pull/847 ;)

wise0704:
Hahaha omg that's twice the size of mine!! That's some intense labour, man. 
image
Even GitHub refuses to list them. Well done, @SlavaRa.

Meychi:
This PR now has conflicts, im ok for merge if @elsassph agrees.

elsassph:
I'm all for it.

wise0704:
Uhm hold on. I'm just adding `string.StartsWith(char)` and `string.EndsWith(char)` extensions.

elsassph:
You want to beat @SlavaRa?

wise0704:
Hahaha no no no. Just for better performance. Yes just for performance.

Meychi:
How much performance improvements we are looking at here? For example when doing 1000 ops?

wise0704:
In general, ordinal comparisons are around 10 times faster than using current culture, and I've seen a lot of posts made about it in MSDN and Stack Overflow.
Since `StartsWith()`, `EndsWith()`, `IndexOf()`, `LastIndexOf()` methods are used frequently, especially when it comes to `ScintillaControl` and code completion codes, converting these to ordinal would definitely increase the performance (although currently on fast computers you wouldn't be able to "see" the difference).

Meychi:
I think unless we can see the difference in average computer we should not do any big changes as it may introduce new issues which might be hard to spot.

elsassph:
Doesn't seem too risky to me.

Neverbirth:
Just an idea, @wise0704 said he is on a slow computer, so he should be able to easily make a profiling session between an unmodified FD version and this one while using the parts of the application that make the most use of the modified code, and then tell the results with the profiling reports.

wise0704:
I'll just do a quick simple performance testing. Can't do all the fancy stuff.

perftest 1 - startswith
`StartsWith` performs approximately **9 times** faster with `Ordinal`.

perftest 2 - indexof
`IndexOf` performs approximately **twice** faster with `Ordinal`

perftest 3 - ignorecase equality
`Equals` with `Ordinal` performs approximately ~~OVER NINE THOUSAND!!!~~ **40~80 TIMES FASTER**.

~~"not good enough quality in your presentation, man. get gud"~~

elsassph:
Culture sensitive ToLower/ToUpper is certainly very expensive, and we use it quite a lot.

Neverbirth:
The speed when using the culture depends on the characters inside the string.

Profiling the application for something like this would be fast and more realistic. Some of the paid versions of VS have the option, if not there are some profilers out there that you can use in trial mode.

When using ToLower or ToUpper the best option is to use ToUpperInvariant, unless you do care about the culture, because some languages also have different characters for lower case and upper case.

wise0704:
@elsassph I described why ToLower/ToUpper comparisons for ignoring case is in fact quite expensive in the comments of #979.

Meychi:
I trust your evalution of the speed differences but i'm more intrested in the real world perf difference with FD. We are not doing 1mil ops for comparisons.

wise0704:
I don't really know how to do all the profiling and thingy @Neverbirth mentioned... Anyone keen to do that for me? hehe

It seems that 1 million operations makes about 0.4 seconds difference, so a few thousand operations will take a few 0.4-milliseconds difference.
One of the huge advantages of FD over any other IDE is the "Fast & Lightweight" part, and I want to contribute what I can to keep that. 

Unless there is any case FD needs to do `"\u0130" == "I"` or `"\u0131" == "i"` kind of comparisons (FD is not a word processor) the changes will not break anything.

elsassph:
Well arguably we could have an issue if we started supporting accented characters for variable names!

wise0704:
@elsassph hahaha exactly!

Neverbirth:
I have that in one of my branches, but only for parsing, because accented and other Unicode chars are 100% valid for AS and Haxe.

wise0704:
hold on... what? you can have accented chars for as3 identifiers? never knew when i was using as3
well, still the completion and stuff checks for keywords, and hopefully as3 and haxe doesn't support accented keywords.

Neverbirth:
No, there are no keywords with accented characters, but it would be entirely possible for them. For fully supporting members with unicode characters several things need to be changed, but I modified the parser to at least support that case (I don't have intentions of making the other changes).

wise0704:
Well, to think of it, still a non-culture-specific comparisons **must** be done for the identifiers.

If unicode is supported for identifiers, say `a\u0130` is a valid identifier, it should be distinguished from `ai`, right? Culture-specific comparisons will say true for environments in Turkish.

Neverbirth:
Yep, you are right.

Meychi:
Conflicts!

Meychi:
Should this be merged? Please comments folks.

elsassph:
Yes

Neverbirth:
As I told you before, what I checked was fine. But I couldn't check every single case... if something breaks you can blame it to the 64bits part ;P.

Meychi:
@Neverbirth, ok i blame you :)

Neverbirth:
Haha, ok, if it's just blaming I'm ok with it :P.

PR #972, opened 2015-12-03
closes #237

### LINKED ISSUE #237: [Haxe] Refactor rename should update getters/setters
Currently if you rename a property with a getter/setter, all instances get updated, but the getter/setter for the property do not get renamed. Seeing as in Haxe 3 these will always be get_propertyName and set_propertyName, could these also be updated when a property is renamed?

SlavaRa:
Result:
rename_haxe_property

### DIFF
--- a/External/Plugins/CodeRefactor/Provider/RenamingHelper.cs
+++ b/External/Plugins/CodeRefactor/Provider/RenamingHelper.cs
@@ -46,7 +46,7 @@ namespace CodeRefactor.Provider
 
         static void RenameMember(ClassModel inClass, string name, string newName, bool outputResults)
         {
-            MemberModel m = inClass.Members.Cast<MemberModel>().FirstOrDefault(it => it.Name == name);
+            MemberModel m = inClass.Members.Items.FirstOrDefault(it => it.Name == name);
             if (m == null) return;
             ASResult result = new ASResult();
             ASComplete.FindMember(name, inClass, result, FlagType.Dynamic | FlagType.Function, 0);
--- a/External/Plugins/CodeRefactor/Provider/RenamingHelper.cs
+++ b/External/Plugins/CodeRefactor/Provider/RenamingHelper.cs
@@ -49,6 +49,7 @@ namespace CodeRefactor.Provider
             if (m == null) return;
             ASResult result = new ASResult();
             ASComplete.FindMember(name, inClass, result, FlagType.Dynamic | FlagType.Function, 0);
+            if (result.Member == null) return;
             queue.Add(new Rename(result, outputResults, newName));
         }
 
--- a/External/Plugins/CodeRefactor/Provider/RenamingHelper.cs
+++ b/External/Plugins/CodeRefactor/Provider/RenamingHelper.cs
@@ -1,0 +1,78 @@
+﻿using CodeRefactor.Commands;
+using PluginCore.FRService;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Windows.Forms;
+using ASCompletion.Completion;
+using ASCompletion.Context;
+using ASCompletion.Model;
+using PluginCore.Localization;
+using PluginCore.Managers;
+using ProjectManager.Helpers;
+
+namespace CodeRefactor.Provider
+{
+    class RenamingHelper
+    {
+        static readonly List<Rename> queue = new List<Rename>();
+        static Rename currentCommand;
+
+        public static void AddToQueue(ASResult target)
+        {
+            AddToQueue(target, true);
+        }
+        public static void AddToQueue(ASResult target, bool outputResults)
+        {
+            string name = RefactoringHelper.GetRefactorTargetName(target);
+            string label = TextHelper.GetString("Label.NewName");
+            string title = string.Format(TextHelper.GetString("Title.RenameDialog"), name);
+            LineEntryDialog askName = new LineEntryDialog(title, label, name);
+            if (askName.ShowDialog() == DialogResult.OK && askName.Line.Trim().Length > 0 && askName.Line.Trim() != name)
+            {
+                string newName = askName.Line.Trim();
+                queue.Add(new Rename(target, outputResults, newName));
+                if (ASContext.Context.CurrentModel.haXe && target.Member != null &&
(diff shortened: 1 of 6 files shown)


Meychi:
Great stuff as usual SlavaRa! @elsassph, ok for merge?

External/Plugins/CodeRefactor/Provider/RenamingHelper.cs
```diff
@@ -0,0 +1,78 @@
...
+            LineEntryDialog askName = new LineEntryDialog(title, label, name);
+            if (askName.ShowDialog() == DialogResult.OK && askName.Line.Trim().Length > 0 && askName.Line.Trim() != name)
+            {
+                string newName = askName.Line.Trim();
+                queue.Add(new Rename(target, outputResults, newName));
+                if (ASContext.Context.CurrentModel.haXe && target.Member != null &&
+                    (target.Member.Flags & (FlagType.Getter | FlagType.Setter)) > 0)
+                {
+                    List<MemberModel> list = target.Member.Parameters;
+                    if (list[0].Name == "get") RenameMember(target.InClass, $"get_{name}", $"get_{newName}", outputResults);
```
Gama11: String interpolation requires C# 6 / VS15, I think @elsassph wanted to avoid that for now?

SlavaRa: I should be replaced by a concatenation?

Gama11:
This doesn't seem to be treated as a single undo action - after renaming, it takes three `Ctrl+Z`s for me  to undo the rename (one for the setter, one for the getter, and one for the actual field).

Gama11:
Also, I get a crash in `Rename.cs:172` because `target.Type` is null with this code when renaming `test`:

```haxe
package;

class Test
{
	public var test(get, set):Int;
	
	public var get_test():Int
	{
		return 5;
	}
	
	public var set_test(i:Int):Int
	{
		return i;
	}
}
```

Gama11:
This code produces an exception in `Rename.cs:172` when renaming `test` because `target.Type` is null (not valid Haxe code of course because `get_test` is a `var`, but it still shouldn't crash):

```haxe
package;

class Test
{
	public var test(get, never):Int;
	
	public var get_test:Int;
}
```

SlavaRa:
I think that the problem Undo Refactoring is much more extensive, and it must be considered separately

About crash: Thanks, I corrected

Neverbirth:
IMHO that RenamingHelper class shouldn't be only static, I seriously think it may pose some problem in the future.

SlavaRa:
Yes, may be. I think that we will come to this quickly if we do small iteration

External/Plugins/CodeRefactor/Provider/RenamingHelper.cs
elsassph: Yes let's stick with VS2013 compatibility please (eg. concat or String.Format).

SlavaRa: Okay. Fixed

elsassph:
About the static classes: yes that's not great, but let's try to get it right when comes the occasion to add some other action to a refactoring queue - that should be simple to do it right.

SlavaRa:
I should do something else in this issue(#237)?

I have a few more improvements for refactoring, I'm sure that in the process of their adoption helpers move and rename are merged into one tool, but I wanted to come to this gradually.
Now, I prepare
the movement of classes by hotkey
the movement of class members by hotkey
find all implementations
improve rename

External/Plugins/CodeRefactor/Provider/RenamingHelper.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+        public static void AddToQueue(ASResult target, bool outputResults)
+        {
+            string name = RefactoringHelper.GetRefactorTargetName(target);
+            string label = TextHelper.GetString("Label.NewName");
+            string title = string.Format(TextHelper.GetString("Title.RenameDialog"), name);
+            LineEntryDialog askName = new LineEntryDialog(title, label, name);
+            if (askName.ShowDialog() == DialogResult.OK && askName.Line.Trim().Length > 0 && askName.Line.Trim() != name)
+            {
+                string newName = askName.Line.Trim();
```
Neverbirth: Not very nice to see 3 Trims over the same string.

External/Plugins/CodeRefactor/Provider/RenamingHelper.cs
```diff
@@ -0,0 +1,79 @@
...
+                    if (list[0].Name == "get") RenameMember(target.InClass, "get_" + name, "get_" + newName, outputResults);
+                    if (list[1].Name == "set") RenameMember(target.InClass, "set_" + name, "set_" + newName, outputResults);
+                }
+                if (currentCommand == null) ExecuteFirst();
+            }
+        }
+
+        static void RenameMember(ClassModel inClass, string name, string newName, bool outputResults)
+        {
+            MemberModel m = inClass.Members.Cast<MemberModel>().FirstOrDefault(it => it.Name == name);
```
Neverbirth: Use Members.Items instead of Members.Cast<MemberModel>

SlavaRa: Fixed

External/Plugins/CodeRefactor/Provider/RenamingHelper.cs
SlavaRa: Fixed

wise0704:
I see you using my contextual code generator feature... huehuehue......

Meychi:
Is this now ok @elsassph?

Gama11:
This also fails, but probably due to an unrelated issue like `get_test()` not being parsed as a function correctly due to the missing curly brackets:

```haxe
package;

class Test
{
	public var test(get, never):Int;
	
	public function get_test():Int
		return 5;
}
```

SlavaRa:
Yes, it's a bug ;(
But It has nothing to do with this refactoring

SlavaRa:
Now I look forward to considering this PR to make to make the following

SlavaRa:
@elsassph please review this issue

PR #974, opened 2015-12-04
Updates to MainForm (plus Related Things)

wise0704:
Small changes. Currently 2 changes are made.

**1.**
`ShortcutManager.RegisteredItems` is changed to the type `Dictionary<string, ShortcutItem>` for better performance of searching by ID. In relation to this change, 
```C#
string IMainForm.GetShortcutItemId(Keys)
Dictionary<Keys, string> IMainForm.GetShortcutItemsByKeys()
```
have been added. With those changes, codes like below
```C#
if (keys == PluginBase.MainForm.GetShortcutItemKeys("Namespace.Shortcut1") { }
else if (keys == PluginBase.MainForm.GetShortcutItemKeys("Namespace.Shortcut2") { }
else if (keys == ...
```
which had to loop through the whole shortcut list _n_ number of times previously, can be replaced with either
```C#
switch (PluginBase.MainForm.GetShortcutItemId(keys))
{
    case "Namespace.Shortcut1": { } break;
    case "Namespace.Shortcut2": { } break;
    ...
}
```
or if the shortcut values would not change during a process (such as the `LineEntryDialog` class)
```C#
Dictionary<Keys, string> shortcuts = PluginBase.MainForm.GetShortcutItemsByKeys();
...
string id;
if (shortcuts.TryGetValue(keys, out id))
{
    switch (id) { ... }
}
```

**2.**
Added an option for retrieving a boolean theme value from MainForm.
```C#
bool IMainForm.GetThemeFlag(string)
bool IMainForm.GetThemeFlag(string, bool)
```

### DIFF
--- a/External/Plugins/ProjectManager/Helpers/LineEntryDialog.cs
+++ b/External/Plugins/ProjectManager/Helpers/LineEntryDialog.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Drawing;
 using System.Collections;
+using System.Collections.Generic;
 using System.ComponentModel;
 using System.Windows.Forms;
 using PluginCore;
@@ -13,13 +14,12 @@ namespace ProjectManager.Helpers
     /// </summary>
     public class LineEntryDialog : Form
     {
-        readonly Keys shortcutToLowercase;
-        readonly Keys shortcutToUppercase;
         string line;
+        readonly Dictionary<Keys, string> shortcuts;
 
         #region Form Designer Components
 
-        protected System.Windows.Forms.TextBox lineBox;
+        private System.Windows.Forms.TextBox lineBox;
         private System.Windows.Forms.Button btnOK;
         private System.Windows.Forms.Button btnCancel;
         /// <summary>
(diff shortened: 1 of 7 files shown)


External/Plugins/ProjectManager/Helpers/LineEntryDialog.cs
```diff
@@ -172,18 +171,22 @@ private void btnCancel_Click(object sender, System.EventArgs e)
...
-            lineBox.Paste(selectedText);
-            SelectRange(selectionStart, selectionLength);
+            string shortcutId;
+            if (!shortcuts.TryGetValue(e.KeyData, out shortcutId)) return;
+            switch (shortcutId)
+            {
+                case "EditMenu.ToLowercase":
+                case "EditMenu.ToUppercase":
+                    string text = lineBox.SelectedText;
+                    if (string.IsNullOrEmpty(text)) break;
```
SlavaRa: I think that if no text is selected, it does not make sense to check hotkeys

Meychi:
Thanks but you are not just updating MainForm :)

Please keep these PR's smaller by not doing multiple things in them and also do not do code formatting/style changes.

External/Plugins/ProjectManager/Helpers/LineEntryDialog.cs
wise0704: I made it that way because more shortcut options may be available which might work even when no text is present, an example might be Undo/Redo (I haven't done any other implementations however, since I was meant to only make changes to the structure). If the user customized these shortcuts in the main keyboard settings, they might as well want them to work on these controls.

wise0704:
Sorry for changing the code style, I didn't quite mean them. I might have done some whole-file refactoring with ReSharper by accident.

I did end up changing some more parts of the code, like changing `if` statements to `switch` and repeated `return` statements into single statements. But (correct me if I'm wrong, I sometimes do things that I didn't plan at the start) I don't think I've made any significant changes in irrelevant code.

wise0704:
Just a quick question. 
For the coding style, do you prefer type names instead of keywords (i.e. `String` instead of `string`) in all places?
And is it alright using the keyword `var`?

Neverbirth:
Sure, Mika loves `var` ;P.

wise0704:
I feel that is sarcasm.

Neverbirth:
I'd never do that! heh. It's an ongoing discussion, some people like `var` and he doesn't, the same goes for string vs String. For now I'd say that if in your file String outweights string, use that, the same for var.

wise0704:
Changed the code style back!

I added a method that gets an `Enum` value from the theme values, since from #954 @Meychi mentioned that a dialogue for modifying themes may be created.
I thought current theme options such as `"ImageManager.ImageSet"` which has values `"Bright"`, `"Dim"`, `"Dark"`, `"Darker"` and `"Black"`,  would work better as enum values.

wise0704:
I see that `Enum.Parse` uses cached hash tables to retrieve the values.
I'm not sure how much more memory it would take to do so, but I now think that just having a predefined array of strings (if a theme dialogue would be made) may be a better option. Or even raw string entry with checking for validity on enter, idk.

PR #977, opened 2015-12-04
Implement usage completion, closes #701

### LINKED ISSUE #701: [haxe] implement compiler usage completion
http://haxe.org/manual/cr-completion-usage.html

This can be used for more reliable "Find all references" results (and in cases where FD doesn't find any results at all).

Gama11:
Back in August, I implemented usage completion, but never really did anything with it. I still don't, but might as well do a PR with what I have so far. :) (find all references and refactor->rename should use this, but it wasn't as straightforward to integrate there as I had hoped). @SlavaRa might find this useful to improve refactoring and whatnot.

The xml results for usage and position completion are very similar - the difference is that usage completion can have multiple position results ([as seen in the docs](http://haxe.org/manual/cr-completion-usage.html)).

@elsassph Since it's not actually used anywhere yet - you can test `GetUsages()` with [this branch](https://github.com/Gama11/flashdevelop/commit/d1f103872bc57852eccce01c9ebe19890acb4681), just press F4 and set a breakpoint in `OnPositionResult()`.


SlavaRa:
yes, this is good stuff

Neverbirth:
Really interesting, although I guess that for public members and types we still need to merge the results with our current implementation, because since this is based on the compiler unused types won't be caught, will them?

I can more or less devise an abstraction (IFindReferencesService or similar) that could have several implementations (one for the current one, another one for any possible compiler service offering this feature, and another "composite" one that may use any of the others in a loop).

Gama11:
@Neverbirth Right, I hadn't really considered that. I think we could potentially add `--macro include` to the hxml, [using the `classPaths` argument](http://api.haxe.org/haxe/macro/Compiler.html#include) to include all haxelibs / source paths.

A composite implementation sounds messy.. I'd prefer if we could just always only use the compiler results when available, since that's bound to be much more reliable than FD's find references (which works terribly in my experience).

Neverbirth:
I don't know that argument, but when using it what about projects with junk source code files that have wrong code? Will they break the compiler results?

If done right a composite implementation is not messy, you may have something like: HaxeCompilerFindReferencesService and DefaultCompilerFindReferenceService, and then you can have a CompositeFindReferenceService that just iterate over the services it has inside and when it has all the resultsets it's a matter of filtering, which shouldn't be a problem since all the implementations should return the same type for results (and if they are different for some reasons you could inject some IEqualityComparer or something else similar in concept).

elsassph:
I think it's good enough if Find References/Usages returns you results for the current compiler options - the benefit in terms of speed is considerable.

elsassph:
I'm happy with the implementation BTW. 
Find Usages results could be sent to Results panel (trace to output).

Neverbirth:
I'm well aware the benefits are several heh, but I'm concerned with getting a good implementation rather than getting something that works.

I haven't looked at the implementation anyway.

Gama11:
Do I need to make any changes here?

elsassph:
Nope, I forgot to merge.

PR #978, opened 2015-12-05
Updates to ShortcutManager

wise0704:
Fixed the bug of Secondary Shortcut items not being updated. ~~Oops~~


wise0704:
Was this fixed already?

Meychi:
I reverted the previous MainForm commit as it had issues and you missed functionalities that needed to be changed too. You can PR the important parts again with new smaller PR's.

PR #979, opened 2015-12-05
Delayed Update for ShortcutDialog

wise0704:
The dialog waits for a certain delay when the filter is updated, before re-populating the list.
This prevents the lag caused on slow computers (mine) when typing fast enough.

### DIFF
--- a/FlashDevelop/Dialogs/ShortcutDialog.cs
+++ b/FlashDevelop/Dialogs/ShortcutDialog.cs
@@ -16,6 +16,7 @@ namespace FlashDevelop.Dialogs
 {
     public class ShortcutDialog : SmartForm
     {
+        private Timer updateTimer;
         private ToolStripMenuItem removeShortcut;
         private ToolStripMenuItem revertToDefault;
         private ToolStripMenuItem revertAllToDefault;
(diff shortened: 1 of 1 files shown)


Meychi:
You are doing these ordinal changes here and there. How much of a performance improvements it brings? If not greatly please change only the necessary things in these PR's you do.

wise0704:
Ok. This is some stuff I learnt recently :)
When comparing two strings ignoring cases, there are two ways of doing it. The popular way (i.e. the way most programmers use in many languages):
```C#
string1.ToLower() == string2.ToLower()
```
The C# recommended way:
```C#
string1.Equals(string2, StringComparison.OrdinalIgnoreCase)
```
 * The problem with the first method is this: firstly, a **culture-sensitive** process is done on `string1` (note that culture sensitive processes cost a lot lot more than ordinal), and also a **copy** of `string1` is created. Secondly, the same thing happens to `string2`. These result in two culture-sensitive operations and two copies of strings. Lastly, an **ordinal** equality check is done with the two copies.
 * The second method does one **ordinal** equality check while ignoring case (no culture-specific operations).

The second method is faster and consumes less memory. 

The overall performance increase on the changes I make depend on how frequent the segment is executed, but looking at the proportional performance increase per execution, I think it would be good to change those small bits of code as I see. A bunch of small improvements would sum to make a big improvement, right? :)

PR #981, opened 2015-12-07
Fixes to the bug of image colors not adjusting according to the theme.

wise0704:
Changes are made to `ImageManager` to allow fixes to the bug of image colors not being adjusted according to the theme. 

Previously, image colors are only adjusted when FD starts, therefore changing the theme while FD is running does not cause the images to be refreshed. 

Now, most of the images are refreshed; some images are still not refreshed when the theme changes. This is because those images are copies of the images returned from `MainForm.FindImage()` or `MainForm.GetAutoAdjustedImage()`. Examples are adding those images to `ImageList`, or calling `ScaleHelper.Scale()` on the image. (`ImageList` keeps copies of the images.)

More commits (or a separate PR) will include the fixes of those making copies of the images. 
One way to fix the issue is to not have the images adjusted (by calling `MainForm.FindImage(image, false)` instead of `MainForm.FindImage(image)`, or by not calling `MainForm.GetAutoAdjustedImage()` on the images).
Another way is to introduce an alternative to the usages of the existing `ImageList` class if possible. I would prefer this, as it means that no images are left out on theme changes.


Meychi:
There are other settings too that would require also a theme restart so instead of rewriting new code which could break, i think we will request user to restart FD.

We really do appreciate your PR's but you need really to avoid rewriting big chunks of code as we have then trouble spotting what are you doing and verifying can it contain issues.

wise0704:
The PR seems large at a glance as I renamed the method `MainForm.ImageSetAdjust()`, which occurs a lot in some files.
Pretty much the only things really changed are `ImageManager` class and `ImageKonverter.ImageAdjust()` method.

`ImageKonverter.ImageAdjust()` has been modified to instead of returning a new adjusted copy of the specified image, it **modifies** the specified destination image. The original image is still kept for use. This technique is used so that no caching of all usages of the images are needed to be made. 
`ImageManager` and `MainForm` provides methods that can get both original and automatically modifying images.

I don't know if I would as a user want to restart FD every time I change the theme - viewing the theme change is only available through the program itself. Also, restarting requires all work being saved, and all opened files need to be opened after restart.
I think it would be better to show changes as much as possible when the theme changes.

I don't think the code will break since it only consist of simple logic. It just doesn't apply to all images due to the way Microsoft's `ImageList` behaves.

Meychi:
I did a quick glance review.

Changing plugincore functionality or changing the interfaces may/will break existing 3rd party plugins. This is something we need always to keep in mind when doing changes.

wise0704:
Ah ok. Fixed.

Meychi:
I tested this and it seems to work ok for most of the images. Your note about the scaling, is this a new restriction or did you just document existing behaviour?

Instead of adding a new method, could/should this be implemented internally without any additional changes?

wise0704:
Currently the mechanism is that every call to `ImageManager.GetAutoAdjustedImagePair()` stores a copy of the image passed as `ImagePair.Adjusted` so that calls to `ImageManager.AdjustAllImages()` will adjust the stored images. 
However, for an image returned by a method that calls `GetAutoAdjustedImagePair()` internally (for example, `MainForm.FindImage()` or `MainForm.GetAutoAdjustedImage()`), if a copy of it was made using either `ScaleHelper.Scale()` or `ImageList`, and no references to the auto-adjusting image is present (except for the internal caching), the copy will firstly not be auto-adjusted as it is not the same image as the one returned, and also `ImageManager` will update the cached auto-adjusting image even though the image is not used - doing unnecessary task.
However, upcoming commits I have will resolve this using `WeakReference`.

I added the method `MainForm.GetAutoAdjustedImage()` in place for `MainForm.ImageSetAdjust()` as now the previous functionality of `ImageSetAdjust()` is quite different. I will restore the previous function of `ImageSetAdjust()`, but if I make `ImageSetAdjust()` do what `GetAutoAdjustedImage()` does right now, there will be some problems. 

`MainFormFindImage(string, bool)` was added so that an image not affected by the theme can be retrieved, but right now this method still calls `ImageManager.GetAutoAdjustedImagePair()` internally even if `false` is passed, which I have to tweak a little bit more.

New commits will come in a few days.

wise0704:
Recent commit made `ImagePair` now keep a weak reference of the adjusted image.

Now, methods that create&cache an automatically adjusted image:
```c#
MainForm.FindImage(image)
MainForm.FindImage(image, true)
MainForm.GetAutoAdjustedImage(image)
```
If images returned by these are lost (due to a copy of it is made without keeping the old one, etc.) the image is automatically garbage collected, so now we don't need to worry about cases where unused images are taking up the memory.

wise0704:
Some images are still not updated when the theme changes, because I can't really do anything about them. Most (if not all) images that are not updated are the ones stored in `ImageList`.
Easiest way to deal with them would probably change
`imageList.Images.Add(PluginBase.MainForm.FindImage("???"))`
to
`imageList.Images.Add(PluginBase.MainForm.FindImage("???", false))`
and change
`imageList.Images.Add(PluginBase.MainForm.ImageSetAdjust(image))`
to
`imageList.Images.Add(image)`
so that the original image is used instead of one-time adjusted images.

Another way I can think of is to replace the usages of `ImageList` with a custom image list object that does not store copies of the image passed.

What are your opinions, guys?

PR #986, opened 2015-12-09
[Flash] Add support of new API Vector.<T> 

SlavaRa:
http://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/Vector.html#insertAt()
http://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/Vector.html#removeAt()


wise0704:
Probs under a new folder `FlashDevelop/Bin/Debug/Library/AS3/intrinsic/FP19`?

SlavaRa:
Nope. Will not work.
But my decision we have discussed in slack

wise0704:
Ah ok. I was just curious.
But do you know the reason why putting it in a new folder does not work?

elsassph:
Yeah FP19 could possibly work, but I figured it was probably not worth the complication.

PR #988, opened 2015-12-10
Revert "Merge pull request #639 from Gama11/TrackActiveDocument"

Neverbirth:
This reverts the PR as discussed [here](https://github.com/fdorg/flashdevelop/pull/639#issuecomment-163182738).

Note, I had to update the basque translation, because before the reverted PR it was using English text.

Opinions are always welcome, the people I've talked to either agree or don't mind this change.

### DIFF
--- a/External/Plugins/ProjectManager/Controls/TreeBar.cs
+++ b/External/Plugins/ProjectManager/Controls/TreeBar.cs
@@ -1,9 +1,9 @@
-using System.Windows.Forms;
-using PluginCore.Localization;
 using System.Drawing;
-using PluginCore.Helpers;
+using System.Windows.Forms;
 using PluginCore;
 using PluginCore.Controls;
+using PluginCore.Helpers;
+using PluginCore.Localization;
 using ProjectManager.Controls.TreeView;
 
 namespace ProjectManager.Controls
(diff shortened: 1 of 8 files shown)


elsassph:
@Meychi what do you think? 
I believe we want to keep this feature optionally: maybe we could add a setting allowing to choose the behaviour of this button (single or sticky synchronization).

Meychi:
Maybe a setting would be good. This way we have both.

Neverbirth:
Mmm tracking the active document is still a setting, so the feature is not lost, while this button now provides a quick and accessible way to locate the active document, which was there before but lost with the reverted PR.

Meychi:
Very true. I'm fine reverting this.

Neverbirth:
In the current development branch, for example in a big project, if you for example are in a file, and select some other node in the file, or collapse folders, etc, you may have to spend some time locating the file node back, or you have to untoggle and toggle back the button, which isn't very nice nor intuitive. I like tracking (I use it in all my IDEs, and I added the tracking setting in FD, so... :P), but this is a different use case than tracking and is a common need in big projects.

With this change the user still have both the tracking setting, and the locate button to quickly select the file node.

elsassph:
Sounds good then.

PR #989, opened 2015-12-11
Contextual Code Generation update

wise0704:
Display a message on the status bar when no contextual generation options are available.

### DIFF
--- a/External/Plugins/ASCompletion/Completion/ASGenerator.cs
+++ b/External/Plugins/ASCompletion/Completion/ASGenerator.cs
@@ -6,7 +6,6 @@ using System.Globalization;
 using System.IO;
 using System.Text;
 using System.Text.RegularExpressions;
-using System.Windows.Forms;
 using ASCompletion.Context;
 using ASCompletion.Model;
 using ASCompletion.Settings;
@@ -17,7 +16,6 @@ using PluginCore.Localization;
 using PluginCore.Managers;
 using PluginCore.Utilities;
 using ScintillaNet;
-using WeifenLuo.WinFormsUI.Docking;
 
 namespace ASCompletion.Completion
 {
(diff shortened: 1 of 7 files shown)


External/Plugins/ASCompletion/PluginMain.cs
```diff
@@ -279,12 +279,12 @@ public void HandleEvent(Object sender, NotifyEvent e, HandlingPriority priority)
...
-                                    setup.Version = (string)info["version"];
-                                    setup.TargetBuild = (string)info["targetBuild"];
-                                    setup.Classpath = (string[])info["classpath"];
-                                    setup.HiddenPaths = (string[])info["hidden"];
+                                    setup.Platform = (string) info["platform"];
+                                    setup.Lang = (string) info["lang"];
+                                    setup.Version = (string) info["version"];
+                                    setup.TargetBuild = (string) info["targetBuild"];
+                                    setup.Classpath = (string[]) info["classpath"];
+                                    setup.HiddenPaths = (string[]) info["hidden"];
```
wise0704: This happened because I closed a brace at a outer level...
Not intentional.

External/Plugins/ASCompletion/Completion/ASGenerator.cs
```diff
@@ -6,18 +6,16 @@
 using System.IO;
 using System.Text;
 using System.Text.RegularExpressions;
-using System.Windows.Forms;
 using ASCompletion.Context;
 using ASCompletion.Model;
 using ASCompletion.Settings;
 using PluginCore;
-using PluginCore.Controls;
+using PluginCore.Controls;
```
wise0704: git plz

wise0704:
In case people get confused of what I did (because I'm quite confused myself..)

The `ASGenerator` has been refactored to accept a parameter instead of having a static variable which eats up memory.
Changed the huge `if else` statement to `switch` statement so that the performance increases from O(n) to O(1).
The status bar now reports "No contextual code generation available." when the context generation list is empty.

PR #990, opened 2015-12-11
IMainForm Documentation

wise0704:
When projects other than `FlashDevelop` accesses the main form, they get the type of the interface `IMainForm` instead of `MainForm`. So they don't get to see any documentation on the properties and methods.
Copy & pasterino-ed all documentation from `MainForm` to `IMainForm`.

### DIFF
--- a/PluginCore/PluginCore/Interfaces.cs
+++ b/PluginCore/PluginCore/Interfaces.cs
@@ -98,79 +98,287 @@ namespace PluginCore
     {
         #region IMainForm Methods
 
+        /// <summary>
+        /// Refreshes the main form.
+        /// </summary>
         void RefreshUI();
+        /// <summary>
+        /// Stop the currently running process.
+        /// </summary>
         void KillProcess();
+        /// <summary>
+        /// Refreshes the scintilla configuration.
+        /// </summary>
         void RefreshSciConfig();
+        /// <summary>
+        /// Themes the controls from the parent.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


PluginCore/PluginCore/Interfaces.cs
```diff
@@ -98,79 +98,287 @@ public interface IMainForm : IContainerControl, IWin32Window
...
+        /// <summary>
+        /// Gets the status strip.
+        /// </summary>
         StatusStrip StatusStrip { get; }
+        /// <summary>
+        /// Gets or sets the working directory.
+        /// </summary>
         String WorkingDirectory { get; set; }
+        /// <summary>
+        /// Gets the tool strip panel.
```
Gama11: Yeah, `Gets the <property name>` style docs are really useful. ;P

wise0704: Haha, I didn't like it either, but that is what is written in the corresponding method's documentation in `MainForm`.
All I did was copy & paste in most of the cases.

elsassph: Yeah Mika was a bit uninspired when writing these comments :D

wise0704: hehehehe i could feel that
originally I was going to add very descriptive documentation, including <param> tags and stuff. After like 5 methods, I was like. "You know what? NO. No just no."

PR #992, opened 2015-12-11
AS3 Intrinsic Library Improvement

wise0704:
Related to the PR #986.

Intrinsic library paths are now added from the highest version number downwards.
PR #986 modified `Vector.as` in the folder `FP11` but it still does not work since there is a same intrinsic class defined in `FP10` folder. It seems that first registered is being used, so the solution is to register from the highest version number of FP.

Added `FP19\Vector.as` as well.

### DIFF
--- a/External/Plugins/AS3Context/Context.cs
+++ b/External/Plugins/AS3Context/Context.cs
@@ -334,20 +334,19 @@ namespace AS3Context
             }
 
             // intrinsics (deprecated, excepted for FP10 Vector.<T>)
-            string fp9cp = as3settings.AS3ClassPath + S + "FP9";
-            AddPath(PathHelper.ResolvePath(fp9cp));
-            if (majorVersion > 9)
+            // add from the highest version number (FP11 > FP10 > FP9)
+            if (majorVersion >= 9)
             {
-                for (int i = 10; i <= majorVersion; i++)
+                string fp = as3settings.AS3ClassPath + S + "FP";
+                for (int i = majorVersion; i >= 10; i--)
                 {
-                    string fp10cp = as3settings.AS3ClassPath + S + "FP" + i;
-                    AddPath(PathHelper.ResolvePath(fp10cp));
-                    for (int j = 1; j <= minorVersion; j++)
+                    for (int j = minorVersion; j >= 0; j--)
                     {
-                        string fp101cp = as3settings.AS3ClassPath + S + "FP" + majorVersion + "." + minorVersion;
-                        AddPath(PathHelper.ResolvePath(fp101cp));
+                        AddPath(PathHelper.ResolvePath(fp + i + "." + j));
                     }
+                    AddPath(PathHelper.ResolvePath(fp + i));
                 }
+                AddPath(PathHelper.ResolvePath(fp + "9"));
             }
 
             // add external pathes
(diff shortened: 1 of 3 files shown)


External/Plugins/AS3Context/Context.cs
```diff
@@ -334,20 +334,19 @@ public override void BuildClassPath()
...
+                string fp = as3settings.AS3ClassPath + S + "FP";
+                for (int i = majorVersion; i >= 10; i--)
                 {
-                    string fp10cp = as3settings.AS3ClassPath + S + "FP" + i;
-                    AddPath(PathHelper.ResolvePath(fp10cp));
-                    for (int j = 1; j <= minorVersion; j++)
+                    for (int j = minorVersion; j >= 0; j--)
                     {
-                        string fp101cp = as3settings.AS3ClassPath + S + "FP" + majorVersion + "." + minorVersion;
-                        AddPath(PathHelper.ResolvePath(fp101cp));
```
wise0704: @elsassph, are you serious??? :P
That would add the folder `FP10.9` 9 times if `majorVersion` is 10 and `minorVersion` is 9. Hehe

elsassph: Haha OMG that's silly :)
In my defence this was probably written when FP 11 just came out...

And it still isn't correct: when we'll have FP 20.0 it will never find the intrinsics for FP11.1!

wise0704: Ah, right.
Somehow that line of code make people suddenly noobs... lol

wise0704: Probabily loop from 9 instead of minorVersion?

elsassph: Yes this sounds safe enough.
But actually this all seems rather wasteful as it brute-force checks for a large number of non-existent paths - it was ok when there was just a pair of versions.

Neverbirth: I think checking if a directory exists isn't a very time consuming method. How other IDEs manage this case? Is it really not possible to get the Vector methods from the swc? Or is the problem because we remove that __AS3Vec__ namespace or whatever?

wise0704: Ok for the simplest solution how about just ignoring `minorVersion`?
Like, just check for FP9, FP10 and so on, since Adobe isn't going to introduce any minor versions from now on.
Also, we already know that we're not going to add any intrinsic classes for like FP11.4 or like FP10.3 or FP3.14159265 whatever

elsassph: True, minor versions don't matter that much now.

External/Plugins/AS3Context/Context.cs
```diff
@@ -361,20 +352,20 @@ public override void BuildClassPath()
...
-            // add initial pathes
+            // add initial paths
             foreach (PathModel mpath in initCP) AddPath(mpath);
 
             // parse top-level elements
             InitTopLevelElements();
             if (cFile != null) UpdateTopLevelElements();
 
-            // add current temporaty path
+            // add current temporary path
```
wise0704: What are thoooooseeeee

wise0704:
So is this alright now?

PR #993, opened 2015-12-11
Customizable Output Markers

### LINKED ISSUE #714: Custom UI Scheme: the color of command trace() in the Output Panel
I created custom UI Scheme for FlashDevelop.
How can I change the color of command trace() in the Output Panel?
(The dark UI Scheme background is very good for the eyes)

-Philippe
>>It isn't possible yet but I think Mika's working on it. Please create a github issue to track it.

Forum topic: www.flashdevelop.org/community/viewtopic.php?f=9&t=12226

Thanks!

Screenshot:
trace

### LINKED ISSUE #932: Output panel - customizable number of log levels (or just add a few more)
I have a logger class which, similar to syslog, has these log levels:
```
debug = debug
info = info
notice = something unusual but not an error
warn = warning
err = error
crit = critical error
alert = fatal (app unusable)
```
The current system doesn't have enough log levels to support different colors for each. Additionally, I would prefer if app's info log entries be colored different from info output by FD (e.g. compile process) - info is a higher severity log level than debug, therefore should be more noticeable color than debug, bu compile process is more of a debug info, maybe even lower, so I'd use separate colors for app info and FD info messages.

The user can customize output panel's log level markers in the output panel settings, but the number and names of log levels are predefined. I think default log levels are great for many (most) users, but I think it wouldn't be a big modification to allow custom log levels. So the "Level" combo box in the output panel's highlight markers settings should allow entering custom text, for example the word "Alert", and output panel can just fetch "OutputPanel.AlertColor" from the theme (which a user would add to his theme). I know this is a low priority feature request, so if you don't want to allow custom log levels, please just add a few more predefined levels.

wise0704:
Added support for customizable output highlight markers.
Closes #932 and closes #714.

### DIFF
--- a/External/Plugins/OutputPanel/Settings.cs
+++ b/External/Plugins/OutputPanel/Settings.cs
@@ -68,14 +68,16 @@ namespace OutputPanel
         {
             get
             {
-                if (highlightMarkers == null || highlightMarkers.Count == 0)
+                if (highlightMarkers == null)
                 {
-                    this.highlightMarkers = new List<HighlightMarker>();
-                    this.highlightMarkers.Add(new HighlightMarker("Info:", LogLevel.Info));
-                    this.highlightMarkers.Add(new HighlightMarker("Debug:", LogLevel.Debug));
-                    this.highlightMarkers.Add(new HighlightMarker("Warning:", LogLevel.Warning));
-                    this.highlightMarkers.Add(new HighlightMarker("Error:", LogLevel.Error));
-                    this.highlightMarkers.Add(new HighlightMarker("Fatal:", LogLevel.Fatal));
+                    this.highlightMarkers = new List<HighlightMarker>
+                    {
+                        new HighlightMarker("Info:", LogLevel.Info),
+                        new HighlightMarker("Debug:", LogLevel.Debug),
+                        new HighlightMarker("Warning:", LogLevel.Warning),
+                        new HighlightMarker("Error:", LogLevel.Error),
+                        new HighlightMarker("Fatal:", LogLevel.Fatal)
+                    };
                 }
                 return highlightMarkers;
             }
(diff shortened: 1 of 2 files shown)


External/Plugins/OutputPanel/Settings.cs
```diff
@@ -86,28 +87,65 @@ public List<HighlightMarker> HighlightMarkers
...
-            set { this.level = value; }
+            set
+            {
+                this.level = value;
+                if (value != LogLevel.Custom) this.highlightColor = Color.Empty;
+            }
+        }
+
+        [Category("Properties")]
+        [Description("User defined color for custom highlight markers.")]
```
wise0704: Should be a localised description.

wise0704:
Possibly closes #714 as well.

Meychi:
This change is against the current direction where we try to put all theme settings into themes and not the main UI. #714 is already done and should be closed.

wise0704:
Ah, right. To be honest, the main goal of this PR was to fix #932. I just thought that I could also fix #714 at the same time. Turns that it was already solved :D

#932 is asking for a way to allow users to add more log levels. I think custom log levels are not part of the theme, but the colors may be.

If the current way of defining more highlight markers with custom colors seems against the direction, how about maybe instead allow users to enter custom text (like `CustomLevel` property) when `Level == LogLevel.Custom`, so that it retrieves `("OutputPanel.{0}Color", CustomLevel)` from the theme? (Or maybe even have `Level` as a `string` value in the first place.)
However one problem that can arise would be the user having to manually add those custom color definitions for every theme he/she has.

Meychi:
@elsassph: What do you think about this? I would not add it but i'm not against either if you think its good.

elsassph:
I would like to see `0:bah` to be just deprecated so it sounds a serious added complication in the opposite direction.

Meychi:
Ok, closing.

PR #994, opened 2015-12-11
[Refactoring] Replace "String" by ASContext.Context.Features.stringKey

SlavaRa:


Neverbirth:
I'm curious, can you name benefits from this change?

SlavaRa:
It removes the "magic" values and makes the code more uniform

PR #995, opened 2015-12-12
Rename project from Project Manger

SlavaRa:
Result: 
rename_project_from_project_manager_ui

### DIFF
--- a/External/Plugins/ProjectManager/PluginUI.cs
+++ b/External/Plugins/ProjectManager/PluginUI.cs
@@ -239,7 +239,14 @@ namespace ProjectManager
                     if (Rename(oldName, newName))
                     {
                         PluginBase.MainForm.OpenEditableDocument(newName);
-                        File.Delete(oldName);
+                        try
+                        {
+                            File.Delete(oldName);
+                        }
+                        catch (Exception)
+                        {
+                            // ignored
+                        }
                     }
                     else e.CancelEdit = true;
                 }
--- a/External/Plugins/ProjectManager/PluginUI.cs
+++ b/External/Plugins/ProjectManager/PluginUI.cs
@@ -230,17 +230,16 @@ namespace ProjectManager
                         newName = Path.Combine(project.Directory, label);
                         newName = Path.ChangeExtension(newName, Path.GetExtension(oldName));
                     }
-                    catch (Exception exception)
+                    catch (Exception)
                     {
                         e.CancelEdit = true;
                         isEditingLabel = false;
-                        ErrorManager.ShowError(exception);
                         return;
                     }
                     if (Rename(oldName, newName))
                     {
                         PluginBase.MainForm.OpenEditableDocument(newName);
-                        File.Delete(oldName); //slavara: because the old file remains after renaming
+                        File.Delete(oldName);
                     }
                     else e.CancelEdit = true;
                 }
--- a/External/Plugins/ProjectManager/PluginUI.cs
+++ b/External/Plugins/ProjectManager/PluginUI.cs
@@ -224,8 +224,19 @@ namespace ProjectManager
                     string label = e.Label;
                     int index = label.IndexOf(languageDisplayName);
                     if (index != -1) label = label.Remove(index).Trim();
-                    var newName = Path.Combine(project.Directory, label);
-                    newName = Path.ChangeExtension(newName, Path.GetExtension(oldName));
+                    string newName = string.Empty;
+                    try
+                    {
+                        newName = Path.Combine(project.Directory, label);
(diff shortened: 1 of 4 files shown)


elsassph:
I think you should remove the target name (Haxe) when edit begins.

Meychi:
Or what about using the rename dialog avoiding all issues?

elsassph:
Good point - it's fancier to rename in the tree but it would be fair to prompt for a new name.

SlavaRa:
After last commit:
rename_project_from_project_manager_ui_24122015

External/Plugins/ProjectManager/PluginUI.cs
```diff
@@ -218,12 +215,30 @@ private void tree_BeforeLabelEdit(Object sender, NodeLabelEditEventArgs e)
...
             if (!string.IsNullOrEmpty(e.Label) && Rename != null)
             {
-                if (!Rename((e.Node as GenericNode).BackingPath, e.Label))
+                if (e.Node is ProjectNode)
+                {
+                    var oldName = project.ProjectPath;
+                    string label = e.Label;
+                    int index = label.IndexOf(languageDisplayName);
+                    if (index != -1) label = label.Remove(index).Trim();
+                    var newName = Path.Combine(project.Directory, label);
```
elsassph: Needs a try/catch to make sure the path is valid.

External/Plugins/ProjectManager/PluginUI.cs
```diff
@@ -218,12 +215,41 @@ private void tree_BeforeLabelEdit(Object sender, NodeLabelEditEventArgs e)
...
+                    {
+                        e.CancelEdit = true;
+                        isEditingLabel = false;
+                        ErrorManager.ShowError(exception);
+                        return;
+                    }
+                    if (Rename(oldName, newName))
+                    {
+                        PluginBase.MainForm.OpenEditableDocument(newName);
+                        File.Delete(oldName); //slavara: because the old file remains after renaming
```
elsassph: Doesn't need `slavara` in the comment, and a silent `try/catch` would be safer.

External/Plugins/ProjectManager/PluginUI.cs
```diff
@@ -218,12 +215,41 @@ private void tree_BeforeLabelEdit(Object sender, NodeLabelEditEventArgs e)
         /// </summary>
         private void tree_AfterLabelEdit(Object sender, NodeLabelEditEventArgs e)
         {
+            string languageDisplayName = "(" + project.LanguageDisplayName + ")";
             if (!string.IsNullOrEmpty(e.Label) && Rename != null)
             {
-                if (!Rename((e.Node as GenericNode).BackingPath, e.Label))
+                if (e.Node is ProjectNode)
+                {
+                    var oldName = project.ProjectPath;
```
elsassph: Can this whole project renaming code be moved into a function?

External/Plugins/ProjectManager/PluginUI.cs
SlavaRa: I send in the near future

External/Plugins/ProjectManager/PluginUI.cs
SlavaRa: I send in the near future

External/Plugins/ProjectManager/PluginUI.cs
SlavaRa: @elsassph about "and a silent try/catch would be safer", is it about:
```
catch (Exception exception)
{
  e.CancelEdit = true;
  isEditingLabel = false;
  ErrorManager.ShowError(exception);
  return;
}
```
?

External/Plugins/ProjectManager/PluginUI.cs
```diff
@@ -218,12 +215,40 @@ private void tree_BeforeLabelEdit(Object sender, NodeLabelEditEventArgs e)
...
+                    catch (Exception)
+                    {
+                        e.CancelEdit = true;
+                        isEditingLabel = false;
+                        return;
+                    }
+                    if (Rename(oldName, newName))
+                    {
+                        PluginBase.MainForm.OpenEditableDocument(newName);
+                        File.Delete(oldName);
```
elsassph: Why do we have to delete the old file in fact?
Also this should be guarded by a `try/catch`.

SlavaRa: > Why do we have to delete the old file in fact?
Otherwise it turns out the two files, old and new

> Also this should be guarded by a try/catch.
today

Meychi:
@elsassph: Should we merge this as such and not with a dialog? Ok for me.

PR #997, opened 2015-12-12
Modifier order

### LINKED ISSUE #885: ASCompletion: access modifiers before static, but after override (like flex sdk)
ASCompletion already has a useful option ``start declaration with access modifiers``. There should also be an option to control where the ``override`` keyword goes.
Basically I think there should be a new option to replace the ``start declaration with access modifiers``.
My proposal is a multi-value option (dropdown) with different combinations of access-modifiers, override and static.
Personally I need it because I want to start declaration with override, then access modifiers, then static.
This is the same style as used in Flex SDK, and it makes sense to me.
```
override public function x():void { ... }
public static function s():void { ... }
override public static function s():void { ... }
```

wise0704:
More configurable options for ordering of modifiers.
image

Closes #885.


wise0704:
This PR is under reconstruction, as suggested by @SlavaRa.

Meychi:
Also by changing the type of the setting, this will break when the user has this setting modified and FD tries to deserialize it.

wise0704:
I know. But FD nicely handles it by just showing an error dialog on first run, then the setting is set to the new default value (well, technically not set). 

Unless the user freaks out after seeing an error dialog, it shouldn't be much of a problem though. hehe

Neverbirth:
Mmm... You may make the changes backward compatible by using OnDeserializingAttribute or OnDeserializedAttribute.

wise0704:
Could do. But again, unless a user freaks out and faints and gets carried to the nearest hospital after seeing the error message, I would prefer a one-time error dialog than more unnecessary code written. 
The settings data will be overridden on first run, but the code will still be there.

Meychi:
Give the setting a new name and it works fine then. It should not give an error. @elsassph?

wise0704:
Ta da
image

The `<Others>` may be custom modifiers like AS3 `namespace`.

Gama11:
Is there a separate setting for Haxe? A lot of these it doesn't have, and additionally there's `macro` and `dynamic`.

wise0704:
It doesn't need separate settings, as it only orders what is currently in the code.
For example, if the setting is `override, public, internal, protected, private, <Others>, static`, and the code being inserted is `static public function...`, it is sorted to `public static function...`.

I'll add `macro` and `dynamic`. Are there any more modifier keyword in any other languages?

Neverbirth:
Mmm maybe you missed part of Gama11 comment, if so just in case I'll repeat: It's not only about adding those ones, it's also that for example protected is not on Haxe. Since ASCompletion is used for several languages I don't know if you could provide in a nice way different alternatives for each of them...

wise0704:
If `protected` is not available in Haxe, it means that the code snippets won't have `protected` in them so the ASComplete will simply not (cannot, in fact) sort it.
Check out the [`RemoveAndExtractModifier()`](https://github.com/fdorg/flashdevelop/pull/997/files#diff-7a22daae79e80d8fd96ede6e9e1801b5R4351) method. If `protected` is in the sorted list in setting but does it appear in code, nothing happens.

If I explain the mechanism in simple words:
 1. The code checks if the `DeclarationModifierOrder` only has 1 item. Since the item `<Other>` cannot be removed, 1 item means no sorting. Don't call the sorting method.
 2. Otherwise, for each line of code being inserted:
 3. Loop through the `modifierOrder` array. If the matching modifier is found in the modifiers in the line (modifiers are extracted with regex, which is not my code) remove it from the string and add it to buffer.
 4. Concatenate all extracted modifiers. This will effectively sort only the modifiers specified in `modifierOrder`, leaving any unmatching modifiers alone as `<Other>`.
 5. If there was any change, apply the changes.

If `DeclarationModifierOrder` contains `inline`, for example
`public internal protected private`**`inline`**` <Other> override static`
and the inserted AS3 code (which does not have `inline` keyword) is
```ActionScript
override public function toString():String {
    
}
```
it will turn the first line into
```ActionScript
public override function toString():String {...
```

Am I still missing something? Let me know if understood something wrongly.

wise0704:
In case the picture looks kind of misleading:
It **does not add** those modifiers to the snippets being inserted.
It **does sort** what it detects from the sort list in the code being inserted.

Neverbirth:
I understand what you mean, I'm talking from the configuration side of things, which was also what Gama tried to convey with that "separate setting for Haxe", because some Haxe developers will find it strange to see all those non-existent keywords (ok... they will also find strange that you have to modify a setting in a place called ASCompletion, but that's a different subject).

Neverbirth:
Or maybe these settings could be added to the *Context plugins and then inject the values to the ASCompletion plugin? It leaves the question what to do if a *Context plugin that uses ASCompletion doesn't set the value... maybe ASCompletion should also provide the setting like it currently does...

wise0704:
Yeah I get it. But then as you mentioned it is ASCompletion trying to be greedy and manage all languages by itself, a separate settings would not quite follow the pattern. It would be simpler to just merge them all if the plugin itself is designed that way.

Meychi:
I could to check if we could construct setting groups by context or category.

wise0704:
Having this setting across all *Context could work. But I think *Context imports the ASCompletion, not the other way, so the implementation might be quite difficult (idk maybe there is something like CurrentContext?)

Neverbirth:
That's why I said  that the *Context plugins would inject their settings to ASCompletion. ASContext.Context would be the current context, so ASCompletion could also get the settings from the Context, although that way you have to do make sure that the base context class have that property..

wise0704:
However IMO I think current version seems alright....?
It doesn't really look confusing though... or is it?
A good enough description for the setting would do.

Meychi:
The dialog should be done like the other ones we have. No embedded resources, use existing icons and make sure its scales correctly on different DPIs. Otherwise this seem good to me you folks can agree on it. :)

Gama11:
There's also `extern` in Haxe.

wise0704:
New look:
image

wise0704:
Actually, can I use `ResourceHelper.LoadBitmap()` and use the previous images instead of the green arrows? The green arrows look quite ugly...

wise0704:
This can wait for #1013 so that it can use `FindImage16()` for the button images.

Meychi:
@elsassph: Good for merge when finished?

Meychi:
Is this ok now for all parties?

elsassph:
Ok

wise0704:
Does this need any more fixes?

elsassph:
Hum I'm giving it a try and I don't understand how it works (besides the fact that UI isn't high-dpi scalable).
Why are there even 2 lists? 

As far as I can tell, there should be just one list with a list of tokens (like `dynamic`, `extern`, `inline`, `static`,...). Whatever is in this list will be added first and ordered as in the list; whatever isn't in the list is added after in the order they are provided by the user.

wise0704:
There are two lists to provide flexibility. 
It's like a normal collection editor, that you have one list of values (which is the list on the right) that you can add/remove/order items.
Except that you can only add a fixed list of items, and each item only once (unlike a normal collection with enum types, which allows addition of the same item multiple times). Also there can be items that cannot be removed ("<rest>" in this case).

Adding/removing items are supported to treat certain items the same as `<rest>`. When sorting modifiers anything that is in the list is sorted as the list specifies, and everything else will be left in the same order as they were, and inserted where `<rest>` sits in the list.

There isn't any default editor for just ordering items in a collection, so that means it needs a new UI anyway, so I just made a useful one that could be used on other settings too.

wise0704:
Setting `AutoScaleMode` to `Font` and setting `Font` to `PluginBase.Settings.DefaultFont` does all the things needed for HDPI support, right?
It's really difficult to test it... Requires restart every time I change DPI.

Meychi:
@elsassph?

elsassph:
Merged - looking fine on high-dpi. Should be a candidate for unit tests coverage but it's really been sitting here for a long time :)

PR #998, opened 2015-12-12
More Results Panel Shortcuts

### LINKED ISSUE #940: Add keyboard shortcut for clear results (items in Results panel)
Find&Replace in files highlights all the matches in the editor. Once I don't need these matches any more, I have to open the results panel and clear the items (have to use the mouse), or search for some nonsense text (pollutes the search history with nonsense).
Please add a customizable keyboard shortcut that clears all the items from the Results panel.

wise0704:
Added more options for results panel shortcuts.
Fixed a bug of results panel context menu not updating shortcut key texts.
Closes #940.

### DIFF
--- a/External/Plugins/ResultsPanel/PluginMain.cs
+++ b/External/Plugins/ResultsPanel/PluginMain.cs
@@ -236,10 +236,10 @@ namespace ResultsPanel
             String title = TextHelper.GetString("Label.ViewMenuItem");
             ToolStripMenuItem viewMenu = (ToolStripMenuItem)PluginBase.MainForm.FindMenuItem("ViewMenu");
             ToolStripMenuItem viewItem = new ToolStripMenuItem(title, this.pluginImage, new EventHandler(this.OpenPanel));
-            PluginBase.MainForm.RegisterShortcutItem("ResultsPanel.ShowNextResult", this.pluginUI.nextEntry);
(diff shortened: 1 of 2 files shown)


Meychi:
I didn't test this but by the looks of it,  i think you made it so that Control+C doesn't work in the editor anymore. This is why the shortcut display string was used and not handled top level.

Correct me if i'm wrong.

wise0704:
The editor still handles Ctrl+C it seems. Maybe the editor has higher priority for handling key events?

However, you are still right. I should be using `ShortcutKeyDisplayString` instead of setting the shortcut themselves. And also handle `EventType.Shortcut` to keep track of the shortcut changes.

wise0704:
asdfkjsadfy8u I'm stupid.
I made `MainForm.GetShorcutItemId()` for cases like this and I'm not using that what...
Things got much easier.

Meychi:
Why not rename copyEntryContextMenuItem, did it do something bad? :)

wise0704:
Hehe, I think I was renaming those fields when I was doing something like `ke.Value == this.pluginUI.clearIgnoredEntriesContextMenuItem.ShortcutKeys` and decided that the names are too long to be readable...
I guess I don't have to do that now, so I'll just return them all back.

PR #1001, opened 2015-12-13
Added HaxeDevelop startpage / icon / about

markknol:
Hi! I hope this is the correct location to place the startpage files.

### The startpage will look like this:
image

I reordered the page and put the HaxeDevelop logo on it. 

Some notes:
* The news section is still the same since I have no idea how we handle this.
* The download are linking to FlashDevelop, also don't know how we handle this.
* I've kept the design easy with native UI buttons.  

### The aboutpage will look like this:
image

### This icons look like this:
image


markknol:
cc @ncannasse

Meychi:
Are all startpage files modified? The idea in this Distros setup is that only files differing from originals should be places here.

ncannasse:
Looks good. Tell me if I can help with something.

markknol:
@Meychi Yes all files are updated. The html, javascripts, css and images.

Meychi:
Done manually.

elsassph:
@markknol I think you didn't understand the iOS logo template. 

- The dotted rounded rectangle is the icon's background, which is indeed rounded on iOS. It shouldn't really be part of the icon itself, but you can use it on the website's hero visual.
- The dotted circle is given as an indication of the size that the icon's main visual should occupy, it isn't expected to be used as areal visual element.
- There should not be any text inside the icon (eg. icons over 64x64).

Happy to show you on a PSD/AI what I mean if you shares the sources (do you have my email by BTW? I couldn't figure yours).

markknol:
@elsassph 

* I agree the logo is not correct in the startpage. I dislike that circle too, but at some places I like the rounded rectangle.
* I will share the psd soon.
* I already updated the application icon, the new one looks like this:
image

PR #1002, opened 2015-12-13
HaxeDevelop README.md

markknol:
To have some kind of landingspage on Github for the HaxeDevelop distro.

cc @ncannasse


Meychi:
The correct place for this would be in the Distros/HaxeDevelop dir as i am thinking that for easier maintainability this will be merged to main branches. Then we can also do the builds with the same CI instance on one go.

Also the README should contain only HaxeDevelop specific info not repeating FD info. I will handle this when i have everything ready.

Issue #379, opened 2014-09-24
[haxe] Parsing problems with nested function types


Gama11:
FD doesn't like types like this one much:

```haxe
var f:(Void->Void)->Void;
```

The resulting `MemberModel` has the type `Void->Void -> Void`. Because of that, it's also displayed incorrectly in the outline panel and in completion.

Neverbirth:
I'd say this can be closed now.

Gama11:
You're right, that case seems to work now, but if type parameters are involved it breaks:

![](http://i.imgur.com/zGjXH5e.png)

Gama11:
Haven't managed to construct an example where it fails anymore with the improvements.. (so far ;))

Issue #617, opened 2015-03-07
a bug of autocomplete


matrix3d:
![rtx](https://cloud.githubusercontent.com/assets/4150631/6540547/5d389902-c4dd-11e4-8aa1-8c3bc1e8e521.jpg)

get error

Meychi:
What's the error? OS and FD version?

Neverbirth:
I guess that the completion list is not appearing? I've noticed this bug as well when there is a Vector used around, I guess the file parsing doesn't like something about that. If you comment the info.arguments line it will work again, and after adding some more code uncommenting the line won't stop the completion list to appear.

I've seen this on FD5, but surely 4.7 will behave the same (maybe 4.6 as well).

matrix3d:
@Neverbirth  yes.it is can fixed?

Neverbirth:
Can it be fixed? of course, will it be fixed? it depends...

The problem is in ASFileParser as I said, the problem is that info is identified as the right type first, but then when the second line is parsed, it gets confused because of that Vector.<String>(args.split(" ")) and creates a new info model with Vector.<String>(args.split(" ")) as its type.

Neverbirth:
I'd say this can be closed now.

PR #1012, opened 2015-12-18
Unit tests and other improvements

Neverbirth:
I wanted to do this PR with more work done, but I'll do it already in order to get all people to help with which direction to follow, how to do some future work, etc. I don't intend to see this PR merged as-is, but I wouldn't mind if that happens heh.

This PR includes many unit tests, but also several other improvements, why? because with the unit tests it got easier and safer! so it also serves as a showcase of what they may allow in the future.

Some of the improvements: Way better code generation in many cases (for most jobs, because the GetBodyStart failed under many cases...), fixed field from parameter and implement interface under several conditions (they could completely break code or not work at all!), launch proper Haxe auto completion under several cases where it was failing before, better Haxe parsing (fixes issue #379, this one maybe could be fixed in another way but couldn't check it yet, and #617, which also affects Haxe, among other improvements).

My main concern right now is that when running the tests the referenced FD assemblies are placed in the same directory as the FD executable. We should think of some way of being able to run the tests without doing this. Another thing still undecided is how to organize the tests: a test project for each FD project, or a single test project for all of them? Each one has its advantages and disadvantages. I'm OK with any.

Feel free to ask any question or make any comment.

EDIT: I haven't seen anything broken after all my changes, but if I broke something, I think it's good, because then we'll make tests covering those cases and we´ll be confident we don´t break the ones already working.


SlavaRa:
WoW, :+1:

FlashDevelop/MainForm.cs
```diff
@@ -1032,6 +1031,8 @@ private void InitializeComponents()
             this.LocationChanged += new EventHandler(this.OnMainFormLocationChange);
             this.GotFocus += new EventHandler(this.OnMainFormGotFocus);
             this.Resize += new EventHandler(this.OnMainFormResize);
+
+            ScintillaManager.ConfigurationLoaded += ApplyAllSettings;
```
elsassph: Shouldn't the call to `ApplyAllSettings` be removed from `OnMainFormLoad` then?

Tests/External/Plugins/ASCompletion.Tests/Completion/ASCompleteTests.cs
```diff
@@ -0,0 +1,692 @@
...
+                var pluginUiMock = new PluginUIMock(pluginMain);
+                pluginMain.MenuItems.Returns(new List<System.Windows.Forms.ToolStripItem>());
+                pluginMain.Settings.Returns(new GeneralSettings());
+                pluginMain.Panel.Returns(pluginUiMock);
+                ASContext.GlobalInit(pluginMain);
+                ASContext.Context = new HaXeContext.Context(new HaXeSettings());
+
+                var sci = GetBaseScintillaControl();
+                sci.Text = "var a:String = (2 > 3) ?";
+                sci.ConfigurationLanguage = "haxe";
```
elsassph: Maybe we should have some enum at ASComplete level defining if we're in AS3, Haxe, etc. mode.

Tests/External/Plugins/ASCompletion.Tests/Completion/ASGeneratorTests.cs
```diff
@@ -0,0 +1,611 @@
...
+                Assert.AreEqual(34, funcBodyStart);
+                Assert.AreEqual("function test<T:{}>(arg:T):void{\r\n\r\n}", sci.Text);
+            }
+        }
+
+        public class GenerateJob : ASGeneratorTests
+        {
+            private ClassModel GetAs3ImplementInterfaceModel()
+            {
+                var interfaceModel = new ClassModel { InFile = new FileModel(), Name = "ITest", Type = "ITest" };
```
elsassph: Don't you think it would be more practical / scale better for tests to generate models using the parser?

Tests/External/Plugins/ASCompletion.Tests/Completion/ASGeneratorTests.cs
```diff
@@ -0,0 +1,611 @@
...
+                    }
+                });
+
+                return classModel;
+            }
+
+            [Test]
+            public void FieldFromParameterPublicScopeWithEmptyBody()
+            {
+                var pluginMain = Substitute.For<PluginMain>();
```
elsassph: I really think the generators tests should be made completely declaratively: input, operation, output.

elsassph:
This is colossal.

As much as NUnit look really clever, tests seem quite complicated to develop/extend: we need something as declarative as possible so test cases can be added without too much friction.

FlashDevelop/MainForm.cs
Neverbirth: I'll check it when I'm back

Tests/External/Plugins/ASCompletion.Tests/Completion/ASCompleteTests.cs
Neverbirth: Do you mean because of sci.ConfigurationLanguage? It's needed for loading the snippets/templates.

Tests/External/Plugins/ASCompletion.Tests/Completion/ASGeneratorTests.cs
Neverbirth: I already thought of it a lot in the past, I know what you mean, but the thing is that then you don't have the desired control over the input, and there are higher chances of having problems.

Neverbirth:
Of course things can change. I don't know all of NUnit API, so we may learn in the future new methods of doing things, but this is a start.

Of course it's huge, there are tons dependencies in order to make functions work. Some code could be moved to setup methods, and maybe use partial classes.

Parameterized tests could be used for some things, but I'm not sure I want to use it for the input of generators, because that way they would be less declarative, and a bit more tedious to debug or find problems.

Have you thought of something? Maybe we can ask for suggestion to the NUnit community.

Neverbirth:
Also for declarative assertions there are projects like FluentAssertions, but i don't think we need it.

FlashDevelop/MainForm.cs
Neverbirth: I checked it, but I'm not completely sure at the moment, I need some rest heh. That ApplyAllSettings call is an old one, I'd also say it can be removed, but I don't know if there is some reasoning behind it, because the old code although less OOP had a similar result.

Tests/External/Plugins/ASCompletion.Tests/Completion/ASCompleteTests.cs
```diff
@@ -0,0 +1,692 @@
...
+                ASContext.GlobalInit(pluginMain);
+                ASContext.Context = new AS3Context.Context(new AS3Settings());
+
+                var sci = GetBaseScintillaControl();
+                sci.Text = "function test(arg:String, arg2";
+                sci.ConfigurationLanguage = "as3";
+
+                var coma = ASComplete.DisambiguateComa(sci, 30, 0);
+
+                Assert.AreEqual(ComaExpression.FunctionDeclaration, coma);
```
rprouse: It looks like all of the tests in this class follow the same pattern. In that case, you should reduce the code to one test that takes in the parameters that are different between each test and then use multiple `[TestCase(...)]` attributes to feed in the parameters.

On a quick scan, you could reduce this entire class down to one method with many TestCase attributes.

Tests/External/Plugins/ASCompletion.Tests/Completion/ASGeneratorTests.cs
```diff
@@ -0,0 +1,611 @@
...
+            {
+                var sci = GetBaseScintillaControl();
+                // TODO: Should we reindent second line?
+                sci.Text = "function test():void{\r\n\t\t\t}";
+                sci.ConfigurationLanguage = "haxe";
+                sci.Colourise(0, -1);
+                int funcBodyStart = ASGenerator.GetBodyStart(0, 1, sci);
+
+                Assert.AreEqual(26, funcBodyStart);
+                Assert.AreEqual("function test():void{\r\n\t\t\t\r\n}", sci.Text);
```
rprouse: Once again, whenever you see patterns repeating in your tests, try to abstract out the common code and use TestCase attributes to drive the differences. This test might become something like,

```C#
[TestCase("function test():void{\r\n\t\t\t}", "function test():void{\r\n\t\t\t\r\n}", Description = "End on second line")]
[TestCase("function test():void{\r\n}", "function test():void{\r\n\t\r\n}", Description = "End on second line no extra indent")]
public void BodyStartTests(string text, string expected)
{
    var sci = GetBaseScintillaControl();
    sci.Text = text;
    sci.ConfigurationLanguage = "haxe";
    sci.Colourise(0, -1);
    int funcBodyStart = ASGenerator.GetBodyStart(0, 1, sci);
 
    Assert.AreEqual(24, funcBodyStart);
    Assert.AreEqual(expected, sci.Text);
}
```

If there are more differences, add parameters for them.

Tests/External/Plugins/ASCompletion.Tests/Completion/ASCompleteTests.cs
Neverbirth: First, thanks for taking your time to come here. Second, since there are so many different cases, some still unsupported that will require debugging, etc don't you think it's a bit unproductive? I read a bit about TestCaseSource which seems like a better fit, but didn't fully document myself on it yet.

Tests/External/Plugins/ASCompletion.Tests/MainForm.cs
```diff
@@ -0,0 +1,378 @@
...
+using System.Drawing;
+using System.Windows.Forms;
+using FlashDevelop.Utilities;
+using PluginCore;
+using ScintillaNet.Configuration;
+using WeifenLuo.WinFormsUI.Docking;
+
+namespace FlashDevelop
+{
+    public class MainForm : Form, IMainForm
```
rprouse: When you are unit testing, you should make it obvious that you are working with a mock object. We would tend to call something like this MockForm or MockMainForm and maybe drop it in a Mocks subdirectory of the project. This way, as people add additional tests, they can easily see what mocks exist and extend them as necessary.

The fact that you need to derive from Form is also a code smell. You should be able to test with just an IMainForm so that no UI is created. That may require refactoring your project so that you are never casting an IMainForm to a Form or passing a Form around. Any methods/properties on Form that you are using should be in the interface.

I realize this can be a difficult goal to achieve though. Unfortunately, Microsoft did not design WinForms for testability. If they did, the child controls would be something like IControl. You can get there by creating your own interfaces for everything, but it can be more work than it is worth.

Tests/External/Plugins/ASCompletion.Tests/Completion/ASCompleteTests.cs
rprouse: TestCaseSource can work too, but I don't think you need it in your case.

The fact that there are so many test cases make is more productive in my opinion. You can add entirely new tests with one line of code and if you need to debug the one test, just use the visual studio adapter. It will show you a test for each TestCase so you can run or debug individual test cases. You can do the same from the command line, but then you need to know the test name.

Any tests that still require work can be excluded from the run by adding Ignore into the TestCase.

BTW, I am using NUnit 3.0 exclusively these days, so some of my code examples may not be correct for 2.6.4.

Neverbirth: Interesting, I didn't know about that debugging part. I use Resharper tho, I'll have to check it.

rprouse:
As @elsassph says, you need to take a more declarative approach to creating tests. As I stated in my review, using TestCase or TestCaseSource would greatly reduce the amount of code required for all of these tests. It will allow you to add new tests very easily and extend all of the tests in one place. For example, the way you have it, if you want to add new asserts to your tests, you need to do it in every test. TestCase allows you to add it once and apply everywhere.

Your project also requires a lot of setup for the tests. You have started creating some test utilities. Extending those to help with the setup will go a long way to making it easier for developers to add new tests.

Tests/External/Plugins/ASCompletion.Tests/Completion/ASCompleteTests.cs
rprouse: Resharper also supports running the individual test cases. If you are using Resharper, I would suggest holding off on going to NUnit 3 since their support for it isn't complete. That said, the Visual Studio test adapters we created work great and would allow everyone on your team to run the tests in Visual Studio even if they don't have Resharper.

Tests/External/Plugins/ASCompletion.Tests/MainForm.cs
Neverbirth: I know all of that, I cannot change the name of the class, for some reason needed setup code checks for the type name... There are a lot of dependencies and couplings between concrete classes, plus a lot of code is mainly static, but that's not something I have intention on changing for the moment.

Neverbirth:
Well, I have the hope the other collaborators will help with those helpers, making tests for other parts of code, complete existing tests with formatting options, tests for still uncovered code in current tests, etc.

So far tests have been really helpful, and aside for the first time where I had to discover which dependencies to setup and which was the best way to control those ones, making the tests was fast and easy.

Tests/External/Plugins/ASCompletion.Tests/Completion/ASGeneratorTests.cs
Neverbirth: @rprouse I'm curious about what you think about this. The parser is one of our custom classes. In my experience it would be bad practice to use it as we'd lose the control over the input, so it would be similar to use a concrete class with business logic where we can use a mock. As most of the times, using the concrete class has its advantages and disadvantages, plus feels wrong on a few ways.

What would you say?

Tests/External/Plugins/ASCompletion.Tests/Completion/ASCompleteTests.cs
Neverbirth: Is there any place documenting the missing features? I've googled, but didn't find any detailed list. For me using a runner is not always an option, some of my PCs are still on old VS versions, I didn't take my time to update yet (in one of them I don't have enough space and have to keep 2010 still around...).

External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -184,7 +184,8 @@ static public bool OnChar(ScintillaControl Sci, int Value, bool autoHide)
                         if (features.hasGenerics && position > 2)
                         {
                             char c0 = (char)Sci.CharAt(position - 2);
-                            if (c0 == '.' /*|| Char.IsLetterOrDigit(c0)*/)
+                            //TODO: We should check if we are actually on a generic type
```
elsassph: Maybe that was the reason why `Char.IsLetterOrDigit(c0)` was commented out.

External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -3582,7 +3587,17 @@ private static ComaExpression DisambiguateComa(ScintillaControl Sci, int positio
                         if (":,(=".IndexOf(c) >= 0)
                         {
                             string line = Sci.GetLine(Sci.LineFromPosition(position));
+                            //TODO: Very limited check, the case|default could be in a previous line, or it could be something else in the same line
                             if (Regex.IsMatch(line, @"\b(case|default)\b.*:")) break; // case: code block
+                            if (c == ':' && Sci.ConfigurationLanguage == "haxe")
```
elsassph: We need to replace all the `ASContext.Context.CurrentModel.haXe`, `ASContext.Context.CurrentModel.Version == 3`, `Sci.ConfigurationLanguage == "haxe"` by something less heuristic.

External/Plugins/ASCompletion/Model/ASFileParser.cs
```diff
@@ -1189,7 +1189,7 @@ public void ParseSrc(FileModel fileModel, string ba, bool allowBaReExtract)
                     else
                     {
                         // valid chars for identifiers
-                        if (c1 >= 'A' && c1 <= 'Z')
+                        if (char.IsLetter(c1))
```
elsassph: Accented characters in identifiers? Is it worth the additional processing effort?

External/Plugins/ASCompletion/Model/ASFileParser.cs
```diff
@@ -1246,7 +1246,7 @@ public void ParseSrc(FileModel fileModel, string ba, bool allowBaReExtract)
                                     }
                                 }
                             }
-                            else if (inGeneric && (c1 == ',' || c1 == '.' || c1 == '-' || c1 == '>' || c1 == ':' || c1 == '(' || c1 == ')'))
+                            else if (inGeneric && (c1 == ',' || c1 == '.' || c1 == '-' || c1 == '>' || c1 == ':' || c1 == '(' || c1 == ')' || c1 == '{' || c1 == '}' || c1 == ';'))
```
elsassph: You can have `;` in a generic type declaration?

External/Plugins/ASCompletion/Completion/ASComplete.cs
Neverbirth: The change was fine in all the tests I made (plus the build used by my workmates has this change, and nothing strange was detected in all this time), but I don't like this kind of simplistic check, there are cases where c0 may not return true and we are still in a generic, also, on current development branch if you try in Haxe to write .<, no matter where you are, it will always display the completion list with the available classes.that's the reason for the comment. This is the reason of the comments, the change is an improvement, but not a proper fix.

FlashDevelop/Managers/ScintillaManager.cs
```diff
@@ -20,14 +20,51 @@ namespace FlashDevelop.Managers
...
+
+        private static void Initialize()
+        {
+            if (!initialized)
+            {
+                lock (initializationLock)
+                {
+                    if (!initialized)
+                    {
+                        LoadConfiguration();
```
elsassph: Why the deferred initialization?

External/Plugins/ASCompletion/Completion/ASComplete.cs
Neverbirth: I agree, but better than nothing...

External/Plugins/ASCompletion/Model/ASFileParser.cs
Neverbirth: Yes, by using anonymous types as constraints.

Tests/External/Plugins/ASCompletion.Tests/Test Files/parser/haxe/IdentifiersWithUnicodeCharsTest.hx
```diff
@@ -0,0 +1,10 @@
+package;
+
+class Test
+{
+	public function thísIsVälid():String
```
elsassph: Maybe in AS3 but not in Haxe :P

External/Plugins/ASCompletion/Model/ASFileParser.cs
Neverbirth: I want to modify this part, because it's possible, but in Haxe I think it depends on some factors and so I prefer it to be disabled for that one, unless I made something wrong in my tests, but for AS is perfectly valid (AS allows for some strange things heh, and allows for some strange Unicode characters to be used as the replacement of other things). Also, $ is not valid in Haxe.

Processing is not affected that much with this method, it's already quite optimized, of course there it's slower, but really not that much.

elsassph:
I'm going to give it a try asap to code at work and then merge if nothing explodes.

FlashDevelop/Managers/ScintillaManager.cs
Neverbirth: The previous static constructor was also doing the initialization in a deferred way, but with the problem that it was causing a big coupling between it and MainForm, which is not nice, and since it was a static constructor there was no way to set anything before, or calling any event handler or callback.

This ScintillaManager class looks like it could be in PluginCore after removing some other couplings with MainForm, I think some plugins could make use of it.

Neverbirth:
Tomorrow I'll try to clean the other generator tests. There is still this in the air, but maybe can be decided after merge so in the meantime other people can look into making more tests:

"My main concern right now is that when running the tests the referenced FD assemblies are placed in the same directory as the FD executable. We should think of some way of being able to run the tests without doing this. Another thing still undecided is how to organize the tests: a test project for each FD project, or a single test project for all of them? Each one has its advantages and disadvantages. I'm OK with any."

Advantages of test Project per plugin:
- IMHO it's better organized and more clear since it will mimic the tested plugin structure.
- If one plugin is removed, you also remove the test Project in an easy way, the other way I'd say it's more problematic.

Advantages of single test Project:
- Faster compilation and testing times.
- For internal methods we only need to add a single friend assembly.
- Only a single Project in the solution explorer, and for the developer to load (more projects, more time to load, and more memory VS will use).

elsassph:
I think we can "group" tests logically - like ASCompletion.Test can include tests for all the completion contexts. If we start testing MainForm/PluginCore we should use another test project.

Neverbirth:
I've been trying to load the plugins from the plugin folder, but I've been unsuccessful :/, in my current setup the framework tries to resolve the assemblies before I can handle the AssemblyResolve event by myself. There are some other ways around it, but this one seemed the most reasonable and maybe elegant solution. I see there are plans for this case, but I didn't see any ETA.

Neverbirth:
If we make sure there are no exposed elements at class level that use the plugins (like function return or argument types), we can handle the AssemblyResolve event... but in this case it's not so easy because of PluginUIMock. If we extract an interface out of it we could mock it witthout doing a custom implementation, plus that way we could make any view for it (but in the first place, I don't think PluginUI should hold the lookup data).

Neverbirth:
Do you have some groups in mind? At this moment I can only think of the group you mentioned and the other plugins would end with a single test project too.

elsassph:
Sounds good and we'll create more projects when it will fell right.

elsassph:
Checked out the branch and when I build (Ctrl+Shift+B) I get a bunch of plugins generated at the root of `bin\Debug` instead of `bin\Debug\Plugins`. It prevents FD to run properly - is it related to unit tests?

PR #1013, opened 2015-12-20
All Images Update Properly

### LINKED ISSUE #832: Some themes features not activated until FD restart
I'm using a dark theme in the evening and a light theme in the daytime, so I switch themes couple times a day.
I've noticed some settings don't get applied until FD is restarted.

What I've noticed so far:

    DockPanelControl.BorderColor
    VS2005DockPaneStrip.ImageColor - refreshed by using document tabs list button then selecting another tab
    ImageManager.ImageSet
    ProjectTreeView.*Library*Color - refreshed by adding/removing some library in the project panel

Additionally, when switching between themes with a different VS2005AutoHideStrip.TabStyle setting, the auto hide items width doesn't get updated and becomes incorrect.

Some more settings I've noticed but forgotten.

Are there plans to fix those, or basically I should restart FD between theme changes?

wise0704:
I did it lel. All images are now properly adjusted when the theme changes.

Pretty much this should now be the new usage of `ImageList` class:

Previously, this is how it was used:
```C#
imageList = new ImageList();
imageList.ColorDepth = ColorDepth.Depth32Bit;
imageList.ImageSize = ScaleHelper.Scale(new Size(16, 16));
imageList.Images.Add(...);
```

Now:
 * If image lists are really needed, for example using `ListView` which does not allow you to set images of individual items, use the `ImageListManager` class: (e.g. [Lines 229~245](https://github.com/fdorg/flashdevelop/pull/1013/files#diff-8e8416681061d34913742532812eabcfR229))
 * If image lists are only used as a list of images, please use actual `Array` or `List<>` since `ImageList` makes copies of the images added so they will not be refreshed. (e.g. [Lines 64~126](https://github.com/fdorg/flashdevelop/pull/1013/files#diff-c55e126616f773458dcf028bf0007de2R64), [Lines 33~141](https://github.com/fdorg/flashdevelop/pull/1013/files#diff-1cc971095f0b0c016e65af9468f1050dR33)) 
 * If image lists are used only to resize images to 16x16, now use the `MainForm.FindImage16()` method. (e.g. [Lines 122~126](https://github.com/fdorg/flashdevelop/pull/1013/files#diff-419cbd06961df74728e1928b8a0610d9R122))

There are 3 ways of using `ImageListManager`.
 1. Use the `void Initialize()` method and `Populate` event. `Populate` gets fired whenever the theme changes, and it fires after the image list is cleared automatically. `Initialize()` simply raises `Populate` event. e.g. [When the `ImageList` doesn't add any images on initialization.](https://github.com/fdorg/flashdevelop/pull/1013/files#diff-13f49fd6afadb3183faece7e50f0fd5aR290)
 2. Use the `Initialize(EventHandler)` method. This is a simpler version of the option 1 since it assigns the handler to `Populate` event and invokes it. This method is used in most cases. e.g. [When the `ImageList` is populated at initialization.](https://github.com/fdorg/flashdevelop/pull/1013/files#diff-8e8416681061d34913742532812eabcfR229)
 3. Derive `ImageListManager` class and override `OnRefresh()` method. This is used to perform custom refresh routine (i.e. when you don't want/need to clear the image list). e.g. [When all image list items can/should be re-initialized without clearing&adding.](https://github.com/fdorg/flashdevelop/pull/1013/files#diff-57c37499597dd2f97367e72fb6a6b582R278)

`ImageListManager` can be implicitly converted to `ImageList`, and explicitly converted from `ImageList`. So it will feel more like a derived class of `ImageList`.
Explicit conversion from `ImageList` to `ImageListManager` is also possible, given the `ImageList` is retrieved from an existing `ImageListManager` and the instance is not garbage collected[*](https://github.com/fdorg/flashdevelop/pull/1013/files#diff-ef96615924d4c11c8ffd648f0772b0b3R71).

Closes #832.


External/Plugins/BookmarkPanel/PluginUI.cs
```diff
@@ -228,20 +228,22 @@ private void InitializeTexts()
...
             this.removeBookmarksItem.Image = PluginBase.MainForm.FindImage("402|4|4|4");
             this.searchButton.Image = PluginBase.MainForm.FindImage("484|26|-4|4");
         }
 
+        private void ImageList_OnInitialize(object sender, EventArgs e)
+        {
+            this.imageList.Images.Add("Bookmark", PluginBase.MainForm.FindImageAndSetAdjust("559|26|0|1"));
+            this.imageList.Images.Add("Info", PluginBase.MainForm.FindImageAndSetAdjust("229"));
+            this.imageList.Images.Add("Error", PluginBase.MainForm.FindImageAndSetAdjust("197"));
+        }
```
wise0704: Use `FindImageAndSetAdjust()` inside `ImageListManager.OnInitialize` event. It is equivalent to `ImageSetAdjust(FindImage(data, false))`, making sure that the image color is adjusted according to the theme but the image is not cached to be updated on theme change. (The cached image uses `WeakReference` however.)

External/Plugins/FlashDebugger/Controls/BreakPointUI.cs
```diff
@@ -132,11 +119,11 @@ private void init()
...
-            tsbRemoveSelected.Image = imageList.Images["DeleteBreakpoint"];
-            tsbRemoveFiltered.Image = imageList.Images["DeleteBreakpoints"];
-            tsbAlternateFiltered.Image = imageList.Images["ToggleBreakpoints"];
-            tsbExportFiltered.Image = imageList.Images["ExportBreakpoints"];
-            tsbImport.Image = imageList.Images["ImportBreakpoints"];
+            tsbRemoveSelected.Image = PluginBase.MainForm.FindImage("548|27|5|5");
+            tsbRemoveFiltered.Image = PluginBase.MainForm.FindImage("549|27|5|5");
+            tsbAlternateFiltered.Image = PluginBase.MainForm.FindImage("136|23|5|5");
+            tsbExportFiltered.Image = PluginBase.MainForm.FindImage("549|22|4|4");
+            tsbImport.Image = PluginBase.MainForm.FindImage("549|8|4|4");
```
wise0704: Here, the image list is only used as a "temporary" hash table of images. Not so temporary because it is stored as a static field as an unnecessary object in memory. Since `FindImage()` stores the images in cache already, retrieve them directly from the `ImageManager`.

The image list might have been used as a way to "resize" the images to 16x16, in case it is running in HDPI. However, since those buttons are contained in the tool strip `tsActions` and [`ImageScalingSize` of it is set to 16x16](https://github.com/fdorg/flashdevelop/pull/1013/files#diff-c55e126616f773458dcf028bf0007de2R248), the resize is unnecessary.

PluginCore/PluginCore/Controls/MessageBar.cs
```diff
@@ -121,13 +121,8 @@ public void InitializeComponent()
...
         {
-            ImageList imageList = new ImageList();
-            imageList.ImageSize = ScaleHelper.Scale(new Size(16, 16));
-            imageList.ColorDepth = ColorDepth.Depth32Bit;
-            imageList.Images.Add(PluginBase.MainForm.FindImage("196")); // warning
+            label.Image = PluginBase.MainForm.FindImage16("196"); // warning
             buttonClose.ImageEnabled = ResourceHelper.LoadBitmap("MessageBarClose.bmp");
-            label.ImageList = imageList;
-            label.ImageIndex = 0;
         }
```
wise0704: When `Image` property can be set directly, always use it instead of an image list.

External/Plugins/FileExplorer/PluginUI.cs
```diff
@@ -293,14 +289,21 @@ private void InitializeContextMenu()
...
-            this.imageList = new ImageList();
+            this.imageList = new ImageListManager();
             this.imageList.ImageSize = ScaleHelper.Scale(new Size(16, 16));
             this.imageList.ColorDepth = ColorDepth.Depth32Bit;
+            this.imageList.OnInitialize += ImageList_Initialize;
+            this.AddNonWin32Images();
             this.syncronizeButton.Image = PluginBase.MainForm.FindImage("203|9|-3|-3");
             this.browseButton.Image = PluginBase.MainForm.FindImage("203");
             this.fileView.SmallImageList = this.imageList;
         }
```
wise0704: `RefreshFileView()` cannot be called here, so don't call `Initialize()`.

External/Plugins/ProjectManager/Controls/Icons.cs
```diff
@@ -275,6 +275,77 @@ public static Image Overlay(Image image, Image overlay, int x, int y)
...
+                Images[DocumentClass.Index] = DocumentClass.Img;
+                Images[CommandPrompt.Index] = CommandPrompt.Img;
+                Images[CollapseAll.Index] = CollapseAll.Img;
+                
+                foreach (FDImage image in extensionIcons.Values)
+                {
+                    Images[image.Index] = image.Img;
+                }
+            }
+        }
```
wise0704: Here, it is much easier to re-assign all images instead of clearing and adding them in the correct order.

Meychi:
Just to follow current coding practices:

 * Avoid creating unnecessary spaces in code, comment lines if you want to clarify parts
* Use ImageListOnInitialize vs. ImageList_Initialize for event handler names

Otherwise everything seems to work nicely, tested at 150% on Win7.

wise0704:
I used to have the event handler names as `ImageList_OnInitialize` because the event name is `OnInitialize`, but since `On` doesn't really make sense, I renamed them anyways. (There is a method called `Initialize` so I had to add `On` on the event name.)

I tried to look for unnecessary spaces to remove, but couldn't find much? You mean blank lines, right? If you meant something else, please tell me so I can fix it :)

wise0704:
When I did this to try out the 32x32 images
image

The Project Manager **tree** has cropped images for images retrieved by `MainForm.FindImage()`. (But not the ones retrieved from its resources.)
image

Is this expected, since i'm not using true HDPI? Or should I fix this by replacing `FindImage()` by `FindImage16()`?

elsassph:
Problem with high dpi scaling (200% for me):
image

wise0704:
It seems like setting images using `ImageList.Images[Int32]` doesn't trigger image resizing unlike `Add(Image)` method. (Probs a bug)
I've fixed the issue.

Meychi:
This has some conflicts now.

PR #1017, opened 2015-12-22
Added Haxe/C# template

markknol:
It's missing the _project.png_ file.


Meychi:
I think the template should be consistent the AS3/Flex Mobile templates we have. Same naming and bat files under bat dir. What do you think?

markknol:
What naming is exactly different or how should it be? I used the template files based on the Haxe/Neko template.

elsassph:
The templates should include hints on how to load .NET DLLs (custom and STD). I'll try to share something.

markknol:
I've added the project image and changed the batch files

PR #1020, opened 2015-12-29
Decoupled Completion List and Call Tips from Scintilla and removed from MainForm

Neverbirth:
Some details can be seen [here](http://www.flashdevelop.org/community/viewtopic.php?p=54608#p54608), please read it.

Aside from the details provided in that post, this makes ASComplete more testable, and would allow for proper single Ctrl+Z with Smart Indentation enabled, and proper Ctrl + Click without resorting to hacks or low-level hooks.

I know this doesn't merge right now, I make the PR for others to analyze it, comment on it, etc.

IMHO, all of the benefits are worth losing the custom scroll bars, but for that part I've already had a lot of discussions with @wise0704, he is right now in holidays tho, if anyone want to discuss about that part people can reach me (@meychi ?).


Neverbirth:
Forgot another one, I don't recall all of the details right now, but back in the day from some research I did it semed like the work I did in this branch would also help to speed up "find all references" and the different refactor commands that use it.

I may forget some other improvements at this moment...

External/Plugins/FlashDebugger/Controls/BreakPointUI.cs
```diff
@@ -512,6 +557,63 @@ private void TstxtFilter_KeyDown(object sender, KeyEventArgs e)
             }
         }
 
+        #region Auto Completion
+
+        protected override bool ProcessCmdKey(ref Message msg, Keys keyData)
+        {
+            // Ctrl+Space is detected at the form level instead of the editor level, so when we are docked we need to catch it before
+            if (keyData == (Keys.Control | Keys.Space) || keyData == (Keys.Control | Keys.Shift | Keys.Space))
```
elsassph: I'm confused: why do you have key handling inside `BreakPointUI`?

Neverbirth: Because as a sample/obvious addition after making the decoupling, I added the completion list to some other places inside FD. That one is [this](http://imgbox.com/Oq6h9yN1).

PluginCore/PluginCore/Controls/CompletionList.cs
```diff
@@ -19,37 +19,25 @@ public class CompletionList
         /// <summary>
         /// Properties of the class 
         /// </summary> 
-        private static System.Timers.Timer tempo;
-        private static System.Timers.Timer tempoTip;
-        private static System.Windows.Forms.ListBox completionList;
+        internal static CompletionListControl completionList;
```
elsassph: We probably want to drop this wrapper class if it only forwards calls to the instance.

Neverbirth: Do you mean to remove the CompletionList class? It is for making the change completely backwards compatible. Plus removing it would require to change quite several things, but would be possible of course.

There is a comment in the code talking about how it could be done.

elsassph:
I generally like the decoupling but moving the list in a new window may have problematic consequences with Wine. @Meychi

External/Plugins/FlashDebugger/Controls/BreakPointUI.cs
elsassph: Yes I figured it later.

Neverbirth:
Most/All of the logic used is quite standard, I haven't tried it myself because I don't have it, but I've been told by other users that it's working fine. But it's always nice to get some more confirmation, in the thread referenced I posted a build.

PluginCore/PluginCore/Controls/CompletionList.cs
elsassph: Sure, that would be a good start.

Meychi:
I'm pretty uneasy with this level or refactoring fearing API or Wine breakages. PR's should be smaller and only include related changes.

Neverbirth:
Well, that's why I made PR #1026.

Meychi:
Ah ok, great.

Neverbirth:
Sure, I can close it if you want. I have it open so anyone can check the whole thing.

Meychi:
I think its better to not have it open here. Maybe link to an issue from your fork?

Neverbirth:
Maybe it's enough with the link in the other PR. What is the advantage of creating an issue in my fork with a link here? Will it add some special marker somewhere or similar?

Issue #960, opened 2015-11-25
[Haxe] Editor auto-indents the lines after variables starting with "class"\"enum"


YellowAfterlife:
Say, you have
```haxe
class Some {
	public var classList:Array<Dynamic> = [];
	|
}
```
upon writing `public ` on the next line, the editor will indent it to level 2. Changing "classList" to "clasList" or any other name (not starting with the keyword) will have it work as usual.
The same happens if a variable name starts with "enum".
Minor stuff.

elsassph:
Should be fixed now in dev builds.

PR #1030, opened 2016-01-14
When looking up keywords, verify they have spaces around

elsassph:
(eg. `classList` isn't a `class` declaration)

Fixes issue #960

### DIFF
--- a/External/Plugins/ASCompletion/Completion/ASComplete.cs
+++ b/External/Plugins/ASCompletion/Completion/ASComplete.cs
@@ -1257,20 +1257,43 @@ namespace ASCompletion.Completion
             return true;
         }
 
+        /// <summary>
+        /// Lookup type declaration keywords anywhere in the provided text
+        /// </summary>
         private static bool IsTypeDecl(string line, string[] typesKeywords)
         {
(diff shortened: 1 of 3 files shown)


External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -1257,22 +1257,45 @@ static private bool HandleDeclarationCompletion(ScintillaControl Sci, string tai
...
         }
 
+        /// <summary>
+        /// Look if the provided text starts with any declaration keyword
+        /// </summary>
         private static bool IsDeclaration(string line, ContextFeatures features)
         {
             foreach (string keyword in features.accessKeywords)
-                if (line.StartsWith(keyword)) return true;
+                if (line.StartsWith(keyword) && SpaceFollows(line, keyword)) return true;
```
Neverbirth: To me this "line.StartsWith(keyword) && SpaceFollows(line, keyword)" even if it's used in only two places looks like a total candidate to be converted to an extension method in some other file. And since the logic used in IsTypeDecl is more or less similar, maybe the same could be applied.

PR #1031, opened 2016-01-14
Fix auto indent of closing brace

elsassph:
When pressing Enter on a closing brace `}`, move it to its own line and fix indentation.
Eg.
```javascript
var foo = {|};

var foo = {
    |
}
```

Note: fixing a bad code change that tended to delete the closing brace because it was based on closing brace automatic insertion, which, by design, could NOT insert a closing brace, resulting in a confusing missing brace.

### DIFF
--- a/External/Plugins/ASCompletion/Completion/ASComplete.cs
+++ b/External/Plugins/ASCompletion/Completion/ASComplete.cs
@@ -1008,13 +1008,16 @@ namespace ASCompletion.Completion
                 int line = Sci.LineFromPosition(position);
                 if (line == 0)
                     return;
-                string txt = Sci.GetLine(line-1).TrimEnd();
+                string txt = Sci.GetLine(line - 1).TrimEnd();
                 int style = Sci.BaseStyleAt(position);
 
+                // move closing brace to its own line and fix indentation
                 if (Sci.CurrentChar == '}')
                 {
-                    Sci.DeleteForward();
-                    AutoCloseBrace(Sci, line);
+                    var openingBrace = Sci.SafeBraceMatch(position);
+                    var openLine = openingBrace >= 0 ? Sci.LineFromPosition(openingBrace) : line - 1;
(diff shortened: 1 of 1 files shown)


External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -1008,13 +1008,16 @@ static private void HandleStructureCompletion(ScintillaControl Sci)
...
                 int style = Sci.BaseStyleAt(position);
 
+                // move closing brace to its own line and fix indentation
                 if (Sci.CurrentChar == '}')
                 {
-                    Sci.DeleteForward();
-                    AutoCloseBrace(Sci, line);
+                    var openingBrace = Sci.SafeBraceMatch(position);
+                    var openLine = openingBrace >= 0 ? Sci.LineFromPosition(openingBrace) : line - 1;
+                    Sci.InsertText(Sci.CurrentPos, LineEndDetector.GetNewLineMarker(Sci.EOLMode));
```
Neverbirth: Scintilla has a shortcut method for this: Sci.NewLine()

elsassph: Good reminder, but `Sci.NewLine` is going to change the cursor position as well. I'll give it a try though.

Neverbirth: Oh, I don't remember about that detail, and googling didn't return any useful info. I cannot test it right now.

elsassph: `Sci.NewLine()` simulates pressing Enter, so logically the cursor will follow. Inserting text should only affect cursor position if the text is inserted before it.

elsassph: I looked at it and using this function gets us in an infinite loop ;)

PR #1041, opened 2016-01-26
Update Scintilla to 3.6.3

Gama11:
(latest release from the 18th January)

The changelog can be found here: http://www.scintilla.org/ScintillaHistory.html
The previously used version was 3.2.4.


External/3rdParty/Scintilla/additional_keywords.patch
```diff
@@ -1,65 +1,65 @@
...
-+#define SCE_C_WORD4 25
-+#define SCE_C_WORD5 26
+diff --git a/include/SciLexer.h b/include/SciLexer.h
+--- a/include/SciLexer.h
++++ b/include/SciLexer.h
+@@ -177,6 +177,9 @@
+ #define SCE_C_USERLITERAL 25
+ #define SCE_C_TASKMARKER 26
+ #define SCE_C_ESCAPESEQUENCE 27
++#define SCE_C_WORD3 100
```
Gama11: I started these at 100 to hopefully avoid conflicts in the future so patches can be applied more easily. Also helps distuingish custom and "real" defines.

External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -3686,8 +3686,8 @@ static public bool IsTextStyleEx(int style)
             return style == 0 || style == 5 /*word (secondary keywords)*/
                 || style == 10 /*punctuation*/ || style == 11 /*identifier*/ 
                 || style == 16 /*word2 (secondary keywords: class name)*/ 
-                || style == 19 /*globalclass (primary keywords)*/ || style == 23 /*word3 (add keywords3)*/
-                || style == 24 /*word4 (add keywords4)*/ || style == 25 /*word5 (add keywords5)*/
+                || style == 19 /*globalclass (primary keywords)*/ || style == 100 /*word3 (add keywords3)*/
+                || style == 101 /*word4 (add keywords4)*/ || style == 102 /*word5 (add keywords5)*/
```
Gama11: Shouldn't these functions be using the `Lexers.CPP` enum instead of hardcoding the values? :/

PluginCore/ScintillaNet/Lexers.cs
```diff
@@ -80,9 +80,13 @@ public enum CPP
         TRIPLEVERBATIM = 21,
         HASHQUOTEDSTRING = 22,
         PREPROCESSORCOMMENT = 23,
-        WORD3 = 24,
-        WORD4 = 25,
-        WORD5 = 26,
+        PREPROCESSORCOMMENTDOC = 24,
+        USERLITERAL = 25,
+        TASKMARKER = 26,
+        ESCAPESEQUENCE = 27,
```
Gama11: This could be interesting to get something like this:



I don't think the set of escape sequences can be customized though, so things that aren't escape sequences in Haxe [but C++](http://en.cppreference.com/w/cpp/language/escape) would get highlighted.

Gama11: Actually, it looks like the C++ escape sequences are not highlighted, it simply highlights `\` and the character after it.

elsassph:
Looks good; no obvious regression?

Gama11:
Haven't noticed anything, the custom highlighting still works. Didn't test too extensively though.

FlashDevelop/Bin/Debug/Settings/Languages/Haxe.xml
```diff
@@ -39,9 +39,9 @@
 				<keyword key="1" class="haxe-additional-keywords" />
 				<keyword key="2" class="haxe-documentation-keywords" />
 				<keyword key="3" class="haxe-primary-keywords" />
-				<keyword key="5" class="haxe-additional-keywords3" />
```
Meychi: Only Haxe updated? And this will break on all existing config/theme files. We need to keep the old indexes.

We could force refresh configs but it still breaks on custom or standalone configs.

elsassph: Ouch I wasn't careful enough here.

Gama11: Fixed in #1042.

PR #1042, opened 2016-01-27
Scintilla update: use the same keyword IDs for additional keywords

Gama11:
Otherwise old themes are incompatible, see #1041.


markknol:
Seems to work for me now :smile:

PR #1048, opened 2016-01-28
Korean Locale

### LINKED ISSUE #1000: Add Korean Language Support??
# Congrats guys on the 1000 #references!!!

Anyways, 
I see that a lot of developers in Korea uses FlashDevelop.
Wouldn't it be nice if FlashDevelop also supports Korean?

I can do the job of adding the localized resource file `ko_KR`.
@Meychi

wise0704:
Hi I'm back, kind of :)

It's been a really busy month for me, and I expect next month to be keeping me even busier (sadface).
I had this branch for a really long time untouched (since around Christmas I believe), with only few entries left in the `ko_KR.resX` file to be translated. I wanted to complete this earlier, as my original plan was to finish it before 2015 ended (which I might have achieved seeing I could finish the rest in one day), but unfortunately I couldn't manage more than a few minutes of spare time in a day, so instead of using up those precious time fragments to progress slowly, I decided to just wait until things started to slow down a bit. Ended up having to catch up quite a lot of things.

Anyways, I managed to take advantage of some free time that unexpectedly came to me, and manage to translate the `en_US.resX` resource file, and also the localization of the Start Page. 
I made a modification to the `TextHelper` class, so it accurately deals with strings like `Open (&O)`, which is the syntax used by languages that don't use the alphabets (Korean, Chinese, Japanese, etc.).

I'm pretty sure I found more places to make translations and noted down which ones, but I lost the note haha.
Let me know if there are more things to work on :)

Closes #1000 ...?


Neverbirth:
Good to see you back, and hope you can get free time again soon.

PR #1051, opened 2016-01-28
Add packages/ to .gitignore

Gama11:
The new unit test project from #1012 installs NuGet packages which leads to untracked files.


Neverbirth:
Everything is open to debate, for example, maybe people can decide that adding the NUnit libraries directly to FD may be better, but IMHO, this is the best option.

PR #1058, opened 2016-01-30
appman.xml: Advance -> Advanced

Gama11:
Seems to be a typo, it's "Advanced" in this thread as well: http://www.flashdevelop.org/community/viewtopic.php?f=4&t=12200


Gama11:
Updated.

PR #1060, opened 2016-02-02
Update appman.xml

CamWiseOwl:
Add WakaTime plugin to Appman.

### DIFF
--- a/appman.xml
+++ b/appman.xml
@@ -519,5 +519,18 @@
 			<Url>https://github.com/mikicho/findandreplaceex/releases/download/v1.1/Find.ReplaceEx.fdz</Url>
 		</Urls>
 	</Entry>
+	<Entry>
+		<Id>wakatime</Id>
+		<Name>WakaTime</Name>
+		<Version>1.1.0</Version>
+		<Build>1</Build>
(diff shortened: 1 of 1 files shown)


Meychi:
Thanks! Nice to see new people contributing :)

You should to change the FDZ's extraction path to $(BaseDir)/Plugins to it to work in standalone and normal mode correctly. UserAppDir should not be used directly.

I had some difficulties finding your plugin's ui and i suggest adding it under Tools or offer the setting object for the main setting ui as other plugins do.

Neverbirth:
@CamWiseOwl  I think you missed @Meychi comment?

CamWiseOwl:
I could've sworn I wrote a comment other day.

@Neverbirth sort of, I've been bogged down with real life / job type stuff, sorry.
@Meychi Originally it did use BaseDir, then I read contributing before I PR'd and changed it to UserAppDir, it's now back to BaseDir.

UI was in the Help menu, moved that to Tools.
The external config file is part of WakaTime itself, used by the CLI and other WakaTime plugins.

Meychi:
@CamWiseOwl, I changed the description in the contribution docs for this reason. :) I think its better to have it in tools. Ah ok.

PR #1066, opened 2016-02-05
ShortcutDialog minor cleanup

wise0704:
ContextMenu icons shouldn't change color according to the theme.
Now that there are no conflicts by default, suppressing warning message when reverting all to default is not needed.


wise0704:
hehe how many PR's did I make about the shortcut dialog

Meychi:
A lot, but its really nice now. :)

Issue #372, opened 2014-09-23
[haxe] Unspecified type + metadata leads to MemberModels with invalid types


Gama11:
With the following, `ASContext.Context.CurrentMember` is not null, which it should be. Additionally, the type of the member is `"allow"`. Seems like the logic that parses variables, looking for colons, might be a bit too aggressive.

`|` is the caret position.

``` haxe
class Main 
{
    var test = 1;

    @:allow(haxe.ds.ArraySort|)
    static function main() 
    {

    }
}
```

This breaks my fix for metadata paths (#336) and who knows what else.

Gama11:
This only seems fixed in cases where there already exists a closing paren while typing the module path (which is unlikely to be the case unless you have "Add closing braces" activated). I can reproduce the bug from #204 otherwise.

Neverbirth:
That's a wrong syntax case, but I agree the method shouldn't use the meta as information of the member.

Neverbirth:
Could you give more details? I cannot reproduce it currently :/.

Gama11:
Are you using the example code from the original issue? There has to be a member var using type inference before.

```haxe
var test = 1;

@:allow(
static function main() 
{

}
```

Neverbirth:
I see, you must have the whole code as you typed it. Even if you have:

```haxe
var test = 1;

@:allow(
static function main() 
```

It works fine

Issue #1065, opened 2016-02-04
[haxe] Function not recognized as such with metadata afterwards


Gama11:
```haxe
class Main
{
	private function func():Void
	{
	}
	
	@:allow(flixel)
	private var test:Int;
}
```

The outline panel shows `func` as a regular variable:

![](http://i.imgur.com/Qi46lFm.png)

The important part seem to be the metadata's brackets - works as expected with just `@:allow`.

This feels like a recent regression.

Gama11:
Also, if `func()` has no explicit return value, the metadata's name is used for that.

```haxe
class Main
{
    private function func()
    {
    }

    @:allow(flixel)
    private var test:Int;
}
```

![](http://i.imgur.com/MTjjBjb.png)

Neverbirth:
Yes, the second one is known, and it's been there for 2 years more or less.

elsassph:
This is fixed by the Haxe parsing improvements - we know understand Haxe metas.

PR #1068, opened 2016-02-08
Haxe parsing improvements

Neverbirth:
Generics in function types, and haxe metadata parsing.

Should fix issues #372, #379 and #1065.

### DIFF
--- a/External/Plugins/ASCompletion/Model/ASFileParser.cs
+++ b/External/Plugins/ASCompletion/Model/ASFileParser.cs
@@ -1590,7 +1590,7 @@ namespace ASCompletion.Model
                             }
                         }
 
-                        // metadata
+                        // metadata, contexts should define a meta keyword and a way to parse metadata
                         else if (!inValue && c1 == '[')
                         {
                             if (version == 3)
@@ -1607,8 +1607,17 @@ namespace ASCompletion.Model
                                 if (ba[i] == ']') curMember.Type = features.CArrayTemplate + "@" + curMember.Type;
                             }
                         }
+                        else if (!inValue && c1 == '@' && haXe)
+                        {
+                            var meta = LookupHaxeMeta(ref ba, ref i);
+                            if (meta != null)
+                            {
+                                carriedMetaData = carriedMetaData ?? new List<ASMetaData>();
+                                carriedMetaData.Add(meta);
+                            }
+                        }
 
-                        // Unreachable code????
+                        // Unreachable code???? plus it seems a bit crazy we have so many places for function types
                         // Haxe signatures: T -> T -> T 
                         else if (haXe && c1 == '-' && curMember != null)
                         {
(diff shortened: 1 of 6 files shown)


Gama11:
Nice, even this works:



However, it doesn't deal very well with the incomplete version of it:

Neverbirth:
IMHO since Haxe syntax can be so varied I'd prefer to first get the parser for Haxe separated in its own file and look into some of the wrong cases. I'd consider this as a separate thing from the current issue at hand.

External/Plugins/ASCompletion/Model/ASFileParser.cs
```diff
@@ -1606,8 +1607,18 @@ public void ParseSrc(FileModel fileModel, string ba, bool allowBaReExtract)
                                 if (ba[i] == ']') curMember.Type = features.CArrayTemplate + "@" + curMember.Type;
                             }
                         }
+                        else if (!inValue && c1 == '@' && haXe)
```
elsassph: Should we be stricter? Like making sure we're not inside a function body.

External/Plugins/ASCompletion/Model/ASFileParser.cs
```diff
@@ -1755,6 +1766,63 @@ private ASMetaData LookupMeta(ref string ba, ref int i)
...
+                    }
+                }
+                else if (c == 10 || c == 13)
+                {
+                    line++;
+                    if (c == 13 && i < len && ba[i + 1] == 10) i++;
+                }
+                else if (inString == 1 && c == '"') inString = 0;
+                else if (inString == 2 && c == '\'') inString = 0;
+                else if (inString > 0 && (c == 10 || c == 13)) inString = 0;
```
elsassph: What about multiline strings? It would be a reasonable use case to provide a multiline string as a template for instance in a meta.

Tests/External/Plugins/ASCompletion.Tests/Test Files/parser/haxe/MetadataTest.hx
```diff
@@ -0,0 +1,29 @@
+package test.test;
+
+class MetadataTest
+{
+	private function func():Void
+	{
+	}
+
+	@:allow(flixel)
```
elsassph: More tests like:

    @:build(ResourceGenerator.build("resource/strings.json"))
    @:build(TemplateBuilder.build('
        <div class="mycomponent"></div>')

External/Plugins/ASCompletion/Model/ASFileParser.cs
Neverbirth: I don't know, if the code for AS3 is safe enough, I'd say this one is too.

External/Plugins/ASCompletion/Model/ASFileParser.cs
Neverbirth: I think they should be handled correctly, but sure, I'll make unit tests to be on the safe side.

Neverbirth:
I've just found a parsing problem, I don't know if it's a regression with this PR, if it's something from my previous PR, or if it's something that has been present for some time but just noticed now. I'll look into it later with the corresponding unit test.

Neverbirth:
Just checked, it's not a regression with this PR. I also found that metadata in clases is still breaking other things, so I'll try to fix both problems tomorrow.

Neverbirth:
> I also found that metadata in classes is still breaking other things, so I'll try to fix both problems tomorrow.

I didn't close the meta correctly... and I remember, FD adds the meta to the FileModel, IMHO that's completely wrong, the metadata is decorating the class, not the file. I already did this change in my MXML branch, and it allowed to remove a few calls and made it easier to improve some things, So I'll make a different PR with this change tomorrow.

PR #1072, opened 2016-02-11
Add metadata to the current class, not the file model.

Neverbirth:
This is more correct as metadata is part of the class, not the file, and you may have multiple classes with independent metadatas, before this wouldn't work at all.

For the sake of completeness, added "Default" metadata parsing from SWC files.


Neverbirth:
I was just thinking that this may require to clear the completion cache. Also, thanks to this, the new tests for #1068 now pass correctly, so I'd say that one can be merged now too.

PR #1079, opened 2016-02-15
Clear current member at the final } if it's a method

Neverbirth:
This fixes methods setting some compiler metadata as its type if the meta has arguments and they are not fully written yet.

Fixes issue #372 (last Gama11's comment). EDIT: Fixed only after a function with no type. If it's a var it's not working right yet.

### DIFF
--- a/External/Plugins/ASCompletion/Model/ASFileParser.cs
+++ b/External/Plugins/ASCompletion/Model/ASFileParser.cs
@@ -952,6 +952,7 @@ namespace ASCompletion.Model
                         braceCount--;
                         if (braceCount == 0 && curMethod != null)
                         {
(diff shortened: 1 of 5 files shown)


Neverbirth:
Please, @Gama11 could you try out this? because after making the change I've seen that even without the changed code I cannot reproduce the case...

Neverbirth:
I close it until I get a reproducible case...

Gama11:
This doesn't fix the issue for me.

Neverbirth:
Just checked, it fixes it if the previous member to the wrong meta is a function, but not a var.

Neverbirth:
It seems that fixing the var case is not so trivial, if I clear the current member after the end of the var declaration there is some strange behavior in a case. I'll have to dedicate more time to that one

PR #1087, opened 2016-02-25
Toolbar dropdown size

elsassph:
Made target selector a bit larger by default, and now both drop downs have a configurable size.

### DIFF
--- a/External/Plugins/ProjectManager/Controls/FDMenus.cs
+++ b/External/Plugins/ProjectManager/Controls/FDMenus.cs
@@ -74,7 +74,7 @@ namespace ProjectManager.Controls
             ConfigurationSelector.DropDownStyle = ComboBoxStyle.DropDownList;
             ConfigurationSelector.AutoSize = false;
             ConfigurationSelector.Enabled = false;
-            ConfigurationSelector.Width = ScaleHelper.Scale(85);
+            ConfigurationSelector.Width = ScaleHelper.Scale(GetThemeWidth("ProjectManager.TargetBuildSelectorWidth", 85));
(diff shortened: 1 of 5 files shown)


Neverbirth:
Why not just make the dropdown to adjust to the max size? You can always set a minimum size and/or padding.

elsassph:
Good point - not sure how to do it though.

elsassph:
I'd still prefer a wider fixed width thing in fact.

Neverbirth:
There are several ways to do that, for example:

https://github.com/Neverbirth/flashdevelop/blob/build_configs/External/Plugins/ProjectManager/Controls/FDMenus.cs#L201

elsassph:
Ok, but if we do that you'll need to introduce min and max combo widths.

Meychi:
Lets just make it little bigger without any settings, ok?

elsassph:
@Meychi that was my idea - and for those wanting control there's a theme value that can be changed.

gene-pavlovsky:
Hey guys, off topic here! I only program in AS3 and have never seen the Target combo contain any options. Where is it actually used? Or is it possible to add custom targets by myself? At the moment I do it with ant, e.g. my build.xml has android, ios, desktop and web targets for my app. But maybe if I can see those in the target dropdown, would be nice?

elsassph:
The list has default items for some specific project types. Any value you type here will be available as an argument in commands: $(TargetBuild).

gene-pavlovsky:
@elsassph Thanks; since it's a drop-down, I didn't figure I could just type something in there. This raises a bunch more questions, I opened #1181 to discuss this.

PR #1100, opened 2016-03-02
Making it so that OpenResource panel remembers your last search.

gamedevsam:
This is a major annoyance for me, so this tiny improvement would make me very happy.


elsassph:
Is the text remembered at least pre-selected?

Meychi:
Maybe check the internals of the PR :)

PR #1107, opened 2016-03-05
Initial global custom scrollbar implementation...

Meychi:
May be heavy, requires Win32 and requires restart when activated.


linuxjackto3:
thanks www.movied.org

PR #1110, opened 2016-03-06
AppMan: Updated Solarized theme to 1.1 (matches FD 5.1).

gene-pavlovsky:


Meychi:
You didn't update the MD5, the install will fail if it doesn't match.

gene-pavlovsky:
My bad, updated the MD5 now.

PR #1123, opened 2016-03-11
[AppMan] Adds FP21 and AIR21

SlavaRa:
also need to create:
http://www.flashdevelop.org/appman/frameworks_flexsdk_9-21.zip


Meychi:
Created the package, thanks.

PR #1129, opened 2016-03-16
Externalized Loom support

Meychi:
Loom templates, snippets and plugin removed from default install and added to an extension


elsassph:
We need to add the build of these Dlls to CI to avoid regressions.

Meychi:
Is Loom support actively maintained? If not, we could drop it and someone else could take over.

elsassph:
It's not actively maintained for sure. Let's start with that and we'll rediscuss when it breaks.

PR #1134, opened 2016-03-18
Improves "Extract locale variable" when working with generic types

SlavaRa:
Before commit: coderefactor_extract_local_variable_for_generics_bug

After: ASCompletion.Tests/Test Files/generated/haxe/AfterGenerateExtractVariableGeneric.hx


External/Plugins/ASCompletion/Completion/ASGenerator.cs
```diff
@@ -2793,22 +2792,10 @@ public static void GenerateExtractVariable(ScintillaControl Sci, string NewName)
                         }
                     }
 
-                    int style = Sci.StyleAt(funcBodyStart + lastPos) & stylemask;
-                    if (ASComplete.IsCommentStyle(style))
```
elsassph: Looking generally alright, but don't we want to keep some of these conditions?

SlavaRa: The only condition is that we are interested in right now checking for comments.

elsassph: Ah right I see it later.

External/Plugins/ASCompletion/Completion/ASGenerator.cs
```diff
@@ -2793,22 +2792,10 @@ public static void GenerateExtractVariable(ScintillaControl Sci, string NewName)
...
-
-                    }
-                    else if (!ASComplete.IsTextStyle(style))
-                    {
-                        continue;
-                    }
-
-                    Sci.SetSel(funcBodyStart + lastPos, funcBodyStart + lastPos + expression.Length);
+                    var pos = funcBodyStart + lastPos;
+                    int style = Sci.StyleAt(pos) & stylemask;
```
elsassph: `Sci.BaseStyleAt` does the masking.

SlavaRa: @elsassph, code:
```
int stylemask = (1 << Sci.StyleBits) - 1;
...
Sci.StyleAt(pos) & stylemask
```
equals to
```
Sci.BaseStyleAt(pos)
```
I have to replace "Sci.StyleAt(pos) & stylemask" to "Sci.BaseStyleAt(pos)"?

elsassph: That's the point of this function: to make the code simpler.

Neverbirth:
Really nice to see an unit test for this PR. Do you think you could make some more? Since this is a untested generator I'm sure there are a lot of cases to add. Not asking for every possible case, of course, just the main ones that you can come of, and surely you can come up with some other failing or missing cases.

External/Plugins/ASCompletion/Completion/ASGenerator.cs
SlavaRa: ok, I will refactor soon

SlavaRa:
@Neverbirth yes, I'm going to do yet.

SlavaRa:
@elsassph @Neverbirth guys, I have to do something for this issue? Or I can add new tests and functionality in a separate issue(for example a clear indication to the type for AS3)?

Neverbirth:
I'd prefer to see more tests in this PR, because it will show that both the new, improved, behaviour, and the old ones, still work, and maybe fix even more things before merging.

SlavaRa:
Ok, I add in the near future

SlavaRa:
I added the test cases to test the extract local variable. This is a situation that I face every day on the main work.

elsassph:
Conflict.

SlavaRa:
@elsassph fixed.

PR #1147, opened 2016-03-29
[CodeRefactor] Improves user interaction for "Extract local variable" command.

SlavaRa:
Result: 
if there is only one occurrence
extractlocalvariable_withoutcontextualgenerator

if there are multiple occurrences:
extractlocalvariable_replacealloccurrences
extractlocalvariable_replaceinitialoccurrence


wise0704:
Just an idea.
Instead of "1 occurrence(s)" or "3 occurrence(s)", what about 
```
Extract local variable
Extract local variable for all occurrences
```
For reference, VS displays:
image

Or at least "Replace 1 occurrence" (? doesn't really sound right) and "Replace all/n occurrence(s)"

I haven't tested it but from the gifs it looks really nice.

PluginCore/PluginCore/Resources/en_US.resX
```diff
@@ -6018,4 +6018,16 @@ Custom locales must be an extension of a default locale, e.g. en-US.</value>
     <value>Keeps the caret centered when caret is navigated to.</value>
     <comment>Added after 5.1.1</comment>
   </data>
+  <data name="CodeRefactor.Description.ReplaceAllOccurrences" xml:space="preserve">
+    <value>Replace all highlighted occurrences of initial expression with new variable.</value>
+    <comment>Added after 5.2.0</comment>
+  </data>
+  <data name="CodeRefactor.Description.ReplaceInitialOccurrence" xml:space="preserve">
+    <value>Replace initial expression only.</value>
```
wise0704: I would suggest
`Replace the current expression.`
`Replace all highlighted occurrences of the expression with a new variable.`

SlavaRa:
I would like to support block scope for haxe and then you have to add another key.
But can be @Meychi  have thoughts on this?

PluginCore/PluginCore/Resources/en_US.resX
elsassph: What about `Only this occurence / Select all 3 occurences` (should appear only if there is more than one occurrence.

SlavaRa: and it works if only one entry, then the "menu" is not displayed

SlavaRa:
guys what about this?
@Meychi @elsassph @Neverbirth I have to do something else now?

PR #1155, opened 2016-04-01
Added values for flex and pointer-events

Horsetopus:
Not sure about some of the values.
Seems to work on haxedevelop, and at least provide some help.
But the syntax coloring still indicates the values as not recognized.


Gama11:
Highlighting is separate: https://github.com/fdorg/flashdevelop/blob/development/FlashDevelop/Bin/Debug/Settings/Languages/CSS.xml

Horsetopus:
Ok, Thanks.
updated the CSS.xml file, and added some other flex related css rules.

elsassph:
Nice!

PR #1162, opened 2016-04-06
Add a "Custom Command Prompt" setting

Gama11:
Falls back to cmd.exe if it's not a valid file.
This allows changing the hardcoded cmd.exe to whatever command prompt you like (cmder.exe, Git bash, PowerShell...)

### DIFF
--- a/FlashDevelop/MainForm.cs
+++ b/FlashDevelop/MainForm.cs
@@ -491,7 +491,12 @@ namespace FlashDevelop
         /// </summary>
         public string CommandPromptExecutable
         {
-            get { return File.Exists(Settings.CustomCommandPrompt) ? Settings.CustomCommandPrompt : "cmd.exe"; }
+            get
+            {
+                if (!String.IsNullOrEmpty(Settings.CustomCommandPrompt) && File.Exists(Settings.CustomCommandPrompt))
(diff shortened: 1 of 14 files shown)


FlashDevelop/MainForm.cs
```diff
@@ -487,6 +487,14 @@ public new String ProductName
         }
 
         /// <summary>
+        /// Gets the command prompt executable (custom or cmd.exe by default).
+        /// </summary>
+        public string CommandPromptExecutable
+        {
+            get { return File.Exists(Settings.CustomCommandPrompt) ? Settings.CustomCommandPrompt : "cmd.exe"; }
```
elsassph: If `cmd.exe` is the default value it will always fail so it seems wasteful.
Maybe default to an empty string and test that the string isn't empty first.

elsassph:
Good idea, although executing other prompts may require a different pattern of arguments.

Gama11:
I've tested cmder.exe and PowerShell, at least the way we set the working directory still works for those (which is probably the only thing that matters here?)

FlashDevelop/MainForm.cs
Gama11: Done!

PR #1172, opened 2016-04-08
Incorrect FindParameterIndex reference position when using compiler completion

### LINKED ISSUE #1149: [haxe] Compiler completion doesn't work with "Add Closing Braces"
This example makes use of Haxe's "import statics" feature, which FD doesn't know about, so you only get completion from the compiler:

Here's the same with "Add Closing Braces" enabled - no completion: 

So it seems that with "Add Closing Braces", you get only calltip completion from FDs internal engine.
__________________

`Static.hx`:

```haxe
package;

class Static {
	public static function test(s:String) {}
}
```

elsassph:
Fixes #1149


SlavaRa:
I checked this branch, all ok

Issue #1161, opened 2016-04-06
"import de.polygonal.ds.Map;" is being auto-added to the top of my class


andytwoods:
any tips on how to prevent "import de.polygonal.ds.Map;" from being auto-added to the top of my class I am working on in FlashDevelop, when I mention 'map'?
Must say, being driven slightly mad by this!
Many thanks,
Andy.

Gama11 suggests: "I see what you mean, that's pretty annoying indeed. Top-level imports should be give a higher priority in the completion popup. Try opening an issue on the FD repo!"

PS I originally raised this issue in the haxe google group: https://groups.google.com/forum/#!topic/haxelang/rWVybiZkpP0

Gama11:
![](http://i.imgur.com/anI6SwL.png)

Sorting of the completion list seems completely off to me here - it should be exactly reverse (closest match to least closest match).

ianharrigan:
Something quite annoying also happens with `Void` when using the cpp target, it magically pulls in an import `cpp.Void` which invariably isnt what you actually want.

Gama11:
Yeah, I' ve had that happen to me as well quite a few times..

SlavaRa:
now all ok

SlavaRa:
bump, now all ok

PR #1173, opened 2016-04-09
Modifying a bit weighting of completion list SmartMatch

elsassph:
Fixes issue #1161

Note: collateral damage is that abbreviation matching is now only enabled for all-uppercase words.

### DIFF
--- a/PluginCore/PluginCore/Controls/CompletionList.cs
+++ b/PluginCore/PluginCore/Controls/CompletionList.cs
@@ -479,11 +479,12 @@ namespace PluginCore.Controls
                     ICompletionListItem item;
                     exactMatchInList = false;
                     smartMatchInList = false;
+                    bool allUpper = word == word.ToUpperInvariant();
                     while (i < n)
                     {
                         item = allItems[i];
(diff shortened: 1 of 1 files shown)


Meychi:
@Neverbirth, @SlavaRa, @wise0704, comments?

Neverbirth:
I think I completely missed this PR until now... not a part of code I really know. I'd need to try out how the completion feel with the abbreviation change, by looking at the changes it seems to me the issue is already fixed without that behaviour change?

wise0704:
image
Since the method is `public`, I think you need to add a method with the old signature too...

elsassph:
It needs unit tests as well, because the weighting was changed a bit randomly.

Meychi:
@elsassph: Will you add tests here or should we merge as such?

elsassph:
We shouldn't merge without tests I believe. I'll try to do it quickly.

Meychi:
@elsassph, any progress? :)

elsassph:
Well... I haven't been able to run tests at all so far :/

Meychi:
Any progress?

PR #1192, opened 2016-04-16
HaxeComplete: fix position / usage positions for Haxe 3.3.0

Gama11:
The +1 that was required here previously was actually a Haxe debug fixed recently on the dev branch, which now leads to incorrect results without the version check.
Sci.WordEndPosition() is technically not required on latest Haxe anymore either.

### DIFF
--- a/External/Plugins/HaXeContext/Completion/HaxeComplete.cs
+++ b/External/Plugins/HaXeContext/Completion/HaxeComplete.cs
@@ -12,6 +12,7 @@ using System.Xml;
 using ASCompletion.Model;
 using PluginCore.Helpers;
 using System.Windows.Forms;
+using PluginCore.Utilities;
 
 namespace HaXeContext
 {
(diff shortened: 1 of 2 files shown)


Gama11:
Note that FD doesn't pick up on Haxe version changes until it's restarted, which is a bit unfortunate.

elsassph:
That is quite annoying to have this kind of "breaking change" in a dot release.

elsassph:
What do you mean by "doesn't pick up Haxe version change"? Do you mean if you update (replace) the Haxe SDK while FD is running? I'd say it's minor.

Gama11:
@elsassph Yep, that's what I meant.

Gama11:
You may want to consider using GitHub's new "squash and merge" option when merging PRs, at least for simple ones like this (avoids the merge commit): https://github.com/blog/2141-squash-your-commits

elsassph:
You're right, I didn't hesitate long enough to give it a thought :)

PR #1217, opened 2016-04-25
Fixes #221

### LINKED ISSUE #211: Refactor->Code Formatter should be greyed out when it's not supported for the current file
Looks broken otherwise.

### LINKED ISSUE #221: [haxe] Contextual code generation for getters/setters
Using the contextual code generation hotkey (Ctrl+Shift+1) to create getter and setter for a variable generates this:

``` haxe
function get_test():Int 
{
    return test;
}

function set_test(value:Int):Int 
{
    return test = value;
}

private var test(get, set):Int;
```

Which causes these errors when compiling:

"This field cannot be accessed because it is not a real variable
Add @:isVar here to enable it"

Wouldn't the correct generation be like this:

``` haxe
function get_test():Int 
{
    return test;
}

function set_test(value:Int):Int 
{
    return test = value;
}

private var test(get_test, set_test):Int;
```

SlavaRa:


SlavaRa:
fixes #221

SlavaRa:
@elsassph, please review this PR

PR #1229, opened 2016-05-10
Use FIPS 140-2 compliant algorithm if available when required

elsassph:
After installing Cisco VPN client, FlashDevelop stopped working because it's using non FIPS compliant  algorithms.

### DIFF
--- a/FlashDevelop/Dialogs/HashDialog.cs
+++ b/FlashDevelop/Dialogs/HashDialog.cs
@@ -233,23 +233,30 @@ namespace FlashDevelop.Dialogs
         private void InputTextBoxChanged(Object sender, EventArgs e)
         {
             Byte[] plainBytes = Encoding.UTF8.GetBytes(this.inputTextBox.Text);
-            if (this.hashComboBox.SelectedIndex == 1) 
+            try
             {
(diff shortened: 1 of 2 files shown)


Meychi:
???, AnyConnect?

elsassph:
Yes, Cisco AnyConnect. It will only work with FIPS enabled.
Maybe the change is due to Win10 Insider Preview, but FD basically doesn't work without this fix.

Meychi:
Seems its a local policy setting in AnyConnect and it will "break" other apps also:

http://www.cisco.com/c/en/us/td/docs/security/vpn_client/anyconnect/anyconnect30/administration/guide/anyconnectadmin30/ac08localpolicy.html

Meychi:
I don't have any issues but i have the setting off.

elsassph:
I don't have the problem at work (Win10 Pro regular) but at home (Win10 Insider Preview) AnyConnect won't work without the FIPS flag set and it kills FD/HD.
My change is backward compatible as it defaults to using the regular managed encryption classes.

elsassph:
(as a side note, everything on the computer is slower with the flag set)

PR #1239, opened 2016-05-25
Fixes generate static variable and static function

SlavaRa:
BUG in release 5.1.1, code for example:
```
package {
	public class Main {
		public function Main() {
			Main.test|;
		}
	}
}
```
result after generating variable:
```
package {
	public class Main {
		private var test;//must be private static var test;
		public function Main() {
			Main.test;
		}
	}
}
```

BUG in developmnet version if setting generate explicit scope is true, code for example:
```
package;
public class Main {
		test|;
	}
}
```
result after generating variable:
```
package;
public class Main {
	static private var test;
		this.test;//must be Main.test
	}
}
```

### DIFF
--- a/External/Plugins/ASCompletion/Completion/ASGenerator.cs
+++ b/External/Plugins/ASCompletion/Completion/ASGenerator.cs
@@ -2081,13 +2081,15 @@ namespace ASCompletion.Completion
             FlagType kind = job.Equals(GeneratorJobType.Constant) ? FlagType.Constant : FlagType.Variable;
             // evaluate, if the variable (or constant) should be generated in other class
             ASResult varResult = ASComplete.GetExpressionType(sci, sci.WordEndPosition(sci.CurrentPos, true));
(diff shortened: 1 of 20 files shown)


SlavaRa:
@elsassph @Neverbirth please review this PR.

Neverbirth:
There doesn't seem to be unit tests taking into account the explicit scope case, aren't they? They could be added, what do you think?

SlavaRa:
I doubt that I understood correctly your 1st comment(( You could say, in more detail, what should I do?

Neverbirth:
I mean that you could make tests where you set the explicit scope setting to true, and then validate that everything is working correctly. I hope I explained myself better this time.

I don't think it's mandatory in order to approve the PR, but if you can do it it will be nice to have more coverture of the code.

SlavaRa:
Now I understand, thank you, I will refactor soon

SlavaRa:
I'll reopen this PR tonight

PR #1244, opened 2016-06-05
Adding extension SwcBuild

wise0704:
I posted this extension to the [forum](http://www.flashdevelop.org/community/viewtopic.php?f=4&t=12458) a while ago, and I kinda forgot about it until now, kek.

I wasn't sure on which group to put this on, as it's not a plugin, but it's not quite a compiler either. So I put it in a new group as Extension. 

By the way, wouldn't it be easier if AppMan's Config.xml file refer to https://raw.githubusercontent.com/fdorg/flashdevelop/development/appman.xml rather than http://www.flashdevelop.org/appman.xml?


elsassph:
Pointing on github is a good idea but there are rate limits when hitting github.

PR #1245, opened 2016-06-06
Advanced Shortcut

wise0704:
I know you guys will love this!!!
I had this PR planned since last year, but only started to actually write it two weeks ago...

Basically, I ~~copied~~ implemented VS's shortcut function style - with this FD can support **extended shortcuts which has two parts of normal shortcuts!**

Currently there are 3 ways of how keyboard shortcut mechanism work in FD:
1. Handle `KeyEvent` dispatched to plugins.
2. Add shortcut to `ToolStripMenuItem.ShortcutKeys`, and add the item to the `MainForm`'s tool strip.
3. Add shortcut to `ToolStripMenuItem.ShortcutKeys`, and add the item to a context menu strip.

1 occurs inside `MainForm.ProcessCmdKey()`, and 2&3 occur in `ToolStripManager.ProcessCmdKey()` which is called in a `ProcessCmdKey()` method of the root form, which is `MainForm`.
So I hijacked `MainForm.ProcessCmdKey()` :)

While writing this I figured that the shortcut mechanism should be simplified in order to support even more great features such as multiple shortcuts per command, contextual shortcuts for commands, etc.
Which is to get rid of 2&3 above, and only handle `KeyEvent`, using its `Command` property.
That will be another major upgrade, so I didn't include it in this PR.

This is quite a large PR, so we need to make sure I didn't break anything!


PluginCore/PluginCore/Managers/ShortcutKeysManager.cs
```diff
@@ -0,0 +1,371 @@
...
+using System.Runtime.InteropServices;
+using System.Windows.Forms;
+using PluginCore.Controls;
+
+namespace PluginCore.Managers
+{
+    /// <summary>
+    /// A static manager class for advanced shortcut keys.
+    /// </summary>
+    public static class ShortcutKeysManager
```
wise0704: @elsassph This is the one I was talking about. This class uses `Reflection` frequently. Would there be performance issues? If so, would using the Expression API instead solve it?

FlashDevelop/Dialogs/ShortcutDialog.cs
```diff
@@ -631,31 +747,23 @@ private void DialogClosing(object sender, FormClosingEventArgs e)
...
         /// </summary>
-        private void DialogClosed(object sender, FormClosedEventArgs e)
+        private void ShortcutDialog_FormClosed(object sender, FormClosedEventArgs e)
         {
-            for (int i = 0; i < this.shortcutListItems.Length; i++) this.shortcutListItems[i].ApplyChanges(); 
+            for (int i = 0; i < this.shortcutListItems.Length; i++)
+            {
+                this.shortcutListItems[i].ApplyChanges();
+            }
             Globals.MainForm.ApplyAllSettings();
```
wise0704: Is it necessary to call `ApplyAllSettings()`? Can it not just call `ApplyAllShortcuts()` in `ShortcutManager`?

wise0704:
I did try to log summaries in every commit I make... I could be helpful to get a rough idea of what has been changed.

FlashDevelop/Managers/ShortcutManager.cs
```diff
@@ -4,233 +4,319 @@
...
+                if (File.Exists(FileNameHelper.ShortcutDataOld))
                 {
-                    ShortcutItem item = GetRegisteredItem(arg.Key);
-                    if (item != null) item.Custom = (Keys) Enum.Parse(typeof(Keys), arg.Value);
+                    File.Move(FileNameHelper.ShortcutDataOld, file);
+                }
+                else return;
+            }
+            var shortcuts = new List<Argument>();
+            shortcuts = (List<Argument>) ObjectSerializer.Deserialize(file, shortcuts, false);
```
elsassph: Should it have a `try/catch`?

PluginCore/PluginCore/Helpers/ShortcutKeysConverter.cs
```diff
@@ -0,0 +1,224 @@
...
+            if ((keys & Keys.Shift) == Keys.Shift)
+            {
+                return Shift + GetString(keys & Keys.KeyCode);
+            }
+            return GetString(keys);
+        }
+
+        private static unsafe ShortcutKeys ConvertFromStringInternal(string value)
+        {
+            // For performance reasons, use an unsafe context and char pointers.
```
elsassph: Can you quantify the benefit of using an unsafe pointer?
Is that something we should consider for the AS/Haxe model parser?

PluginCore/ScintillaNet/ScintillaControl.cs
```diff
@@ -2420,33 +2420,34 @@ public bool ScrollWidthTracking
         /// <summary>
         /// Adds a new keys to ignore
         /// </summary> 
-        public virtual void AddIgnoredKeys(Keys keys)
+        public virtual void AddIgnoredKeys(ShortcutKeys keys)
         {
-            ignoredKeys.Add((int)keys, (int)keys);
+            //ignoredKeys.Add((int)keys, (int)keys);
```
elsassph: What happens with `ignoredKeys`? Is it deprecated?

elsassph:
Generally it's really impressive to be able to set complex shortcuts like VS.

The keypreview dialog, although not high-dpi scaled is looking alright, if only a bit "compact" ;)

I find the changes in `LineEntry` dialog rather complex - is that worth/necessary to fully reimplement keyboard shortcuts in textfields? Or is that to demonstrate `HandleShortcutManually`?

The `ShortcutManager` code should probably be reviewed by @Neverbirth - that's too much WinForms magic for me :D

Some bugs:
- I tried to replace `Ctrl+Q` and `Ctrl+Shift+Q` with `Ctrl+Q, L` and `Ctrl+Q, B`. This works only if I restart FD, otherwise when I press `Ctrl+Q` it doesn't start a complex shortcut input and if I press `B` or `L` it just inserts the letter,
- when starting a shortcut (eg. `Ctrl+E`), I see in the status bar that FD is waiting for another key; but if I click in the editor (to cancel the sequence), I see a strange relayout whereby the Scintilla editor grows over the bottom tabs and quickly goes back to normal size.

FlashDevelop/Managers/ShortcutManager.cs
wise0704: The method is called inside a `try/catch` block already, inside `MainForm.InitializeMainForm()`.

PluginCore/PluginCore/Helpers/ShortcutKeysConverter.cs
wise0704: Unsafe code pretty much increases performance by eliminating managed processes, like array bounds check, type safety, etc.
On a second thought, I think this code isn't that critical to use unsafe.

Scintilla control already uses unsafe code on almost all string operations, and the string class uses unsafe internally too. However codes that loop through chars in strings can have their performance improved by using unsafe without worrying about multi-bytes (such as this)
image
If `ba` was `char*` instead of `string` (cool thing is that the syntax doesn't change) there wouldn't be any validity checks for the type of `ba`, checking if `ba` is `null`, index property accessing operation, or index boundary check. It would just be calculation of `ba + i + 1` and comparison to `'n'` etc.

I wouldn't worry about it too much, but I think it's not bad to consider it for better performance.

PluginCore/ScintillaNet/ScintillaControl.cs
wise0704: Yep. Since [this](https://github.com/fdorg/flashdevelop/pull/1245/files#diff-05f30fbc757ea5380e769c4a72b326e9L5323) is removed, `ignoredKeys` and `ignoreAllKeys` have no function at all.

wise0704: What is the actual desired function of `ignoreAllKeys`? `ignoredKeys` is identical to the `MainForm.IgnoredKeys` so it is removed.

Should [this](https://github.com/fdorg/flashdevelop/pull/1245/files#diff-05f30fbc757ea5380e769c4a72b326e9R5323) be changed to:

```c#
if (ignoreAllKeys ||
    ProcessCmdKey(ref m, (Keys) code | ModifierKeys) ||
    (ModifierKeys & Keys.Control) != 0 &&
    (ModifierKeys & Keys.Alt) == 0 &&
    'A' <= code && code <= 'Z')
{
    return true;
}
```

wise0704:
By keypreview dialog you mean `ShortcutModificationDialog`? That is kinda a draft and should be improved. For now I just made it so that it is capable of doing its task.

Yea, the changes to `LineEntry` is to demonstrate the method - it's not really necessary unless users would expect the basic shortcuts of .NET textfields (copy, cut, paste, select all, etc) to match with the setting's shortcuts. The extra lines of `e.Handled`, `e.SuppressKeyPress`, and `else if (...)` which complicated things were needed because the `ShortcutsEnabled` property is not working properly.

wise0704:
I can`t reproduce the second bug. Does it happen every time you cancel a sequence?

wise0704:
First bug is fixed.

wise0704:
How should the shortcut keys editor look like? Currently it's just a dialog form that displays the keys pressed and validates the apply button according to the shortcut.

Below is the default .NET editor for `Keys`. Would this layout be more user-friendly?
image

wise0704:
There is a bug in the .NET `ToolStripMenuItem.set_ShortcutKeys`, which is causing the `ToolStripMenuItemEx` to function wrongly when assigning a simple shortcut, since for simple keys it passes it to `base.ShortcutKeys`.

PluginCore/PluginCore/Managers/ShortcutKeysManager.cs
wise0704: @Neverbirth Can you review this class please?

wise0704:
Is the `ShortcutKeysEditorDialog` properly scaled in HDPI? I did add font scaling and set the fonts to the settings font.

elsassph:
I still see the second bug, but it's not critical. The shortcut dialog looks amazing!

wise0704:
I don't know why that's happening...
image
The problem should be here, but the only operation related to windows forms is `StatusLabelText = null`...

Meychi:
This PR changes lots of things in the PluginCore which i see as an issue regarding other 3rd party plugins, if they have been compiled for older releases. 

About the Win32 stuff, how is the functionality working on CrossOver?

Also the usage has now changed even if the info on bottom instructs to hit the keys when item is selected.

I'm not fully sure about this PR due to the possible breakages it may introduce.

wise0704:
I tried to keep all public signatures as is, such as `MainForm.RegisterShortcut(string, Keys)` and `ShortcutKeys` being able to be implicitly converted to `Keys`.

One thing I'm not sure is that `KeyEvent.Value` is now `ShortcutKeys` instead of `Keys`. It's implicitly convertible but would it still break 3rd party plugins? If so, I can always have it as a new property (like `Keys:ShortcutKeys`, making `Value:Key` as `=> (Keys) this.ShortcutKeys`), or have a new event dispatched before `KeyEvent`.

Also, not all shortcuts need to support extended keys (such as old 3rd party plugins). If `RegisterShortcut` was called with the old signature, or with `ToolStripMenuItem` instead of `...Ex`, the shortcut item is marked as "not supporting extended", and `ShortcutDialog` and others such will respect that. (E.g. `ShortcutDialog` won't allow setting an extended shortcut to that item. It will even show if an item supports extended keys with a small plus on the icon!)

wise0704:
Installed NavigationBar:
image
~~Although the difference in icons is not that easy to see...~~ Only simple shortcuts can be set to those shortcuts, and they work without problems.

Neverbirth:
For me it's ok.

wise0704:
I don't know how stuff works in CrossOver, and I'm not able to test it...

wise0704:
Since "select" and "double-click" are not _that_ much different, instead of invalidating 3 locales by completely replacing the whole sentences with English, I just modified the languages I can write.

Meychi:
Has conflicts now.

gene-pavlovsky:
While I'm used to FD shortcuts, and have a hard time figuring out how to do things in recently installed VS, I can see how this feature (I think VS calls shortcut sequences `chords`) is powerful and might make a lot of sense in the future, if FD introduces some of the actions VS (or even Notepad2) have, such as a bunch of text/block processing commands, more `insert ...`. I think one of nice feature of chords is that they will allow to group related shortcuts together with a common prefix, making it more intuitive and easy to remember a lot of shortcuts.
So, great work! And personally I hope this will be merged sooner or later.
Mika, I think for progress sometimes you have to break backward compatibility, if the gain is big enough. Unless the same functionality can be implemented side-by-side without breaking older plug-ins?

Neverbirth:
> have a hard time figuring out how to do things in recently installed VS

? Advanced shortcuts have been available since at least VS 2008. Maybe you set a different default keyboard layout in your latest install?

gene-pavlovsky:
@Neverbirth by recently installed (last year) I meant that was the first time I ever installed and used VS. It looks at first sight somewhat similar to FD (I guess FD's UI is inspired by VS), so I intuitively try to use shortcuts familiar to me from FD, and then something bad happens :) But I did only a few things in VS (playing around with FD, and writing a small Win32 console app enabling to use the FAR file manager together with bash shell), so I only had time and motivation to learn just a few basic chords so far.

wise0704:
We need to test out the backward compatibility using a plugin that captures shortcut keys using `KeyEvent.Value`. (I've already confirmed plugins that use `ToolStripMenuItem` for shortcuts.)
Is there any that you guys know of?

Neverbirth:
You could make a test plugin by yourself. I know, it's some extra time from your own time for this, but it's the best I can think of right now. I don't know of 3rd party plugins using it.

wise0704:
I tested with QuickNavigate, which does use `KeyEvent.Value`:
>image

Apparently, the types need to match exactly:
>image

--

So guys, which option would you prefer to be used by new plugins:
 1.  `ShortcutEvent` which is fired on both extended and simple shortcuts, and `KeyEvent` fired only on simple shortcuts after `ShortcutEvent` is dispatched.
 2. `KeyEvent` has a new property called `Keys`, which contains the `ShortcutKeys` type. Old `Value` property returns `(Keys) this.Keys` (which means it returns `Keys.None` if the key is extended).

Neverbirth:
Being QuickNavigate maintained by Slavara... in this case maybe I would go on without worrying... were you able to find more plugins making use of `Value`?

Meychi:
We should not break existing apis so that old plugins continue to work. I think this is a high priority due to our project size and resources.

elsassph:
I think we're being too cautious here, let's just merge and move forward.

wise0704:
Guys, I can keep backward compatibility - it's not that hard.

It's just that I don't know if this should be included in the next major update, for FD 6. If so, then this can really wait and also include more stuff I'm planning, which are multiple shortcuts per command and contextual shortcuts, which I think will definitely break backward compatibility.

If this is better merged for 5.2, then I agree with not breaking the existing API. If the API looks horrible after the fixes, it can also be put to how it is now when I'm working on the features mentioned above.

wise0704:
Looking at `Interfaces.cs`, The only changes I need to make are:
 - `GetShortcutItemKeys()` return type back to `Keys`
 - `GetShortcutItemId()` parameter type back to `Keys`
 - `IgnoredKeys` type back to `List<Keys>`

wise0704:
There are more work to do since #952 was merged and it has some code that should be using the new `HandleShortcutManually()`.

External/Plugins/CodeRefactor/Commands/InlineRename.cs
```diff
@@ -804,34 +805,30 @@ bool IMessageFilter.PreFilterMessage(ref Message m)
...
+                    case "EditMenu.Undo":        PerformUndo(); break;
+                    case "EditMenu.Copy":        sci.Copy(); break;
+                    case "EditMenu.Cut":         if (sci.SelTextSize > 0) sci.Cut(); break;
+                    case "EditMenu.ToLowercase": sci.UpperCase(); break;
+                    case "EditMenu.ToUppercase": sci.LowerCase(); break;
+                    case "Scintilla.ResetZoom":  sci.ResetZoom(); break;
+                    case "Scintilla.ZoomIn":     sci.ZoomIn(); break;
+                    case "Scintilla.ZoomOut":    sci.ZoomOut(); break;
+                    default:
+                        //PluginBase.MainForm.StatusLabelText = string.Format("Shortcut \"{0}\" cannot be used during renaming.", shortcutId);
```
wise0704: Any opinions on this line?

wise0704:
Last commit added a setting to disable extended shortcuts for those users who dislike it.

Neverbirth:
I'm with Philippe. I'm all for merging this one.

wise0704:
It won't break any 3rd party plugins anymore either :)

Neverbirth:
Merged through a merge commit in case it's needed to revert or something. I don't think so, but it's not tested in Crossover AFAIK. I'll investigate how we can improve this Crossover support testing because for most of us it's a problem...

Gama11:
Isn't it undoable just as easily without a merge commit? It's only a single commit if squashed after all. :)

Neverbirth:
Oh, yeah, you are right. Anyway, since this is a big one I wanted to keep history, If Mika or Philippe would prefer to squash it they could do it.

wise0704:
Although I wanted this PR to be merged, I don't think we had much discussion on when or which version this feature should've been included... since this is a really major visible change to the users' usage experience.
I assume the next version is 5.2. Would it be alright guys? Maybe okay since it won't break 3rd party plugins. Though I feel like we should have gotten confirmation from both @Meychi and @elsassph.

Neverbirth:
@elsassph gave green light for it. Mika is away, I tried to talk with him some days ago... I don't know when he is going to be back.

Again, this can be reverted at any moment, so I don't think it's a big deal, and in the meantime we may get some useful feedback.

Why is it "a really major visible change to the users' usage experience"? I would agree if it would change default behaviors, but it's not the case.

wise0704:
Well, in my view I think it does change the default behaviour...
But if they said alright, well then no problem! I was just making sure.

Neverbirth:
> Well, in my view I think it does change the default behaviour...
I'll try to play with it more in depth ASAP and decide. But back in the day I didn't notice anything strange :S.

> But if they said alright, well then no problem! I was just making sure.
As I said, I don't know about Mika.

gene-pavlovsky:
Guys, I'm eager to try this. So if I install the latest dev build, how do I
set up some advanced shortcuts?

wise0704:
@gene-pavlovsky You can either manually change the shortcuts using the shortcut dialog, or you can try out the visual-studio-like shortcut set that I included (import button is also in the shortcut dialog).
Also, check that `FlashDevelop.DisableExtendedShortcutKeys` setting is `false`, which should be by default.

gene-pavlovsky:
Thanks for the explanation, sounds simple enough.
Unfortunately after I installed the latest development build, running it gave me around 10 identical exceptions:

```
Method not found: 'Boolean ASCompletion.Completion.ASComplete.OnShortcut(System.Windows.Forms.Keys, ScintillaNet.ScintillaControl)'.

   at PostfixCodeCompletion.PluginMain.HandleEvent(Object sender, NotifyEvent e, HandlingPriority priority)
```

This is the plugin throwing the exceptions: https://github.com/SlavaRa/fdplugin-postfix-code-completion

Next thin I tried to do is open the Tools menu, I pressed Alt-T, and saw "Alt+T was pressed. Waiting for a second key input..." in the statusbar. Is that expected behavior? If so, how to access the main menu?
I don't know if that is relevant, but I'm using [AutohideMenu](http://flashdevelop.org/community/viewtopic.php?f=4&t=8886) plug-in, with the main menu hidden and it's items duplicated in toolbar.

gene-pavlovsky:
P.S. If I disable extended shortcuts in the FD settings, and restart FD, Alt-T still doesn't work - the message in statusbar is instead "Shortcut not defined".

Neverbirth:
For the issue with the Postfix completion plugin the fix is really easy from both sides, but I cannot do it right now, maybe somebody beats me to it before. For the other issue I cannot say, but did you try to open the shortcuts form to check if by any chance you lost that shortcut and you have to reassign it.

wise0704:
First issue is because I changed the method's signature, without realising that it's public and used by plugins. I'll have to fix it. 

Second issue is because that's the modal shortcut provided by winforms, and even though I made sure I call `base.ProcessCmdKey()` when simple keys, somehow things aren't getting through according to your statement.
Try `Alt` and `T` separately. If I'm correct it should work. I'm not able to test it currently.

wise0704:
If I get a chance, I should also check for any other public method signature changes and fix them.

gene-pavlovsky:
`Alt` and `T` separately doesn't work, meaning the `Tools` menu doesn't open, but letter `t` does appear in the code editor. Just `Alt` doesn't activate the main menu. I've already disabled the AutohideMenu plug-in to make sure it's not related. Tried `Disable advanced shortcuts` both `true` and `false`.

Neverbirth:
@wise0704  The source code for the AutoHideMenu plugin is available, so you can take a look into what may be happening. Do you think you can come up with something soon?

gene-pavlovsky:
@Neverbirth to make it clear, I already disabled AutoHideMenu, even without it I can't access the main menu and it's items using standard shortcuts.

Neverbirth:
Another note: for me it works as expected. I'm looking into this menu problem with gene.

wise0704:
@Neverbirth So for you both `Alt+T` and `Alt` then `T` work, or just the latter?
I don't know why pressing `Alt` doesn't activate the main menu, because if you look at `MainForm.ProcessCmdKey()`, it calls base method and return when the key is Ctrl, Alt or Shift.

Neverbirth:
Both work fine for me, yes.

wise0704:
@Neverbirth If that's the case then I have no idea why it doesn't work for some others... Hope you can solve it

Neverbirth:
I checked remotely. First Alt, then T works for Gene, although doing it that way forces you to be careful, because if the File menu opens when you press Alt, FD restarts.

I have a hint on finding the problem, but I don't know if it may be tough or easy to fix it.

wise0704:
Then could you check if holding down Alt causes the toolstrip mnemonic underlines like _F_ile to show up?
If so, the `Form.ProcessCmdKey()` is not functioning properly on searching for toolstrips.
If not, `Form.ProcessCmdKey()` is not working properly. If this is the case, what are the specs?

gene-pavlovsky:
I've enabled always showing the mnemonic underlines somewhere in Windows settings a long time ago, can't recall where.
Found out another thing - I'm using AutoHotKey. When it's on, `Alt` then `T` doesn't work for me. It worked for Hector when he logged in to my PC via TeamViewer. If I disable AutoHotKey, I can use `Alt` then `T`. Of course, that is mighty inconvenient.
Either way, `Alt+T` doesn't work, even with AutoHotKey disabled, statusbar says "Alt+T was pressed, waiting for second key...".

wise0704:
I think that always show mnemonic setting thing could be the problem.
While I was navigating through the decompiled assemblies for WinForms keyboard shortcut mechanism, I did notice that when the native setting is enabled, things work in a quite different way. Not too sure however.

wise0704:
Maybe we should open this as a separate issue and discuss it there?

PR #1246, opened 2016-06-08
fixes #1215

### LINKED ISSUE #1215: Haxe: Syntax highlight: Import with alias, "as" keyword is not highlighted as keyword
Since Haxe 3.2.0 import with alias have two forms:

1. import some.Module in ModuleAlias;
2. import some.Module as ModuleAlias;

HaxeDevelop 5.1.1.1 now is not highlight "as" keyword as keyword in second form.

SlavaRa:


Gama11:
I think this is not added intentionally, similar to the `from` and `to` in abstracts, since it's a contextual keyword and it would be annoying to have it highlighted in other places where it's just used as an identifier.

SlavaRa:
hm... Yes, I had forgotten that it is then highlighted everywhere ((
I think that we must explain that the main issue

PR #1248, opened 2016-06-12
Fixes generate static variable and generate static function

SlavaRa:
BUG in release 5.1.1, code for example:
```
package {
	public class Main {
		public function Main() {
			Main.test|;
		}
	}
}
```
result after generating variable:
```
package {
	public class Main {
		private var test;//must be private static var test;
		public function Main() {
			Main.test;
		}
	}
}
```

BUG in developmnet version if setting generate explicit scope is true, code for example:
```
package;
public class Main {
		test|;
	}
}
```
result after generating variable:
```
package;
public class Main {
	static private var test;
		this.test;//must be Main.test
	}
}
```

### DIFF
--- a/External/Plugins/ASCompletion/Completion/ASGenerator.cs
+++ b/External/Plugins/ASCompletion/Completion/ASGenerator.cs
@@ -2082,14 +2082,15 @@ namespace ASCompletion.Completion
             // evaluate, if the variable (or constant) should be generated in other class
             ASResult varResult = ASComplete.GetExpressionType(sci, sci.WordEndPosition(sci.CurrentPos, true));
             var memberIsStatic = member != null && (member.Flags & FlagType.Static) > 0;
-            if (ASContext.CommonSettings.GenerateScope && !varResult.Context.Value.Contains(ASContext.Context.Features.dot))
+            var dot = ASContext.Context.Features.dot;
+            if (ASContext.CommonSettings.GenerateScope && !varResult.Context.Value.Contains(dot))
             {
                 position = sci.CurrentPos;
                 var start = sci.WordStartPosition(position, true);
                 var length = sci.MBSafeTextLength(contextToken);
                 sci.SetSel(start, start + length);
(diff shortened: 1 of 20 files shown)


SlavaRa:

External/Plugins/ASCompletion/Completion/ASGenerator.cs
```diff
@@ -2081,13 +2081,15 @@ private static void GenerateVariableJob(GeneratorJobType job, ScintillaControl s
...
+            var memberIsStatic = member != null && (member.Flags & FlagType.Static) > 0;
             if (ASContext.CommonSettings.GenerateScope && !varResult.Context.Value.Contains(ASContext.Context.Features.dot))
             {
                 position = sci.CurrentPos;
                 var start = sci.WordStartPosition(position, true);
                 var length = sci.MBSafeTextLength(contextToken);
                 sci.SetSel(start, start + length);
-                var text = "this." + contextToken;
+                var scope = memberIsStatic ? inClass.QualifiedName : "this";
+                var text = string.Format("{0}.{1}", scope, contextToken);
```
wise0704: I suggest never using `string.Format()` in completion/parser codes. They are extremely slow.
Use `scope + "." + contextToken` or `string.Concat(scope, ".", contextToken)` instead.

Also, wouldn't `ASContext.Context.Features.dot` be better than `"."`?

Neverbirth: Speed is not a concern in this case anyway. About the dot you are right.

SlavaRa: fixed ;)

SlavaRa:
@Neverbirth @elsassph, guys what about this PR?

Neverbirth:
I will try to give a final reply for the PR between today and tomorrow, but so far it's fine I think.

Meychi:
If @Neverbirth accepts its good for merge.

Neverbirth:
Good to go.

SlavaRa:
thanks

PR #1251, opened 2016-06-15
Select first entry in settings dialog.

wise0704:
This is not the best solution, as it doesn't ensure the first _category_ to be visible, which is visually the first row before the first item.

### DIFF
--- a/FlashDevelop/Controls/FilteredGrid.cs
+++ b/FlashDevelop/Controls/FilteredGrid.cs
@@ -208,6 +208,18 @@ namespace FlashDevelop.Controls
                     }
                 }
             }
+            m_PropertyDescriptors.Sort(CompareDescriptors);
+        }
+
+        /// <summary>
+        /// Compare two property descriptors.
+        /// </summary>
+        private static int CompareDescriptors(PropertyDescriptor a, PropertyDescriptor b)
+        {
+            if (a == null) return b == null ? 0 : -1;
+            else if (b == null) return 1;
+            int value = string.CompareOrdinal(a.Category, b.Category);
+            return value == 0 ? string.CompareOrdinal(a.DisplayName, b.DisplayName) : value;
         }
 
         /// <summary>
--- a/FlashDevelop/Controls/FilteredGrid.cs
+++ b/FlashDevelop/Controls/FilteredGrid.cs
@@ -218,8 +218,8 @@ namespace FlashDevelop.Controls
         {
             if (a == null) return b == null ? 0 : -1;
             else if (b == null) return 1;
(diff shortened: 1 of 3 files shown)


wise0704:
Probably the smallest PR I've ever made.

Neverbirth:
Maybe it's a stupid question... but... Is this really an improvement? :S. From my side I don't see it like an UX need, and it may even slow down filtering, although I guess we don't have that many items to worry about.

wise0704:
When the locale is in English, for current settings valies it doesn't seem to make much difference. But when you choose some other locale, since the category names are different and sorted in different order, it does get quite annoying when the default selection is at the bottom.
Considering there is a `Refresh()` call in the filter, I don't think selection change will have much impact. :)

wise0704:
But there could be a way to achieve this with better performance...

elsassph:
I'd prefer  having the filter focused by default...

wise0704:
This was a careless modification I see after all. Didn't take into account that this method is called when the filter text changes, and there also could be null references.

I tried to find better solutions, playing around with `FilteredGrid` and `m_PropertyDescriptors`, and now it works well with en_US, but for some reason the internal item selection algorithm (I can't find where it is placed) selects the first item sorted with the English category name (for e.g.):
-en_US- selection:
image
Although there is "A**p**ply File Extension", "A**u**to Filter List" is selected first.
-ko_KR- selection:
image
Is the same as en_US, although it should be:
image

But for some reason, ko_KR only works incorrectly like that for "FlashDevelop" settings. For other plugins it selects the first item correctly.

Meychi:
I don't know why it doesn't work but this selection is defined in FlashDevelop/Settings/Properties.cs:

[DefaultProperty("AutoFilterList")]

wise0704:
@Meychi I think that's exactly why it doesn't work... If I delete it everything works correctly.

FlashDevelop/Controls/FilteredGrid.cs
```diff
@@ -208,6 +208,18 @@ private void RefreshProperties()
...
+
+        /// <summary>
+        /// Compare two property descriptors.
+        /// </summary>
+        private static int CompareDescriptors(PropertyDescriptor a, PropertyDescriptor b)
+        {
+            if (a == null) return b == null ? 0 : -1;
+            else if (b == null) return 1;
+            int value = string.Compare(a.Category, b.Category);
+            return value == 0 ? string.Compare(a.DisplayName, b.DisplayName) : value;
```
wise0704: Better respect the current culture, because the strings are localised...

PR #1253, opened 2016-06-19
Scroll width tracking

wise0704:
http://www.flashdevelop.org/community/viewtopic.php?f=5&t=12589

Setting `ScrollWidth` to `0` should enable `ScrollWidthTracking` too.

### DIFF
--- a/FlashDevelop/Managers/ScintillaManager.cs
+++ b/FlashDevelop/Managers/ScintillaManager.cs
@@ -253,7 +253,8 @@ namespace FlashDevelop.Managers
                 sci.IsTabIndents = settings.TabIndents;
                 sci.IsUseTabs = settings.UseTabs;
                 sci.IsViewEOL = settings.ViewEOL;
-                sci.ScrollWidth = settings.ScrollWidth;
+                sci.ScrollWidth = Math.Max(settings.ScrollWidth, 1);
+                sci.ScrollWidthTracking = settings.ScrollWidth == 0 || settings.ScrollWidth == 3000;
                 sci.TabWidth = settings.TabWidth;
                 sci.ViewWS = Convert.ToInt32(settings.ViewWhitespace);
                 sci.WrapMode = Convert.ToInt32(settings.WrapText);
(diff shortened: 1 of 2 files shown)


FlashDevelop/Managers/ScintillaManager.cs
```diff
@@ -253,7 +253,8 @@ public static void ApplySciSettings(ScintillaControl sci, Boolean hardUpdate)
                 sci.IsTabIndents = settings.TabIndents;
                 sci.IsUseTabs = settings.UseTabs;
                 sci.IsViewEOL = settings.ViewEOL;
-                sci.ScrollWidth = settings.ScrollWidth;
+                sci.ScrollWidth = Math.Max(settings.ScrollWidth, 1);
```
Meychi: Is 1 minimum width?

FlashDevelop/Settings/Accessors.cs
```diff
@@ -294,7 +294,11 @@ public Int32 CaretWidth
         public Int32 ScrollWidth
         {
             get { return this.scrollWidth; }
-            set { this.scrollWidth = value; }
+            set
+            {
+                if (value < 0) throw new ArgumentOutOfRangeException();
```
Meychi: I don't think throwing error here is good. Maybe just reset it to 0 if its negative.

Neverbirth: I think throwing an error is the standard thing to do in this case.

FlashDevelop/Managers/ScintillaManager.cs
wise0704: Yep, setting it to 0 sets it to the scintilla default (which is I think either 2000 or 5000).

FlashDevelop/Settings/Accessors.cs
wise0704: what should i do...

Meychi:
Well match it to other similar settings. Try put an invalid value and check if it gives an error or not.

wise0704:
@Meychi 
# ???????????
image

wise0704:
That I'm sure isn't expected...

wise0704:
`PropertyGrid` by default does this however:
image

PR #1259, opened 2016-06-21
XML Ampersand Problem

wise0704:
This is what happens if I try to edit comments inside XML... :(


wise0704:
I believe the first two commits were already merged??? Git plz?

PR #1261, opened 2016-06-21
Notify Restart Required on Settings Change

wise0704:
I think it's good to let users know which settings take effect only after a restart. 
Really simple syntax for developers: just add a `[RequiresRestart]` attribute.


wise0704:
One minor problem with this is that, when you install something with AppMan and it calls `RestartRequired()`, opening and closing the settings dialog will hide the `restartButton`. But this is really minor. They can always click File > Restart.

Meychi:
Has conflicts now.

I think its better just to just request restart and not cancel it if settings are restored.

wise0704:
Are you sure? I thought that would be annoying...

wise0704:
Well, I guess it shouldn't be canceled if the dialog was closed then reopened, then the value was restored. 
So remove this: https://github.com/fdorg/flashdevelop/pull/1261/files#diff-90b75daa5189dd008a8d9c44e85a41e2R599 and do some checks to see if the required flag was already set.

wise0704:
@Meychi Can you review please?

wise0704:
Which other settings require restart on modification?

Neverbirth:
I'm OK with this one.

PR #1264, opened 2016-06-21
Place caret in the centre when jumping to locations

### LINKED ISSUE #1163: Add option to center screen on search result
When running a search from the Find bar (Ctrl-F), if the next result is off the edge of the screen, the screen will scroll until the result appears on the top or bottom edge of the screen and then stop.

Feature request: Add a "center in view" checkbox to the Find bar that, when checked, will cause the screen to scroll until the search result is vertically centered (if possible, obviously this wouldn't work on the first few or last few lines of a file).

### LINKED ISSUE #1164: Results panel can obscure target code
When clicking a result in the Results panel, the code window will scroll to the corresponding line. If the Results panel is docked to the bottom of the screen and not pinned (i.e. it collapses when you click away from it) the target line may appear _underneath _the Results panel and not be visible until the user clicks outside of the panel to close it.

IIRC this bug seems to be new to 5.1.0.3

### LINKED ISSUE #1206: Back button during F4 navigation
Sometimes when investigating how a class works, I end up following a long trail by pressing F4 with the keyboard cursor on various keywords.  I've had it happen a few times, especially within a file where the scroll position has no history, when I wished that there was some way to go back to the previous position, where I was before I pressed F4.  (Between files it works a bit better, if I can remember the last file, which is usually the neighbouring tab, I just switch to it and it's in the right position.  However, if F4 history could be tracked between tabs, that would be even more useful.)

wise0704:
~~stole @Neverbirth's idea~~

Closes #1163 
Closes #1164
Closes #1206 

Pretty much, when navigating with goto-like commands (Go to definition, go to result, etc.) the caret should be centred, instead of being on the border.

I know that the new setting `Keep Caret Centered` is added, but that will also disable the new caret policy, which is keeping the caret on the border when it goes out of the screen, and instead jump to the centre. 
Scintilla supports this by default, so we should be using this!

I did not take much time when replacing some calls to `EnsureVisible()` to `EnsureVisibleEnforcePolicy()`. Please review the replacements (and also the ones that have not been replaced).

### DIFF
--- a/External/Plugins/AS3Context/Controls/ProfilerObjectsView.cs
+++ b/External/Plugins/AS3Context/Controls/ProfilerObjectsView.cs
@@ -51,7 +51,7 @@ namespace AS3Context.Controls
                         ScintillaControl sci = PluginBase.MainForm.CurrentDocument.SciControl;
                         int pos = sci.PositionFromLine(lineToOpen);
                         sci.SetSel(pos, pos);
-                        sci.EnsureVisible(lineToOpen);
+                        sci.EnsureVisibleEnforcePolicy(lineToOpen);
                     }
                 }
                 fileToOpen = null;
--- a/External/Plugins/AS3Context/Controls/ProfilerObjectsView.cs
+++ b/External/Plugins/AS3Context/Controls/ProfilerObjectsView.cs
@@ -51,7 +51,7 @@ namespace AS3Context.Controls
                         ScintillaControl sci = PluginBase.MainForm.CurrentDocument.SciControl;
                         int pos = sci.PositionFromLine(lineToOpen);
                         sci.SetSel(pos, pos);
-                        sci.EnsureVisibleEnforcePolicy(lineToOpen);
+                        sci.EnsureVisible(lineToOpen);
                     }
                 }
                 fileToOpen = null;
--- a/External/Plugins/ASCompletion/Completion/ASComplete.cs
+++ b/External/Plugins/ASCompletion/Completion/ASComplete.cs
@@ -665,7 +665,7 @@ namespace ASCompletion.Completion
 
                 if (!found)
                 {
-                    sci.EnsureVisibleEnforcePolicy(line);
+                    sci.EnsureVisible(line);
                     int linePos = sci.PositionFromLine(line);
                     sci.SetSel(linePos, linePos);
                 }
--- a/External/Plugins/ASCompletion/Completion/ASComplete.cs
+++ b/External/Plugins/ASCompletion/Completion/ASComplete.cs
@@ -656,7 +656,7 @@ namespace ASCompletion.Completion
                         if (m.Success)
                         {
                             int position = sci.PositionFromLine(i) + sci.MBSafeTextLength(text.Substring(0, m.Groups["name"].Index));
-                            sci.EnsureVisibleEnforcePolicy(sci.LineFromPosition(position));
+                            sci.EnsureVisible(sci.LineFromPosition(position));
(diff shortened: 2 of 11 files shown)


Neverbirth:
I don't think this was my idea at all.

IMHO, why not directly change EnsureVisible to behave like it's wanted? After all, the behaviour was until short time ago that one, so every old plugin out there would expect it I'd say.

wise0704:
True. I was just following the scintilla's default behavior, but tbh `EnsureVisible()` should just call `EnsureVisibleEnforcePolicy()`.

wise0704:
On a second thought, I think some of those calls to `EnsureVisible()` could also mean to follow the caret policy instead of the visible policy...

Meychi:
Please check those before merge.

Neverbirth:
Which ones do you think should follow the caret policy? I don't know, until a couple of months ago or so the caret policy was to use the center or something like that, so I don't know if people will really notice the difference or will complain.

wise0704:
On a third thought, all calls to `EnsureVisible()` would be navigation through commands, instead of being caused by the caret moving from users typing or pressing arrow keys.

The former should centre the caret (as set with `SetVisiblePolicy()`), and latter should follow the `KeepCaretCentered` setting value (as set with `SetX/YCaretPolicy()`). 
So yes, call calls to `EnsureVisible()` should actually call `EnsureVisibleEnforcePolicy()`.

elsassph:
Good for me, EnsureVisible should definitely centre the line.

wise0704:
I'll double check all references to EnsureVisible as soon as possible and confirm.

wise0704:
Found one problem - conflicts between `Strict` and `Jumps`:

When restoring scroll position (e.g. file session, etc.) the scroll position is set to desired, and the caret is already visible on screen. In this case, the visible policy should be `Jumps`, which does not force the caret to be in centre if already visible. _(Or should call `SCI_ENSUREVISIBLE`, if `KeepCaretCentered` is true; or just not call `EnsureVisible` at all)._

When navigating with commands (e.g. go to definition, etc.) the destination could be visible on screen, but it should still be centred on screen. In this case the visible policy should be `Strict`. 

Opinions? If above conventions are to be kept, things should work the way I wrote in italics (or there could be better ways).

elsassph:
Right, maybe we still need two methods so there would be an explicit centering of the line when jumping to definitions (which should include clicking on results panel).

Neverbirth:
I'll try to check this one ASAP if nobody beats me to it...

elsassph:
Cool

Neverbirth:
I really think EnsureVisible should be the one jumping to the center of the screen... I know it goes against Scintilla own definitions, but as I said, this was the old behavior, and any existing plugin will expect this. 

As an example, when changing the stack frame while debugging I'd say the line should be displayed at the center. By searching in the code, it seems the places that should be using the new policy are by far less than the places that are better seen centered.

elsassph:
You may be right @Neverbirth. So in fact we'd want to know where we would not want to apply the policy.

wise0704:
I went through all the references of `EnsureVisible()` manually and checked them all.
The unchanged ones are where we don't want to apply the policy.
Even though jumping to the centre was the old behaviour, that was when the caret policy also jumped to the centre when the caret went off bounds.

Btw @elsassph your opinion changes on every comment :P

Neverbirth:
I don't know, as I said, for me things like the current line where the debugger is stopped, and maybe also when restoring the state of a file look better centered. The first one also happens to be the same in other IDEs.

At any rate, things are better now.

PR #1265, opened 2016-06-21
Save configurations on dialog close

### LINKED ISSUE #1195: Save FD config after editing, or add a button for saving config
Usually my FD settings are long. Rarely, but FD does crash. When it does, it loses any changes I made to keyboard shortcuts, configured SDKs etc. It would be not too resource intensive to save config after closing the settings / keyboard shortcuts dialogs? Maybe after some delay. Or at least periodically (e.g. once in 5 minutes check if there are any unsaved settings, save them).

wise0704:
Closes #1195


wise0704:
Should the buttons have the label changed to "Save" instead of "Close"?

Meychi:
No i think its fine.

PR #1266, opened 2016-06-21
Fixes error on save for hidden files.

### LINKED ISSUE #1166: FD can't save a file if it has the "hidden" attribute set
FlashDevelop can't save an edited file if it has the "hidden" attribute set. (Notepad can save the same file). Not a permission problem: `attrib -h file` and FD can save it again.

```
Access to the path 'C:\project\apps\FXCalc\fonts\.FontLibrary-config.xml' is denied.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   ...
```

wise0704:
Closes #1166 

Currently 10/13 PR's are mine lol.

### DIFF
--- a/PluginCore/PluginCore/Helpers/FileHelper.cs
+++ b/PluginCore/PluginCore/Helpers/FileHelper.cs
@@ -81,7 +81,8 @@ namespace PluginCore.Helpers
         {
             Boolean useSkipBomWriter = (encoding == Encoding.UTF8 && !saveBOM);
             if (encoding == Encoding.UTF7) encoding = new UTF7EncodingFixed();
-            using (StreamWriter sw = useSkipBomWriter ? new StreamWriter(file, false) : new StreamWriter(file, false, encoding))
+            using (FileStream fs = new FileStream(file, FileMode.Open))
+            using (StreamWriter sw = useSkipBomWriter ? new StreamWriter(fs) : new StreamWriter(fs, encoding))
             {
                 sw.Write(text);
                 sw.Close();
--- a/PluginCore/PluginCore/Helpers/FileHelper.cs
+++ b/PluginCore/PluginCore/Helpers/FileHelper.cs
@@ -81,7 +81,7 @@ namespace PluginCore.Helpers
         {
             Boolean useSkipBomWriter = (encoding == Encoding.UTF8 && !saveBOM);
             if (encoding == Encoding.UTF7) encoding = new UTF7EncodingFixed();
-            using (FileStream fs = new FileStream(file, FileMode.Open))
+            using (FileStream fs = new FileStream(file, FileMode.OpenOrCreate))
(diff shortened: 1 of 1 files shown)


PluginCore/PluginCore/Helpers/FileHelper.cs
```diff
@@ -81,7 +81,8 @@ public static void WriteFile(String file, String text, Encoding encoding, Boolea
         {
             Boolean useSkipBomWriter = (encoding == Encoding.UTF8 && !saveBOM);
             if (encoding == Encoding.UTF7) encoding = new UTF7EncodingFixed();
-            using (StreamWriter sw = useSkipBomWriter ? new StreamWriter(file, false) : new StreamWriter(file, false, encoding))
+            using (FileStream fs = new FileStream(file, FileMode.Open))
```
Neverbirth: Shouldn't it be FileMode.OpenOrCreate?

wise0704: Yes, since `StreamWriter` with string did create a new file if the file doesn't exist.
> **StreamWriter Constructor (String, Boolean)**
>.NET Framework 3.5 Other Versions 
>Initializes a new instance of the StreamWriter class for the specified file on the specified path, using the default encoding and buffer size. If the file exists, it can be either overwritten or appended to. If the file does not exist, this constructor creates a new file.
Will fix.

wise0704:
The problem was that `StreamWriter` does a file existence check using `File.Exists()`, and that method returns false on hidden files. Thus then it tries to create an existing file, causing the exception.

gene-pavlovsky:
I'm suspecting that this fix causes #1270 !
To reproduce, open any file, delete some of it's contents (reduce it's size), and save. The new contents are saved on top of the old contents, but the file size is not reduced. Whatever was at the end of the file, remains there. It doesn't show up in FD until file is closed&reopened.

PluginCore/PluginCore/Helpers/FileHelper.cs
```diff
@@ -81,7 +81,8 @@ public static void WriteFile(String file, String text, Encoding encoding, Boolea
         {
             Boolean useSkipBomWriter = (encoding == Encoding.UTF8 && !saveBOM);
             if (encoding == Encoding.UTF7) encoding = new UTF7EncodingFixed();
-            using (StreamWriter sw = useSkipBomWriter ? new StreamWriter(file, false) : new StreamWriter(file, false, encoding))
+            using (FileStream fs = new FileStream(file, FileMode.OpenOrCreate))
```
wise0704: Our problem right now is that the file is not truncated if it already exists. `FileMode.OpenOrCreate` works like `Open` if it exists, and `CreateNew` if not, which is not what we want. `FileMode.Create` (which is what `StreamWriter` uses internally) does `Truncate` if exists, and `CreateNew` if not. However, `Create` seems to not work for hidden files.

PR #1267, opened 2016-06-21
Restore FD position regarding multiple monitors.

### LINKED ISSUE #856: Window position (start on display #2) not remembered in my display configuration
FlashDevelop can't remember the last used monitor/position. My display settings are like this:
Settings
Display 1 is my laptop, and Display 2 is an external wall-mounted monitor, above my laptop. When I start FD, it always appears on laptop's screen, but I want it to start on the other screen. I tried un-maximizing FD window, moving it to screen 2, then restarting FD, then maximizing the window again. Most apps with "remember window size&position" feature usually remember which screen to start on after this procedure, but not in my case. Maybe it's related to my particular screen arrangement. Since screen 2 is above and to the left from screen 1, when FD is on screen 2, it's actual window position is -1920, -1080 (FlashDevelop Settings > State). Maybe the position doesn't get remembered or restored because negative coordinates are not handled correctly? For now as a work-around I have to use an AHK script which triggers after I launch FlashDevelop with a hotkey
```autohotkey
      WinWait, ahk_exe FlashDevelop.exe
      if WinExist("ahk_exe FlashDevelop.exe") {
         WinRestore
         WinMove -1920, -1080
         WinMaximize
         WinActivate
      }
```

wise0704:
Closes #856 

11 out of 14 now :)

### DIFF
--- a/FlashDevelop/MainForm.cs
+++ b/FlashDevelop/MainForm.cs
@@ -904,20 +904,18 @@ namespace FlashDevelop
         {
             this.WindowState = this.appSettings.WindowState;
             Rectangle bounds = new Rectangle(this.appSettings.WindowPosition, this.appSettings.WindowSize);
+            bounds.Inflate(-4, -25);
             Boolean validPosition = false;
             foreach (Screen screen in Screen.AllScreens)
             {
                 if (screen.Bounds.IntersectsWith(bounds))
                 {
+                    this.Location = this.appSettings.WindowPosition;
                     validPosition = true;
                     break;
                 }
             }
-            if (validPosition)
-            {
-                this.Location = this.appSettings.WindowPosition;
-            }
-            else this.Location = new Point(0, 0);
+            if (!validPosition) this.Location = new Point(0, 0);
             // Continue/perform layout!
             this.ResumeLayout(false);
             this.PerformLayout();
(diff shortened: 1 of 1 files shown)


wise0704:
I'm gonna go to bed now... lol
Do I deserve a cookie??

Meychi:
A gigantic one, thanks. :) I will check all these later today.

Neverbirth:
The change seems to be valid, but I don't know if it's missing something else.

FlashDevelop/MainForm.cs
```diff
@@ -903,10 +903,21 @@ public void CheckForUpdates()
         public void InitializeWindow()
         {
             this.WindowState = this.appSettings.WindowState;
-            Point position = new Point(this.appSettings.WindowPosition.X, this.appSettings.WindowPosition.Y);
-            if (position.X < -4) position.X = 0;
-            if (position.Y < -25) position.Y = 0;
-            this.Location = position;
+            Rectangle bounds = new Rectangle(this.appSettings.WindowPosition, this.appSettings.WindowSize);
+            Boolean validPosition = false;
```
wise0704: Maybe adding `bounds.Inflate(-4, -25)` here could be better, since previously it was like that.

Meychi:
Soon as someone can test this, we can merge it.

wise0704:
Just FYI: the code fragment is actually provided by Microsoft. Should be working.

Meychi:
Ok fine for me then.

PR #1274, opened 2016-06-30
Test improvements

Neverbirth:
The changes I did near 4 months ago!, so I'd say you already checked back in the day.

However this is a new branch, so I want to run some new tests before merging, like creating again a CI build. I also didn't test build.cmd on local.


Neverbirth:
It seems everything is working fine. Although there is that fixed line ending style thing that should be added later on.

@SlavaRa Some of your CodeRefactor tests are not working even in the development branch. I didn't have the time to check why.

SlavaRa:
i look today

Neverbirth:
Some ASGeneratorTests.GenerateJob.GenerateExtractVariable tests fail too.

SlavaRa:
Tomorrow I will explain why and ask the Council what to do, okay?

Neverbirth:
If nobody sees a problem before, I'll merge between today night and tomorrow.

PluginCore/PluginCore/Helpers/PathHelper.cs
```diff
@@ -19,7 +20,7 @@ public static String BaseDir
         {
             get
             {
-                if (PluginBase.MainForm.StandaloneMode) return AppDir;
+                if (PluginBase.MainForm.StandaloneMode) return Path.GetDirectoryName(GetAssemblyPath(Assembly.GetExecutingAssembly()));
```
wise0704: Doesn't this make `BaseDir` possibly return a different path from `AppDir` in standalone mode?
Shouldn't `AppDir` be changed accordingly, and `BaseDir` still return `AppDir`?

Neverbirth: That's a bit subjective... It's normal that the directory of the application is where the executable file resides, while the base directory is not necessarily the same as the one where the application resides.

The problem here is that we don't have full control or who is running the tests and where from, so the application making use of the FD libraries is not in the same folder as the base directory where all of the needed files are.

wise0704: I understand the changes made to the path.
But it's just that the `$(BaseDir)` is defined as: `$(UserAppDir)` in normal cases, `$(AppDir)` in standalone mode.
The tests may not only use the base dir (although I think you've only changed base dir since it's the only one related to output path?) and in that case I think unsynchronised values could cause some unexpected results in weird places.

Neverbirth: If I fully understand all of what you say, I don't agree. If some part of code tries to load some file that should be relative to the base directory using the application directory, I would say it's wrong. As I say, I see AppDir and BaseDir as two different things, and there is no reason for them to be always the same thing in standalone mode.

I only changed the base dir not because it's related to ouput path, but because the snippet and template directories use the BaseDir as the root folder.

Anyway, take into account that in normal cases PluginCore and FlashDevelop.exe will always be in the same directory, so there won't be any noticeable change for any user.

Meychi: There should be no changes in this area even if it should work. The implementation is more complex and possibly error prone in complex setups, for example with crossover.

Neverbirth: Actually, it's code inside the .NET framework itself to handle correctly some different complex cases like shadow copies, network drives, etc, but it's not exposed for this case. We "could" use the .Location property without any kind of processing, but then not all of the cases would be handled, although I guess we are not going to use PluginCore in something like a web server?

Anyway, what is your proposal?

Meychi: Keep now as it is. These changes needs to be done in a major release and tested properly.

Neverbirth: OMG, then I hope that major release is going to happen anytime soon.

wise0704: If this change only had to be done to fix problems related to testing, would it be possible to make it compile this way only for testing?

wise0704: What does `GetAssemblyPath()` do exactly? Is it related to just shadow copying and stuff? Or could it return a completely different path than the application directory?

Neverbirth: In this case it returns the original path of PluginCore.dll (avoiding the path used when shadow copying), so yeah, it can return a different path than the application directory. In FD they are in the same folder, but if another application uses PluginCore.dll, it won't be the folder where the application resides, but then again, that's why there is an AppDir, this is BaseDir.

I don't see the point of compiling this code only for testing.

wise0704: So in FD (in Standalone mode) `BaseDir` and `AppDir` will still return the **exact** same paths, right?
If so, I understood your point. It was just the difference in understanding the definition of `BaseDir` - I knew it as `Standalone ? AppDir : UserAppDir` (it's also documented like this).
Your definition was `Standalone ? (PluginCore's directory) : UserAppDir`.

But since the cases you're describing has to deal with PluginCore's directory, and PluginCore.dll and FlashDevelop.exe are placed in the same folder in FD, I think it's okay.

Although I would prefer making a new property that returns `Path.GetDirectoryName(GetAssemblyPath(Assembly.GetExecutingAssembly()))` with `GetAssemblyPath()` inlined in the property. This property could be needed in other scenarios too.

By the way, why is the last line of `GetAssemblyPath()` returning `Location` instead of `CodeBase`? And can't the whole thing be replaced with `new Uri(Assembly.GetExecutingAssembly().CodeBase).LocalPath`?

Neverbirth: What would be the benefit of adding it as another new property? Which other scenarios are you thinking of? I cannot think of one right now, because in not standalone mode there would be most probably write rights problems and this path is less likely to be used.

Why is it returning Location? I don't know right now, I checked my notes just to be fully sure, the method is taken from the NUnit code base, which in turn was taken from other sources, but it's tested inside the NUnit test suite itself, so I think it's safe to use.

Your approach can be used for FD, but it's not compatible with all possible .NET use cases, which I think it's a good idea.  A more complete one would be:

    Uri codeBase = new Uri(cb);
    return codeBase.IsFile ? codeBase.LocalPath : codeBase.ToString();

Neverbirth:
Will solve the merge conflicts later today. Now that FD 5.2 is out we can merge this and enable CI and PR support for unit tests.

Neverbirth:
Fixed conflicts.

PR #1290, opened 2016-07-23
Do not close snippet on backward selection.

### LINKED ISSUE #1277: Snippet close on backward selection
http://www.flashdevelop.org/community/viewtopic.php?f=6&t=12643

wise0704:
Closes #1277

### DIFF
--- a/PluginCore/PluginCore/Controls/UITools.cs
+++ b/PluginCore/PluginCore/Controls/UITools.cs
@@ -303,7 +303,7 @@ namespace PluginCore.Controls
             }
             if (sci != null && sci.IsFocus)
             {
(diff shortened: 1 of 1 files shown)


Neverbirth:
Could you explain a bit please?

wise0704:
@Neverbirth On a backward selection of a keyword, `position` is incorrectly set to the start of the selection instead of the end, causing both `CheckPosition(position)` to return `false`.
That causes `Hide()` to be called on `callTip`, `CompletionList` and `simpleTip`, causing snippet list to glitch and immediately close.

Neverbirth:
OK, my "fear" was if this would modify some unwanted behavior, but I checked and everything was fine.

PR #1291, opened 2016-07-23
Case sensitive renaming in Project Manager

### LINKED ISSUE #1278: Case-sensitive renaming of files in Project Panel
http://www.flashdevelop.org/community/viewtopic.php?f=6&t=12650

### LINKED ISSUE #1288: Renaming class: avoid replace existing file warning if only case is changed (on Windows).
Got a project with interface files called `iSomeInterface`. Renaming it to `ISomeInterface` and getting a confirm dialog that file already exists (since on Windows file system is case-insensitive), "Would you like to replace?". Basically, on Windows, this dialog should only appear if `oldName.toLowerCase() != newName.toLowerCase`. Considering FD would (hopefully) be available for Linux one day, this should only apply on a case-insensitive filesystem.

wise0704:
Closes #1278 
Closes #1288


External/Plugins/CodeRefactor/Commands/RenameFile.cs
```diff
@@ -48,17 +48,19 @@ protected override void ExecutionImplementation()
...
             if (Path.GetFileName(oldPath).Equals(newPath, StringComparison.OrdinalIgnoreCase))
             {
                 // name casing changed
                 string tmpPath = oldPath + "$renaming$";
                 File.Move(oldPath, tmpPath);
-                oldPath = tmpPath;
+                File.Move(tmpPath, newPath);
+                DocumentManager.MoveDocuments(oldPath, newPath);
             }
-            if (!Path.IsPathRooted(newPath)) newPath = Path.Combine(Path.GetDirectoryName(oldPath), newPath);
```
Neverbirth: Why did you remove this one? I should check my notes... but IIRC it was this exact one that back in the day I also thought it was unneeded and didn't find any valid reason for it but left it just in case...

wise0704: It's moved to [line 52](https://github.com/fdorg/flashdevelop/pull/1291/files/fa9a81c728ffdc8cac598b5fbb5a9997cd111fd3#diff-b28e14b0a502fd942600df3ec02eb1ebR52), without the `if` statement, since `Path.Combine()` already checks if `path2` is rooted.

External/Plugins/ProjectManager/Actions/FileActions.cs
```diff
@@ -445,34 +445,29 @@ public bool Rename(string oldPath, string newName)
 
                 if (isDirectory)
                 {
-                    // this is required for renaming directories, don't ask me why
-                    string oldPathFixed = (oldPath.EndsWith('\\')) ? oldPath : oldPath + "\\";
-                    string newPathFixed = (newPath.EndsWith('\\')) ? newPath : newPath + "\\";
-                    if (oldPathFixed.Equals(newPathFixed, StringComparison.OrdinalIgnoreCase))
```
wise0704: I removed those lines because it worked without them. Was there a specific case when this failed without the lines?

External/Plugins/CodeRefactor/Commands/RenameFile.cs
```diff
@@ -48,17 +48,19 @@ protected override void ExecutionImplementation()
                     return;
                 }
             }
+
+            newPath = Path.Combine(Path.GetDirectoryName(oldPath), newPath);
+
             // refactor failed or was refused
             if (Path.GetFileName(oldPath).Equals(newPath, StringComparison.OrdinalIgnoreCase))
```
wise0704: Oops I think I need to remove `Path.GetFileName()` on `oldPath`...

External/Plugins/ProjectManager/Actions/FileActions.cs
Neverbirth: IIRC back in the day it was about normalizing the paths, because they could came from different sources and in different ways, so yeah, it could fail.

External/Plugins/ProjectManager/Actions/FileActions.cs
```diff
@@ -446,33 +446,31 @@ public bool Rename(string oldPath, string newName)
...
                 {
-                    string oldName = Path.GetFileName(oldPath);
-                    if (oldName.Equals(newName, StringComparison.OrdinalIgnoreCase))
+                    if (oldPath.Equals(newPath, StringComparison.OrdinalIgnoreCase))
                     {
                         // name casing changed
                         string tmpPath = newPath + "$renaming$";
                         File.Move(oldPath, tmpPath);
-                        oldPath = tmpPath;
+                        File.Move(tmpPath, newPath);
```
Neverbirth: I see a little gotcha here. Since you are calling directly File.Move, in case the file is the main target one, it doesn't get updated in the project. Of course, in Windows it's OK, but it wouldn't be in case sensitive scenarios.

wise0704: But if that's the case, the `else` block is also not proper, as `FileHelper.ForceMove()` just does `File.Move()`.
Aren't `CodeRefactor.Commands.Rename` and `RenameFile` called in the cases where documents are renamed? I think this bit of code is like generic part, like for miscellaneous files.
Sadly I don't have a computer (gg rip) to run through method references so I can't confirm now.

External/Plugins/ProjectManager/Actions/FileActions.cs
wise0704: It used to be because of `FileHelper.ForceMoveDirectory()`, [this line](https://github.com/fdorg/flashdevelop/pull/1291/files#diff-babcd5980fed43a8a6c7f6fccae1d1afR228).

However since the method now seems to add +1 to the length variable if the backslash isn't present, so it's not an issue anymore.

External/Plugins/CodeRefactor/Commands/Move.cs
```diff
@@ -473,37 +456,15 @@ private void MoveRefactoredFiles()
...
-                            if (project.IsDocumentClass(file))
-                            {
-                                newDocumentClass = file.Replace(oldPath, newPath);
-                                break;
-                            }
-                        }
-                        if (newDocumentClass != null) break;
-                    }
+                    string newDocumentClass;
+                    if (RefactoringHelper.TryGetDocumentClass(oldPath, out newDocumentClass)) newDocumentClass.Replace(oldPath, newPath);
```
wise0704: I've used `.Replace(oldPath, newPath)`, just because the previous was `file.Replace(oldPath, newPath)`, but I think that's not quite safe.

Shouldn't this really be 
```c#
Path.Combine(newPath, newDocumentClass.Substring(oldPath.Length))
```
Assuming (or assuring beforehand) that `oldPath` is rooted?

wise0704:
Note: `FileHelper.MoveFile( , , true)` and `FileHelper.MoveDirectory( , , true)` should be used instead of `ForceMove` or `ForceMoveDirectory` to take case-only rename into account.

Meychi:
@Neverbirth, is this ok?

Meychi:
@SlavaRa: Is this good?

wise0704:
seems too outdated now

PR #1293, opened 2016-08-02
[FDBuild] fix project building in Linux.

Mixaill:
Fixes .as2proj / .as3proj projects build in Linux. Works in Windows as well.

### DIFF
--- a/External/Tools/FDBuild/Program.cs
+++ b/External/Tools/FDBuild/Program.cs
@@ -44,8 +44,10 @@ namespace FDBuild
                 catch { }
 
             string swfmillPath = options.SwfmillPath ?? Path.Combine(toolsDir, "swfmill");
-            if (File.Exists(Path.Combine(swfmillPath, "swfmill.exe")))
+            if (File.Exists(Path.Combine(swfmillPath, "swfmill.exe"))) //Windows binary
                 SwfmillLibraryBuilder.ExecutablePath = Path.Combine(swfmillPath, "swfmill.exe");
+            else if (File.Exists(Path.Combine(swfmillPath, "swfmill"))) //Linux binary
+                SwfmillLibraryBuilder.ExecutablePath = Path.Combine(swfmillPath, "swfmill");
             else
                 SwfmillLibraryBuilder.ExecutablePath = "swfmill"; // hope you have it in your environment path!
 
--- a/External/Tools/FDBuild/Program.cs
+++ b/External/Tools/FDBuild/Program.cs
@@ -47,7 +47,7 @@ namespace FDBuild
             if (File.Exists(Path.Combine(swfmillPath, "swfmill.exe")))
                 SwfmillLibraryBuilder.ExecutablePath = Path.Combine(swfmillPath, "swfmill.exe");
             else
-                SwfmillLibraryBuilder.ExecutablePath = "swfmill.exe"; // hope you have it in your environment path!
+                SwfmillLibraryBuilder.ExecutablePath = "swfmill"; // hope you have it in your environment path!
 
             // figure user settings PlatformData
             string platformsFile = Path.Combine("Settings", "Platforms");
--- a/External/Tools/FDBuild/Program.cs
+++ b/External/Tools/FDBuild/Program.cs
@@ -1,11 +1,11 @@
 using System;
-using System.Reflection;
 using System.IO;
 using ProjectManager.Projects;
 using ProjectManager.Building;
 using ProjectManager.Building.AS3;
 using FDBuild.Building;
 using System.Text;
+using PluginCore.Helpers;
 
 
 namespace FDBuild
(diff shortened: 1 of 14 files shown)


PluginCore/PluginCore/Helpers/JvmConfigHelper.cs
```diff
@@ -79,11 +79,11 @@ public static string GetJavaEXE(Dictionary<string, string> jvmConfig)
         }
         public static string GetJavaEXE(Dictionary<string, string> jvmConfig, string flexSdkPath)
         {
-            string defaultExe = "java.exe";
+            string defaultExe = "java";
             string home = GetJavaHome(jvmConfig, flexSdkPath);
             if (!String.IsNullOrEmpty(home) && !home.StartsWith("%", StringComparison.Ordinal))
             {
-                return Path.Combine(home, "bin\\java.exe");
+                return Path.Combine(home, "bin\\java");
```
elsassph: We probably also want to replace the `\` character in the path.

Mixaill: Fixed.

SlavaRa:
Maybe you should use ```Path.DirectorySeparatorChar```

Neverbirth:
Yeah, either use Path.DirectorySeparatorChar or Path.Combine... Path.Combine should be the best option, although in .NET 3.5 is not as nice as in later version.

wise0704:
`Path.Combine()` is also the easiest option as this
```c#
Path.Combine(home, "bin/java");
```
just becomes
```c#
Path.Combine(home, "bin", "java");
```

Mixaill:
``Path.Combine(string, string, string)`` appears only in .NET Framework **4.0**. 
FDBuild target is .NET Framework **3.5** by default.

Neverbirth:
Yeah, I'd vote for moving to .NET 4 at least, but the program should be tested on CrossOver... and this is another topic.

+1 to merge this already if it's been tested and works.

PluginCore/PluginCore/Helpers/JvmConfigHelper.cs
```diff
@@ -19,8 +19,8 @@ public class JvmConfigHelper
             if (Cache.ContainsKey(hash)) return Cache[hash];
             if (Directory.Exists(configPath))
             {
-                if (File.Exists(Path.Combine(configPath, "bin/jvm.config")))
-                    configPath = Path.Combine(configPath, "bin/jvm.config");
+                if (File.Exists(Path.Combine(configPath, Path.Combine("bin","jvm.config"))))
```
SlavaRa: Maybe you can extract local variable from code ```Path.Combine(configPath, Path.Combine("bin","jvm.config"))``` and ```Path.Combine(configPath, "build.properties"))``` ?

Mixaill: Fixed.

External/Tools/FDBuild/Program.cs
```diff
@@ -47,7 +47,7 @@ public static int Main(string[] args)
             if (File.Exists(Path.Combine(swfmillPath, "swfmill.exe")))
                 SwfmillLibraryBuilder.ExecutablePath = Path.Combine(swfmillPath, "swfmill.exe");
             else
-                SwfmillLibraryBuilder.ExecutablePath = "swfmill.exe"; // hope you have it in your environment path!
+                SwfmillLibraryBuilder.ExecutablePath = "swfmill"; // hope you have it in your environment path!
```
Neverbirth: Just noticed this part... shouldn't it be better that you add better checks using the swfmillPath? I'd say so... more chances of the people having swfmill there than in an environment path.

Mixaill: fixed

External/Tools/FDBuild/Program.cs
```diff
@@ -44,10 +44,12 @@ public static int Main(string[] args)
                 catch { }
 
             string swfmillPath = options.SwfmillPath ?? Path.Combine(toolsDir, "swfmill");
-            if (File.Exists(Path.Combine(swfmillPath, "swfmill.exe")))
+            if (File.Exists(Path.Combine(swfmillPath, "swfmill.exe"))) //Windows binary
                 SwfmillLibraryBuilder.ExecutablePath = Path.Combine(swfmillPath, "swfmill.exe");
+            else if (File.Exists(Path.Combine(swfmillPath, "swfmill"))) //Linux binary
```
Neverbirth: I would add a check for the OS not being Windows. Also, I just saw it's the same for mtascPath v_v. Sorry for not coming up with this before.

Mixaill: Done.

Neverbirth:
If nobody sees a problem I'll look into merging during the weekend.

PluginCore/PluginCore/Helpers/PlatformHelper.cs
```diff
@@ -0,0 +1,36 @@
...
+            return isRunningOnWindows && (Registry.LocalMachine.OpenSubKey(@"Software\Wine\") != null);
+        }
+
+        /// <summary>
+        /// Checks if we are running on Mono
+        /// </summary>
+        public static Boolean IsRunningOnMono()
+        {
+            return isRunningOnMono;
+        }
```
wise0704: It's just a minor thing about coding style, but
I think it's better to have them as getter properties than methods, since (unlike what the xml doc says) the methods only returns a fixed stored values instead of doing the checks on each call.
Usually a method is used instead of a getter property in cases where an extensive calculation is done on each call.

PluginCore/PluginCore/Win32.cs
```diff
@@ -5,43 +5,18 @@
...
-        {
-            return Registry.LocalMachine.OpenSubKey(@"Software\Wine\") != null;
-        }
-
-        /// <summary>
-        /// Checks if we are running on Mono
-        /// </summary>
-        public static Boolean IsRunningOnMono()
-        {
-            return !shouldUseWin32;
```
wise0704: When were these methods added? Were they added recent enough to be sure that 3rd party plugins don't use these methods?
If not, you could mark them deprecated and non-editor-browsable and delegate them back to the new entries.

PR #1294, opened 2016-08-09
Fixes when using hxml

Neverbirth:
- Parse hxml files used inside other hxml files. Doing this results in better project support: classpaths, finding --next statements inside other hxml, etc. I know the code is not very nice because of the method signature, but it's private code, so it's not harmful.
- When parsing the hxml file don't count next lines starting with # as the parameter of a previous command. For example having this inside the hxml could lead to wrong compiler service calling:
```
-debug
#comment
```
- When constructing the compiler command line from the project properties also quote lines for -cmd commands. Having something like -cmd bin/test.bat could break compiler service calling. Maybe we should also quote -x commands. Another option could be maybe to remove these commands, but I don't know if removing them could result in compiler cache problems or something else?

### DIFF
--- a/External/Plugins/HaXeContext/Completion/HaxeComplete.cs
+++ b/External/Plugins/HaXeContext/Completion/HaxeComplete.cs
@@ -21,7 +21,7 @@ namespace HaXeContext
     internal class HaxeComplete
     {
         static readonly Regex reArg =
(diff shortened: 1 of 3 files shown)


Neverbirth:
If nobody sees a problem, I'll try to merge this during the weekend.

PR #1300, opened 2016-08-25
Custom method to compact the text of document tabs with long filenames

Neverbirth:
Current rendering method is a sh*t:

current

The problem is that in order to draw the text in the tabs the code is using TextRenderer with TextFormatFlags.PathEllipsis, and it only removes full path parts, nothing else... which is useless with long filenames... Because of that latest DockPanelSuite releases modified it to use TextFormatFlags.EndEllipsis, which results in this:

newdps

Which is not very nice either, it's not uncommon to have multiple files with similar starts of filenames, or the same filenames but different extensions. Hence I decided to play a bit with different things to come up with this one which I think it's better for everyone:

new

StringTrimming.PathEllipsis used with Graphics.DrawString is able to render this case in a similar way to this custom implementation... but if you google you will find tons of posts from people discouraging to use it and keep TextRenderer, so I decided to use [this code](http://www.codeproject.com/Articles/37503/Auto-Ellipsis) (license fully compatible, of course), but I made further improvements to the path ellipsis behavior, because before it was still behaving in a similar way to TextRenderer, plus I "relaxed the requirements" the method needed.

The class works with a Font object and a width value, but it could be easily made to work with a max number of chars too, why I say this? Because I think the result is better than PathCompact for some cases (as detailed in the article), it's fully managed code, *and*, the managed version present in FD of PathCompact for when not using Win32 seems to be completely broken. But replacing one implementation with another could come later and by anyone else.

I could add in the future unit tests if you want.

### DIFF
--- a/PluginCore/DockPanelSuite/Customization/VS2005DockPaneStrip.cs
+++ b/PluginCore/DockPanelSuite/Customization/VS2005DockPaneStrip.cs
@@ -328,8 +328,7 @@ namespace WeifenLuo.WinFormsUI.Docking
         {
             get
             {   
-                 TextFormatFlags textFormat = TextFormatFlags.PathEllipsis |
-                    TextFormatFlags.SingleLine |
+                 TextFormatFlags textFormat = TextFormatFlags.SingleLine |
                     TextFormatFlags.VerticalCenter |
                     TextFormatFlags.PreserveGraphicsClipping |
                     TextFormatFlags.HorizontalCenter;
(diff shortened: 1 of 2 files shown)


Meychi:
Does this work correctly if you have two files  with the same name open from different folders. ie. (bin) appended?

Neverbirth:
Yes, the result depends on the text and the available size, it could output bin/filename.ext, .../longf...me.ext, etc.

I'll post some pics later when I'm back at home. At any rate, it would be easy to extend the method and add more collapsing options.

Neverbirth:
Note, if for directoryname/filename.ext .../filename.ext is not desired and you prefer directo...ame.ext it's also possible by removing the path option and trim only the middle.

Neverbirth:
@Meychi ok, in the case you are talking about (bin) is appended at the end (as you said), it's not bin/filename.ext. Well, here is current version:

duplicatedcurrent

And this is with this PR:

duplicatednew

Meychi:
Exactly, great.

Neverbirth:
I didn't check, but I guess that if it is a really long directory name, something like `SuperLongFilename.ext` in `SuperLongfoldername` it would get to be something like `SuperLongF...ngfoldername)`. But as I said, we could add custom compacting methods so we could have `Sup...ame.ext (Sup..rname)`, or something like that.

Meychi:
I think the path is added after so i think it should be fine.

Issue #1298, opened 2016-08-25
[question] How do i change the shortcut keys `ctrl+space` to other keys?


R32:
Because `Ctrl + space` has been occupied by system IME。

Neverbirth:
Wow, really? cannot you change it? In Tools -> Shortcuts, it's called ASCompletion.CtrlSpace.

R32:
But I can't find any that are associated with ASCompletion.CtrlSpace or ctrl + space

Neverbirth:
Yeah, just checked, the shortcut is not configurable in FD at the moment. When I replied I seemed to recall so, but I saw that ASCompletion.CtrlSpace reference and thought that maybe I was wrong.

Too bad. It can be fixed, but it's more than just setting it there, plus there is the Ctrl Space handling in other contexts.

wise0704:
The `Ctrl+Space`, `Ctrl+Shift+Space`, `Ctrl+Alt+Space` and `Ctrl+Alt+Shift+Space` combinations are all hardcoded, and it will take some time to make it configurable.
I intended to include it in the advanced shortcuts, but I ended up not touching it because I kind of found it unrelated to the aim of the PR.

wise0704:
Side note: `Ctrl+Space` never worked for me to change IME. `Win+Space` works for changing input language.

elsassph:
Although the hardcoded one persists, there should already be a way to set an alternate shortcut to trigger `Ctrl+Space`. I think that's `ASCompletion.CtrlSpace`; isn't it working anymore?

Neverbirth:
ASCompletion.CtrlSpace is not hooked at all, it's not even registered, so it only exists in comments of the source code. I don't remember seeing it at all any of the times I had to touch the related parts of code :S.

wise0704:
Those hardcoded shortcuts are better if changed to modifiable, as they're the only ones registered in `IgnoredKeys`, and rather create more mess in the shortcut mechanism.
I could work on this while I'm working on contextual shortcuts, as that will have to shake up many shortcut-related things anyway. Don't expect that to be happening anytime soon though.

Issue #1302, opened 2016-08-26
Advanced Shortcut - default main menu shortcuts not working


gene-pavlovsky:
Originally reported in [here](https://github.com/fdorg/flashdevelop/pull/1245#issuecomment-242407347)
E.g. `Alt+T` should open `Tools` menu, instead statusbar says "Alt+T was pressed, waiting for second key...".

Neverbirth:
@wise0704 The reason is "simple", command keys have priority over mnemonics, and they are separate concepts, hence ProcessCmdKey has to return false in order for a mnemonic to be processed.

I've checked VS, by default it doesn't have any list of possible "valid first extended shortcut", they list is built by the key mapping the user has, ie.: by default Alt+T will open the Tools menu, when the user sets an advanced shortcut starting with Alt+T, this would return true in `ShortcutKeysManager.IsValidExtendedShortcutFirst` and the Tools menu will stop working because it will start waiting for the second part of the command.

What do you think, could you do this soon? For me this one is not a top priority at the moment.

wise0704:
I will look again into the .NET shortcut system soon.

wise0704:
By shortcuts, I meant keyboard inputs.
When I was doing the PR, I ran through the decompiled assemblies. The mnemonics are handled in the `ToolStripManager` iirc, which is called in the `ProcessCmdKey()` in the root `Form`. 
Remember, at least for you and me, the mnemonics still work with the changes. There should be some other factor.

Neverbirth:
Yeah, I know what you meant, but for the .NET guys shortcuts were what's processed in ProcessCmdKey, that's why I said "not".

Again, mnemonics are not handled in there (after ProcessCmdKey returns false, after IsInputChar returns false, and other checks "fail" too), and any control may have them, so I don't think ToolStripManager gets a hold of all of them? (could be, but feels strange?). Forgot to say, for me mnemonics were working, but because I had some custom changes, my bad! with baseline development branch they wouldn't work.

Neverbirth:
Played with and checked a bit default keyboard implementation, and it doesn't seem that `ProcessCmdKey `checks mnemonics at all, in the default implementation if there is a shortcut and a mnemonic using the same combination, the shortcut will always have the preference (I played using toolstrips, normal controls, using different tab orders, etc, etc). It was as I was expecting.

[...]

PR #1305, opened 2016-08-29
Advanced shortcuts fixes

Neverbirth:
Check registered shortcuts to check if the pressed key combination is really a valid advanced shortcut, and revert method signature change in ASComplete.

If @wise0704 thinks something should be done in a different way he can later change it. The method signature change in ASComplete makes the show help shortcut to not accept advanced shortcuts, but I think this is really minor at the moment, the way this method is called should be different (through a proxy check maybe) so all the shortcuts like Ctrl+Space, Ctrl+Alt+Space, etc can be customized to solve #1298.

### DIFF
--- a/FlashDevelop/MainForm.cs
+++ b/FlashDevelop/MainForm.cs
@@ -8,6 +8,7 @@ using System.ComponentModel;
 using System.Diagnostics;
 using System.Drawing;
 using System.IO;
+using System.Linq;
 using System.Reflection;
 using System.Text;
 using System.Text.RegularExpressions;
@@ -1700,6 +1701,9 @@ namespace FlashDevelop
              */
             else if (ShortcutKeysManager.IsValidExtendedShortcutFirst(keyData) && !appSettings.DisableExtendedShortcutKeys)
             {
+                if (ShortcutManager.AllShortcuts.FirstOrDefault(s => s.First == keyData) == ShortcutKeys.None)
+                    return false;
+
                 StatusLabelText = string.Format(TextHelper.GetString("Info.ShortcutWaiting"), currentKeys);
                 lockStatusLabel = true;
                 return true;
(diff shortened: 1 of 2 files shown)


Neverbirth:
I don't merge this by myself to let people see it, but this should be merged ASAP.

FlashDevelop/MainForm.cs
```diff
@@ -1700,6 +1701,9 @@ protected override bool ProcessCmdKey(ref Message msg, Keys keyData)
              */
             else if (ShortcutKeysManager.IsValidExtendedShortcutFirst(keyData) && !appSettings.DisableExtendedShortcutKeys)
             {
+                if (ShortcutManager.AllShortcuts.FirstOrDefault(s => s.First == keyData) == ShortcutKeys.None)
+                    return false;
+
```
wise0704: If this is going to be used, it should be added in line with the if condition above it, like
```c#
else if (ShortcutKeysManager.IsValidExtendedShortcutFirst(keyData)
    && !appSettings.DisableExtendedShortcutKeys
    && ShortcutManager.AllShortcuts.FirstOrDefault(s => s.First == keyData).IsExtended)
```
so that the undefined shortcut would go to the `else if` below.

But this behaviour rather makes it complicated for the users. Regardless of whether shortcut is defined or not, when an undefined shortcut is pressed, the app setting is not set to disable extended keys, and the pressed key is not F1~F24 (contains Ctrl, Alt and/or Shift), is should always be regarded as the first part of the combination.

wise0704:
#1302 should be resolved in a different manner in my opinion, than to return false when no extended shortcut of the type exists.
I didn't look much into why I changed the signature of the method in `ASCompletion`, and I didn't yet look into which other methods I changed the signatures. I will soon and reply.

FlashDevelop/MainForm.cs
Neverbirth: If you do it that way it would be still wrong, because it would say Alt+T is undefined, return true, and the mnemonic wouldn't be executed. I don't see in VS notifyications about undefined shortcuts for simple cases, I'd say if you want to have that one then you'd have to check all the menustrips and controls to check for mnemonics (as a button, label, etc may have a mnemonic).

wise0704: When I wrote that, I didn't realise that was a fix for the mnemonics.
While it does work, I wouldn't really want to change the behaviour like this.
If it's urgent, I don't mind merging this, but I will definitely be changing this later.

PR #1306, opened 2016-08-29
Extended Shortcut Backward Compatibility

wise0704:
Closes #1303 
Sorry @Neverbirth :P

PR #1307, opened 2016-08-30
Extended Shortcut Bug Fixes

### LINKED ISSUE #1302: Advanced Shortcut - default main menu shortcuts not working

wise0704:
Commit e80f2dcaffa06908be449879ca892370d4f3f427 is already merged in #1306, but is showing up again for some reason.

Mnemonics should work after this commit:
 - If the current input begins with `Keys.Alt` as the only modifier, and no simple shortcut is handled and there is no extended shortcut that begins with the input as the first part, `MainForm.ProcessMnemonic()` will be called, and if it returns `true`, meaning the mnemonic is defined and worked, the input will not be registered as a first part of an extended shortcut sequence.

Character input with AltGr should work:
 - If the current input has `Keys.Control | Keys.Alt` with an optional `Keys.Shift` as the modifier, and no simple shortcut is handled and the virtual key code and the keyboard state corresponds to a non-null character, the input will not be treated as a first part of an extended shortcut sequence, _regardless of whether there is a shortcut that begins with the specific keys as the first part_, and the input will not be processed in `MainForm.ProcessCmdKey()`.

Edit: closes #1302


PluginCore/PluginCore/Managers/ShortcutKeysManager.cs
```diff
@@ -189,6 +211,28 @@ public static bool IsValidExtendedShortcutSecond(Keys second)
...
+                keyStates[(int) Keys.ShiftKey] = 0xFF;
+            }
+            if ((keyData & Keys.Control) == Keys.Control)
+            {
+                keyStates[(int) Keys.ControlKey] = 0xFF;
+            }
+            if ((keyData & Keys.Alt) == Keys.Alt)
+            {
+                keyStates[(int) Keys.Menu] = 0xFF;
+            }
```
wise0704: I didn't feel the need to call [GetKeyboardState()](https://msdn.microsoft.com/en-us/library/windows/desktop/ms646299(v=vs.85).aspx) as I could just do it this way. 
Would there be any problem doing this?

wise0704:
I see, e80f2dc turned into 6e9da15.

Usage of `MainForm.HandleShortcutManually()` has changed a bit:
 - Return value is the same way as `MainForm.ProcessCmdKey()`, so that now passing its return value directly when using inside `ProcessCmdKey()` should behave more correctly.
 - The ref argument `previousKeys` should not be used after calling the method now, since its value gets altered in the method. Previously the value was not altered so it could be used after the method, but that caused the method behave slightly differently from `MainForm.ProcessCmdKey()`, as well as the variable passed as `previousKeys` had to be set to `None` manually. Dropping the previous key usage was necessary in order to make this method follow the changes `MainForm.ProcessCmdKey()` has gone through.

Neverbirth:
I have the feeling this should be thoroughly tested. Aside from the char conversion that I don't fully trust, I see for example ProcessMnemonic is manually called, but before reaching that point there are more things happening: IsInputKey, ProcessDialogKey, IsInputChar and several others that controls or plugins may be overriding (although not so common together with the Alt key).

wise0704:
`Control.PreProcessMessage()`
- `Msg` is `WM_KEYDOWN` or `WM_SYSKEYDOWN`
  - `Control.ProcessCmdKey()` is first called. If it returns `false` ->
  - `Control.IsInputKey()` is called. This method checks if the pressed keys are regular input keys. If this returns `false` ->
  - `Control.ProcessDialogKey()` is called. 
    - `Form.ProcessDialogKey()` only checks for `Enter` and `ESC` key presses, and skips if either `Ctrl` or `Alt` is pressed.
    - `ContainerControl.ProcessDialogKey()` only checks for `Tab` and arrow keys, and also skips if `Ctrl` or `Alt` is pressed.
    - `Control.ProcessDialogKey()` calls the method on its parent, if not `null`.
    - If `false`, `Control.PreProcessMessage()` exits, sending char messages.
- `Msg` is `WM_CHAR` or `WM_SYSCHAR`
  - If `Msg` is `WM_SYSCHAR` or `Control.IsInputChar()` returns false ->
  - `Control.ProcessDialogChar()` is called. This calls `Control.ProcessMnemonic()`.

In the case when I'm directly calling `ProcessMnemonic()`, `ProcessDialogKey()` is redundant since `Alt` is pressed, meaning `IsInputKey()` is redundant as well. The `WM_SYSKEYDOWN` message gets translated by `TranslateMessage()`, into `WM_SYSCHAR`, since `Alt` is pressed. This also skips the call to `IsInputChar()`, making it directly call `ProcessDialogChar()`, which calls `ProcessMnemonic()` if "`(this.IsMdiChild && (int) charCode != 32)`".
(I could also call `ProcessDialogChar()` instead of `ProcessMnemonic()`.)

Since we're dealing with `MainForm`, we know that there are no overrides of `ProcessDialogKey()` and `ProcessDialogChar()`, and that there are no parent controls of `MainForm`, so they method being have to bubble up is not the case either.

In terms of the logic in `MainForm.GetUndefinedShortcutType()`, I've tested it in Swedish (Finland) keyboard.

wise0704:
As for `ToUnicode()` method:

Our objective is to let character inputs pass through `WM_KEYDOWN`.

When `WM_KEYDOWN` or `WM_SYSKEYDOWN` messages are unhandled (i.e. our `ProcessCmdKey()` should return `false`), the message is passed to [TranslateMessage](https://msdn.microsoft.com/en-us/library/windows/desktop/ms644955(v=vs.85).aspx).

`TranslateMessage()` function either calls [ToUnicode](https://msdn.microsoft.com/en-us/library/windows/desktop/ms646320(v=vs.85).aspx) or [ToAscii](https://msdn.microsoft.com/en-us/library/windows/desktop/ms646316(v=vs.85).aspx). 
>I don't know which... but it shouldn't matter I think, and if it does matter, `ToUnicode()` should be better because it seems to handle multibyte chars better (?) when I read the documentation. Correct me on this part.
When the method returns a non-zero value, `WM_CHAR` or `WM_SYSCHAR` is dispatched, letting the character to be input.

This means that the key inputs that returns non-zero values on `ToUnicode()` are character inputs and can be dispatched to `WM_CHAR` or `WM_SYSCHAR`, and we want these to be passed through `WM_KEYDOWN` (i.e. `ProcessCmdKey()`).
Therefore I let those with `Ctrl+Alt` modifiers, which return non-zero values, pass through `ProcessCmdKey()` and be treated the same as invalid shortcuts (which are keys with no modifiers except `Shift`, meaning a possible character input as well).

wise0704:
#1316 reverted changes

gene-pavlovsky:
Just updated to FD 5.2.0 and I see the Advanced shortcuts features was pulled from that release. That's too bad, I was really expecting to try them out. How is it going, guys?

PR #1309, opened 2016-09-01
Theme and Size LiveDataTip

JoeRobich:
- Delay initializing DataTreeControl in DataTipForm until theming has been loaded
- Apply DataTreeControl ForeColor and BackColor theming
- Size DataTipForm based on AutoHeightRowLayout Bounds height for rows since TreeViewAdv.RowSize doesn't return the value based on AutoHeight being applied
- Default DataTipForm Column widths based on the width of their Header text

Problem:
LiveDataTip with Dark Themeing Looks Bad

### DIFF
--- a/External/Plugins/FlashDebugger/Controls/DataTipForm.cs
+++ b/External/Plugins/FlashDebugger/Controls/DataTipForm.cs
@@ -26,9 +26,20 @@ namespace FlashDebugger.Controls
             }
         }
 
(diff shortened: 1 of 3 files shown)


Neverbirth:
Nice to see this one, it always bugged me but never got to fix that theming issue because I rarely use custom themes.

The LiveDataTip will be rewritten soon, but it will still reuse some of the same code so it will be equally useful.

@Meychi I'd say this is OK to merge.

PR #1310, opened 2016-09-01
Support of AIR 23

SlavaRa:
AIR 23 Beta Release notes: http://fpdownload.macromedia.com/pub/labs/flashruntimes/shared/air23_flashplayer23_releasenotes.pdf


Neverbirth:
I think this is OK to merge. @Meychi

PR #1316, opened 2016-09-07
Revert shortcuts

Meychi:
What do you think?


wise0704:
Could we then merge #1307 first, then revert that as well?
That way it messes up with my branch less, as well as keeping the fix.

PR #1318, opened 2016-09-07
[Haxe 3.3] Adds Support of -D display-stdin. closes #1311

### LINKED ISSUE #1311: [Haxe] Support of -D display-stdin
https://github.com/HaxeFoundation/haxe/pull/5120

SlavaRa:


Neverbirth:
Hmmm, what about multiple unsaved files at the same time? It seems it only gets the content of the current file are the rest are ignored. Am I right? Can we pass the contents of multiple files or only the one you want to get completion from? if it's the second then this seems to have severe drawbacks, doesn't it?

nadako:
It's the second. We thought about this, but decided not to overcomplicate things for now, since the main issue was that IDEs basically had to save the current file on each keystroke and this fixes it.

To do it "properly", we would have to do something like e.g. Roslyn does, when source files are synchronized in the compiler with IDE's in-memory contents using some protocol. Haxe is not ready for this right now, so the current display-stdin solution is the best we can do for the time being.

nadako:
By the way, I'm not really familiar with FD's codebase, but it looks like this PR doesn't implement passing stdin in the non-server mode, which can also be done (you just call `haxe -D display-stdin --display ...` and write contents to the standard input stream of the process.

nadako:
(see description here https://github.com/HaxeFoundation/haxe/pull/5120)

Neverbirth:
Then I guess FD should still save the other files to be on the safe side?

SlavaRa:
I think that I could save other modified files, but not current file. @Neverbirth  what you this about this?

Neverbirth:
It seems we wrote mostly at the same time, yeah, I'm OK with that one.

SlavaRa:
@Neverbirth fixed

elsassph:
When multiple files are modified a nice solution could be to Save all modified excepted the current one.

elsassph:
Amazing stuff guys

SlavaRa:
@elsassph @Neverbirth @Meychi what about this PR?

External/Plugins/HaXeContext/Context.cs
```diff
@@ -1209,6 +1209,14 @@ public override MemberList ResolveDotContext(ScintillaNet.ScintillaControl sci,
             return null; 
         }
 
+        HaxeComplete GetHaxeComplete(ScintillaControl sci, ASExpr expression, bool autoHide, HaxeCompilerService compilerService)
+        {
+            var sdkVersion = GetCurrentSDKVersion();
+            if (hxsettings.CompletionMode == HaxeCompletionModeEnum.CompletionServer && new SemVer("3.2.1").IsOlderThan(sdkVersion))
```
elsassph: Maybe do a semver greater-than-or-equal 3.3.0 if it's possible.

External/Plugins/HaXeContext/HaxeComplete330.cs
```diff
@@ -0,0 +1,37 @@
...
+            var args = base.BuildHxmlArgs();
+            if (args == null) return null;
+            var list = new List<string>(args) {"-D display-stdin"};
+            var result = list.ToArray();
+            return result;
+        }
+
+        protected override string GetFileContent()
+        {
+            return Sci.Text;
```
elsassph: Is there any possible filesize concern here? What about if you're editing a very large file?

External/Plugins/HaXeContext/Context.cs
SlavaRa: Fixed https://github.com/fdorg/flashdevelop/pull/1318/commits/7c68005059ebd11b13cd1a3f65420fd4ead20689

External/Plugins/HaXeContext/HaxeComplete330.cs
SlavaRa: My test file contains more than 20,000 lines of text and all ok)

PR #1322, opened 2016-09-08
[Haxe] Fixed generation of override of property.

SlavaRa:
Overriding the property is only valid for accessor methods of reading and / or writing
but the Editor generate override of property without access methods.
code for example:
```haxe
var foo(get, never):T
```
actual result after generate override:
```haxe
//@:getter(foo)
override function get_foo():T
{
    return super.get_foo();
}

//@:setter(foo)
override function set_foo(value:T):T
{
    return super.set_foo(value);
}
```
expected result:
```haxe
//@:getter(foo)
override function get_foo():T
{
    return super.get_foo();
}
```

Fixed generation of override setter, actual:
```haxe
//@:setter(foo)
override function set_foo(value:T):T
{
    return set_foo(value);
}
```
expected:
```haxe
//@:setter(foo)
override function set_foo(value:T):T
{
    return super.set_foo(value);
}
```

### DIFF
--- a/External/Plugins/ASCompletion/Completion/ASGenerator.cs
+++ b/External/Plugins/ASCompletion/Completion/ASGenerator.cs
@@ -3840,7 +3840,7 @@ namespace ASCompletion.Completion
             List<MemberModel> members = new List<MemberModel>();
             curClass.ResolveExtends(); // Resolve inheritance chain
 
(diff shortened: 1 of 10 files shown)


SlavaRa:
@Neverbirth @elsassph please review this PR and merge if all ok

PR #1327, opened 2016-09-11
Implemented CodePeek on Ctrl+Hover

### LINKED ISSUE #1019: Code peek
https://twitter.com/Jeff__Ward/status/681216872189050880

I saw this, it would be a nice addition for our editor too?!

JoeRobich:
resolves #1019 

When hovering over a member with the control key pressed a CodeTip will popup displaying the lines of code where the member is defined.

codepeek1

codepeek2

### DIFF
--- a/External/Plugins/ASCompletion/Completion/ASComplete.cs
+++ b/External/Plugins/ASCompletion/Completion/ASComplete.cs
@@ -4046,6 +4046,52 @@ namespace ASCompletion.Completion
         #endregion
 
         #region tooltips formatting
+        static public string GetCodeTipCode(ASResult result)
+        {
+            if (result.Member == null)
+            {
+                return null;
+            }
+
+            var file = GetFileContents(result.InFile);
+            if (string.IsNullOrEmpty(file))
+            {
+                return null;
+            }
+
+            var lines = file.Split(new[] { Environment.NewLine }, StringSplitOptions.None);
+            var code = new StringBuilder();
+            for (var index = result.Member.LineFrom; index < result.Member.LineTo; index++)
+            {
+                code.AppendLine(lines[index]);
+            }
+            code.Append(lines[result.Member.LineTo]);
+            return code.ToString();
+        }
+
+        static string GetFileContents(FileModel model)
+        {
+            if (model.FileName.Length > 0 && File.Exists(model.FileName))
+            {
+                foreach (ITabbedDocument doc in PluginBase.MainForm.Documents)
+                {
+                    if (doc.IsEditable && doc.FileName.ToUpper() == model.FileName.ToUpper())
+                    {
+                        return doc.SciControl.Text;
+                    }
+                }
+                var info = FileHelper.GetEncodingFileInfo(model.FileName);
+                return info != null ? info.Contents : null;
+            }
+            else
+            {
+                model.Members.Sort();
+                foreach (ClassModel aClass in model.Classes) aClass.Members.Sort();
+                string src = "//\n// " + model.FileName + "\n//\n" + model.GenerateIntrinsic(false);
+                return src;
+            }
+        }
+
         static public string GetToolTipText(ASResult result)
         {
             if (result.Member != null && result.InClass != null)
--- a/External/Plugins/ASCompletion/Completion/ASComplete.cs
+++ b/External/Plugins/ASCompletion/Completion/ASComplete.cs
@@ -4053,30 +4053,10 @@ namespace ASCompletion.Completion
                 return result.Type != null ? result.Type.ToString() : null;
             }
 
-            if (result.InFile == null)
-            {
-                return null; // Could do something similar to MemberToolTip
-            }
-
             var file = GetFileContents(result.InFile);
             if (string.IsNullOrEmpty(file))
             {
-                return null;
-            }
-
-            if (result.Member.LineFrom == 0 && result.Member.LineTo == 0)
-            {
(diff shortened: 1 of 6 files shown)


JoeRobich:
I apologize for the changes to ASCompletion.cs. I seem to have a VS extension that is removing trailing spaces. The changes are really minor when you look at the diff with ignore whitespace enabled. 

https://github.com/fdorg/flashdevelop/pull/1327/files?w=1

External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -4063,8 +4068,15 @@ static public string GetCodeTipCode(ASResult result)
             {
                 ClassModel inClass = result.InClass ?? result.Type;
                 result.InFile = ASContext.Context.GetCodeModel(file);
-                inClass = result.InFile.GetClassByName(inClass.Name);
-                result.Member = inClass.Members.Search(result.Member.Name, 0, 0);
+                if (inClass.Name != "void")
```
SlavaRa: i think that you should use Context.Features.voidKey instead of 'void'

JoeRobich: Thank you for the suggestion. I changed to comparing against `ClassModel.VoidClass`.

markknol:
Hi, the screenshot looks great, however I tried this but it does not seem to do anything in Haxe? I installed the latest HaxeDevelop build. I keep de control key down and hovered many things but I only see the yellow tooltip.

image

JoeRobich:
@markknol There won't be a development build with this code in it until this PR is accepted. If you felt comfortable building FlashDevelop from sources you could try it out sooner.

I don't develop with Haxe but thought that I would pull down a library to test with. Here is a screenshot from one of the Samples included with the Luxe library.
codepeek3

Also, I love QuickNavigate and these changes play nicely with it.

markknol:
Ah sorry, I was confused since https://github.com/fdorg/flashdevelop/issues/1019 was closed so I assumed it was merged. Looking forward to this feature!

Neverbirth:
I'd like to be able to provide builds for PRs in the future so more people can give feedback.

JoeRobich:
One thing I've noticed is that the ABCConverter doesn't read the value out of slots. This causes the CodeTip not to show the value for constants from libraries since the MemberModel doesn't have a Value set.

I've made a change locally to populate them but I'm not sure that is a change you would want.

Neverbirth:
Just wanted to notice in case it was not known, that this PR didn't move forward because of the upcoming near release. Don't worry!

Neverbirth:
This one will be merged during this week.

SlavaRa:
I installed the editor version 5.2.1.26.
I wanted to try this functionality, I aim cursor on the code with pressed ctrl, but nothing happens(

markknol:
I tried it too, doesn't do anything yet either.

Neverbirth:
True, I guess some incompatibility happened since last time the PR was checked, but I prefer not to enable forcing up to date PRs because it will make some things to progress in a slower way.

Surely this is an easy fix, so not worth worrying.

JoeRobich:
I just installed 5.2.1.28 and am seeing the functionality work in a Flex project that I have.

SlavaRa:
issue1327

Neverbirth:
After some investigation I found on some setups DwellStart is not triggering when Ctrl is pressed.... it will need some more investigation.

JoeRobich:
I've come up with a hacky way to get dwell to fire after pressing the control key. https://github.com/JoeRobich/flashdevelop/commit/3ce418ab8878b55144c59925a6d6176298aa17be

Neverbirth:
I prefer to avoid that hack as much as possible, DoEvents can bring many unexpected side effects. Maybe there are other solutions? I'm away this weekend, but maybe I can help with this from tomorrow.

JoeRobich:
Totally understand. 

This one may be a little better. It works so long as DwellStart hasn't fired and triggered the regular tooltip. https://github.com/JoeRobich/flashdevelop/commit/a288cb741d500444c492868ccd4291cf2ff16c39

I've been looking through the scintilla source trying to figure out to reset the dwelling timer in the case that dwelling was ended by a key press, but haven't worked anything out yet that doesn't involve moving the cursor.

Neverbirth:
That hack looks better, but it seems it doesn't really solve the issue? The behavior is the same in my current machine. Maybe you could write in the Scintilla group about this issue? If you prefer we can always discuss this subject Slack.

Neverbirth:
After playing a bit more with this I noticed this behaviour:

- If I hover over a known variable, and the tooltip pops up, and then I try to display the CodePeek, it won't work.
- If I hover over an unknown word, and then I try to display the CodePeek over a known member, it will work.

JoeRobich:
Once a ToolTip has popped up, pressing any key including the control key will trigger DwellEnd and the dwell time is effectively set to infinity within scintilla. The only way to reset the dwell time and have scintilla add back the dwell timer is with a mouse move. https://sourceforge.net/p/scintilla/code/ci/default/tree/src/Editor.cxx#l4362

I've tried sending fake WM_MOUSEMOVE messages via SPerform, but couldn't get that to work. You can see above I also tried moving the mouse in the DwellEnd handler which does work for me, but now the mouse is in a new location which isn't ideal.

Neverbirth:
I've also found the following bug: https://github.com/fdorg/flashdevelop/issues/1384.

PR #1335, opened 2016-09-14
[Haxe] Fixed contextual generator for  `var some(get, set)`

SlavaRa:
Before fix:
bug_contextualgenerator_for_getset

Result after fix:
fixed_contextualgenerator_for_getset


Neverbirth:
I guess we'll need to also add unit tests for this part at some point, because I think this is OK for merging, but cannot give 100% of confidence right now. @Meychi ?

PR #1339, opened 2016-09-15
Get Haxe libraries from project CompilerOptions instead of building virtual HXML

Neverbirth:
Building a virtual HXML is a problem, because if rawHxml is set it will return that, and if libs are defined in HXMLs inside other HXMLs this info gets lost, causing some FD features to not work properly.

I don't know why we'd want to return a virtual HXML, I checked and it's quite old code, so maybe it was needed back in the day... nowadays I don't see a need for doing it that way?

### DIFF
--- a/External/Plugins/HaXeContext/Context.cs
+++ b/External/Plugins/HaXeContext/Context.cs
@@ -409,10 +409,10 @@ namespace HaXeContext
             // add haxe libraries
             if (proj != null)
             {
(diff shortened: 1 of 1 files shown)


Neverbirth:
Hmmm, I can see that if someone has inside a HXML --next before some -lib entry then FD would drop it doing it my way, but it's already this way with classpaths, so I think my way is still better than the old one.

SlavaRa:
I tested this PR, all ok.
@Neverbirth please merge this PR.

PR #1341, opened 2016-09-16
Avoid some unneeded exceptions.

Neverbirth:
Just avoid some unneeded exceptions.

### DIFF
--- a/External/Plugins/ProjectManager/Controls/TreeView/OtherNodes.cs
+++ b/External/Plugins/ProjectManager/Controls/TreeView/OtherNodes.cs
@@ -33,7 +33,11 @@ namespace ProjectManager.Controls.TreeView
 
             if (References != null && References.Parent == null)
             {
-                if (recursive) RefreshReferences(recursive);
+                if (recursive)
+                {
+                    RefreshReferences(recursive);
+                    return;
+                }
                 Nodes.Insert(0, References);
             }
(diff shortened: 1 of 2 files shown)


External/Plugins/ProjectManager/Controls/TreeView/OtherNodes.cs
```diff
@@ -33,7 +33,11 @@ public override void Refresh(bool recursive)
 
             if (References != null && References.Parent == null)
             {
-                if (recursive) RefreshReferences(recursive);
+                if (recursive)
+                {
+                    RefreshReferences(recursive);
```
Neverbirth: This causes an exception in `Nodes.Insert(0, References)`, because `RefreshReferences` already parents the references node.

PluginCore/PluginCore/Helpers/DrawHelper.cs
```diff
@@ -11,6 +11,8 @@ public class DrawHelper
         /// </summary>
         public static Int32 MeasureDisplayStringWidth(Graphics graphics, String text, Font font)
         {
+            if (String.IsNullOrEmpty(text)) return 0;
```
Neverbirth: We don't need to run all the code below in this case, plus in the end it will throw an `IndexOutOfRangeException`.

PR #1346, opened 2016-09-22
Fixed priority of sorting items of completion list for imports and type declaration. fixes #812

### LINKED ISSUE #812: Auto class import priority for global class
global (no package) classes should have higher priority then classes in packages:

image

For example, in the screenshot when I type `:Xml` FD will auto import neko.NativeXml.Xml for me which is not desired. I want the global Xml instead.

### LINKED ISSUE #821: [suggestion] autocompletion list, "up"-key wrap
It would be nice if you press the arrow "up" key when you are on top of the list, it would select to one at the bottom of the list. Now it just stops.

image

SlavaRa:


Neverbirth:
This doesn't have anything to do with #821.

SlavaRa:
sorry, #812

Neverbirth:
@elsassph @Meychi ?

SlavaRa:
@elsassph @Meychi guys, what you think about this PR?

Meychi:
@SlavaRa, this has conflicts now.

Neverbirth:
Now that I check this better, is it really good to use this sorting method only in the HandleAllClassesCompletion? I don't think so, because depending on the settings code may go another way (I don't remember right now the function name).

PR #1350, opened 2016-09-26
SourceControl intergrated blame

wise0704:
Added integrated support for "blame" feature. Currently only added support for git.

Support for other vcs plugins can be added by following the pattern used in Sources/Git/BlameCommand.

Example:
image

@Neverbirth

### DIFF
--- a/PluginCore/ScintillaNet/Events.cs
+++ b/PluginCore/ScintillaNet/Events.cs
@@ -19,8 +19,8 @@ namespace ScintillaNet
     public delegate void NeedShownHandler(ScintillaControl sender, int position, int length);
     public delegate void UserListSelectionHandler(ScintillaControl sender, int listType, string text);
     public delegate void URIDroppedHandler(ScintillaControl sender, string text);
-    public delegate void DwellStartHandler(ScintillaControl sender, int position);
-    public delegate void DwellEndHandler(ScintillaControl sender ,int position);
+    public delegate void DwellStartHandler(ScintillaControl sender, int position, int x, int y);
+    public delegate void DwellEndHandler(ScintillaControl sender ,int position, int x, int y);
     public delegate void HotSpotClickHandler(ScintillaControl sender, int modifiers, int position);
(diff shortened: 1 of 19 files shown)


wise0704:
needs localisation for strings here and there, but would do after all the reviews

Neverbirth:
Note that there is quite randomly some Win32 exception, but we don't know what is the cause yet. Anyway, I want to merge this during this week.

PluginCore/ScintillaNet/Events.cs
```diff
@@ -19,8 +19,8 @@ namespace ScintillaNet
     public delegate void NeedShownHandler(ScintillaControl sender, int position, int length);
     public delegate void UserListSelectionHandler(ScintillaControl sender, int listType, string text);
     public delegate void URIDroppedHandler(ScintillaControl sender, string text);
-    public delegate void DwellStartHandler(ScintillaControl sender, int position);
-    public delegate void DwellEndHandler(ScintillaControl sender ,int position);
+    public delegate void DwellStartHandler(ScintillaControl sender, int position, int x, int y);
```
Meychi: Can this break compability with older plugins using this?

Neverbirth: Sadly @wise0704 is busy again with personal stuff like university. I'll try to make tests covering that soon.

Neverbirth: After some long time I finally checked this... although I've been using this in my own build since this was done and never found any problem with other plugins.

I found the following during my tests:

- Using .NET 3.5 this is fully compatible for old plugins.
- Using .NET 4+ this would be compatible for old plugins targetting .NET 4+.
- Using .NET 4+ this would be compatible for old plugins targetting .NET 3.5.
- Using .NET 3.5 for PluginCore this would be incompatible for old plugins targetting .NET 4+ as the handler would take `y` as `position`. **This wouldn't happen anyway**

Neverbirth:
Ok, this one will get merged soon.

PR #1359, opened 2016-09-29
[HaxeContext] Fixed parsing of hxml when building classpath

SlavaRa:
I have a project with `hxml` platform and 2 build files:
build.hxml
```
libs.hxml
-cp src
-main Main
```

libs.hxml
```
-lib openfl
-lib lime
```
Project properties -> Configuration file: build.hxml
The editor does not display references to libs from libs.hxml in Project Panel and thus does not use libs for building classpathes

### DIFF
--- a/External/Plugins/HaXeContext/Context.cs
+++ b/External/Plugins/HaXeContext/Context.cs
@@ -389,39 +389,49 @@ namespace HaXeContext
                 if (Directory.Exists(extraCP)) AddPath(extraCP);
             }
             HaxeProject proj = PluginBase.CurrentProject as HaxeProject;
-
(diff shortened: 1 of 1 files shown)


Neverbirth:
From my side I think [this solution](https://github.com/fdorg/flashdevelop/pull/1339) is better. Cleaner, easier to understand, more efficient, and also a bit less error prone as this one for example assumes that the hxml file exists.

I like how the block regarding Flash libraries was moved tho.

SlavaRa:
I'll check your decision in the morning

SlavaRa:
Duplicate #1339

PR #1361, opened 2016-10-02
Support of shortcuts "To Lowercase", "To Uppercase" for "Find and Replace" dialogs

SlavaRa:
@Meychi @elsassph @Neverbirth please review this PR and merge if all ok.

### DIFF
--- a/External/Plugins/ProjectManager/Helpers/LineEntryDialog.cs
+++ b/External/Plugins/ProjectManager/Helpers/LineEntryDialog.cs
@@ -167,20 +167,20 @@ namespace ProjectManager.Helpers
 
         void OnLineBoxOnKeyDown(object sender, KeyEventArgs args)
         {
-            string shortcutId = PluginBase.MainForm.GetShortcutItemId(args.KeyData);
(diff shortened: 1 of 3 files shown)


Neverbirth:
Personally I don't see this as something that would really improve UX or productivity, but I left the other guys to decide. What I don't like is the use of var for things like int.

SlavaRa:
Now I have to work with a large number of ported code from c# to haxe. One of the tasks is to rename the constants. I have to translate their names to upper case by hand and it takes a long time, so I decided to add hotkeys.
About var for int, I think that the style should be one for the string and int and so on

Neverbirth:
I don't really get your last sentence.

SlavaRa:
instead of
```csharp
int pos = 1;
var str = "string";
```
just
```csharp
var pos = 1;
var str = "string";
```
I think that the style of writing should not depend on the type of a variable.

Neverbirth:
Oh, well, then I guess we don't agree on it, I don't think type inference is an all or nothing case. Anyway, in this case I would write both int and string.

SlavaRa:
In this case the type is obvious when initializing.
Or we should opt out of the use of "var" for all.

Issue #467, opened 2014-10-25
Refactor->Rename fails when renaming function parameters in some cases


Gama11:
Renaming a parameter to have the same name as another variable currently available in the context can break the code.

Example:

``` haxe
class Test
{
    var v:Int; 

    public function new(w:Int)
    {
        v = w;
    }
}
```

Renaming `w` to `v` leads to:

``` haxe
class Test
{
    var v:Int; 

    public function new(v:Int)
    {
        v = v;
    }
}
```

However, it should be `this.v = v;`

Similarly, if `v` is a `static var`, it should turn into `Test.v = v;`.

SlavaRa:
Fixed

PR #1362, opened 2016-10-03
Fixed fails after renaming when renaming function parameters in some cases. part1 #467

SlavaRa:


SlavaRa:
@elsassph @Neverbirth please review this PR

Neverbirth:
Is it really needed to use FRSearch in this case? cannot we check the local vars of the function and the existing members in the class hierarchy?

SlavaRa:
I got rid of `FRSearch`, I wanted to do PR in the evening)

Neverbirth:
Not a problem at all.

PR #1366, opened 2016-10-07
Setting to define when to clear the output panel.

Neverbirth:
This is useful for many cases. One case:

- I have a Haxe project. I run it and the build starts. Output panel cleared.
- During the build I have some macros that outputs some stuff.
- The build finishes and I have for running the project another process. The output panel is cleared again and I lose the traces from the macros.

Another one:

- I have important stuff in the output panel.
- I run a FD macro. I lose the stuff in the output panel.


Meychi:
If this setting happens to restore to null, will it fallback to OnEveryProcess as with int values to 0?

PR #1373, opened 2016-10-11
Restore bookmarks after file reload 

Shenjoku:
This will restore bookmarks after a file reload so you don't lose your place as easily. This helps to make file reload a less destructive operation.


Shenjoku:
Messed this up...ignore this :)

PR #1374, opened 2016-10-11
Restore bookmarks after file reload

Shenjoku:
This will restore bookmarks after a file reload so you don't lose your place as easily. This helps to make file reload a less destructive operation.


Neverbirth:
This somehow feels hacky to me, but I think @Meychi is the one that knows this part the best.

FlashDevelop/Docking/TabbedDocument.cs
```diff
@@ -468,6 +468,13 @@ public void Reload(Boolean showQuestion)
                 this.SciControl.IsReadOnly = FileHelper.FileIsReadOnly(this.FileName);
                 this.SciControl.SetSel(position, position);
                 this.SciControl.EmptyUndoBuffer();
+
+                List<Int32> bookmarksCopy = this.bookmarks;
```
Shenjoku: You're right. I'm still fairly inexperienced with C# since I mainly use C++ so I thought that worked. I stepped through it again and it turns out that line isn't event needed since OnMarkerChanged() prevents it from adding duplicates to this.bookmarks, which I thought was an issue.

Shenjoku:
@Neverbirth I kinda feel the same way about this change, but I don't see any other possible way to fix it since the Scintilla system automatically clears all the markers when the file reloads. You have to manually re-apply them somehow, unless there's some way to tell it not to clear these particular markers.

Meychi:
I'm not sure about this, bookmarks may be totally off if you reload a file if it has changed a lot.

How do you use this? Manually or for example when SC has changed files?

Neverbirth:
Bookmarks and breakpoints being off because of file changes outside of it is a common problem in IDEs, some of them try to compensate storing also some of the type information in the line, but imho that's a bit extreme and still not bullet proof.

Shenjoku:
I use bookmarks to more easily remember what part of a file I was working on and if a file reload comes along and makes me lose all of those bookmarks it puts a huge dent in my workflow. Even if the file has changed, having those markers nearby the area I was working on would make a huge difference. Most of the time the files have very minor changes so they wouldn't be pushed that far off. I feel like it should be the user's decision to clear everything if they don't want to use them anymore.

wise0704:
I think this makes sense, since technically you're bookmarking lines, not pieces of code, so even if the file had been modified externally, the bookmarks should stay on their lines.

As long as it doesn't throw some exception for the lines being out of range, though. Documentation doesn't say anything on that about SCI_MARKERDELETE.

Neverbirth:
As no better solution or comment came up I'm merging this for now and will add the check for invalid line numbers.

PR #1376, opened 2016-10-12
Added ConsolePanel plugin

chipshort:
I think it might be a good idea to have something like this integrated into FlashDevelop.
I also posted it as a separate plugin on the forums: http://www.flashdevelop.org/community/viewtopic.php?f=4&t=12842


Meychi:
Hi, thanks for this contribution. We will review and test this.

Meychi:
It works nice but i had some flicker and minor lockups, could these be improved? Does this work under wine? If not, does it disable itself correcly?

Also i think the menu label should be Console Panel as the plugin name states. Also if you wish to contribute this please change the info to match the other plugins.

SlavaRa:
WIP

SlavaRa:
via https://github.com/fdorg/flashdevelop/pull/3098

PR #1377, opened 2016-10-13
Fix to open read-only projects

Shenjoku:
Currently, an exception will be thrown if you attempt to open a FD project files that is read-only. This is a problem when using source control programs like Perforce that force everything to read-only if you aren't editing them.

This fix will allow FD to open the project as normal. You won't be able to edit it in any way, but it gives you a clear error when you try so it's not an issue.


Shenjoku:
That didn't work like I thought it would...Still trying to figure out this git stuff. Sorry.

Neverbirth:
No problem. You'll have to make a new branch from development, apply your project changes there and make the PR for that branch.

PR #1378, opened 2016-10-13
Fill the find-and-replace text with current selection every time it's activated

Shenjoku:
When opening the find-and-replace dialog the first time it automatically populates the find text with whatever you have selected. Every additional attempt will no longer update the text.

This change makes it so every time you activate the dialog, usually through Ctrl+Shift+F, it will update the find text to use whatever you had selected at the time, making it much easier to use.


Neverbirth:
The Find and Replace dialog should also be updated to have consistent behavior.

Shenjoku:
Ahhhh I see what you mean. I forgot there was a second dialog. Looking into it now.

Issue #1165, opened 2016-04-06
ProjectManager exception on delete when selection includes files and directories


Gama11:
I assume that's what causes it anyway. Deleting only files or only directories works.

![](http://i.imgur.com/HvbcILj.gif)

Shenjoku:
The problem here is that you've selected a directory for delete as well as files within that directory. It deletes the directory first, deleting everything within, then it attempts to delete those files that no longer exist and the exception is thrown. It needs to be smarter about detecting that the file no longer exists before trying to delete it.

PR #1380, opened 2016-10-14
#1165 Fix handling of file delete in ProjectManager

Shenjoku:
Issue #1165: The issue is that if you select a directory and some files inside of it for deletion, it would throw an exception trying to delete the files inside the directory. This is because the directory and all its files will get deleted first, and then it will attempt to delete the files that no longer exist, causing the exception.

The fix is to make the deletion smarter by parsing the paths and ignoring all files that are included in a directory about to be deleted.


Neverbirth:
Maybe @SlavaRa and some other people can take a look into this before I do?

SlavaRa:
I will look tonight

External/Plugins/ProjectManager/Actions/FileActions.cs
```diff
@@ -385,79 +385,17 @@ public void Delete(string[] paths)
...
+                            if (File.Exists(path) || Directory.Exists(path))
                             {
                                 if (!FileHelper.Recycle(path))
                                 {
                                     String error = TextHelper.GetString("FlashDevelop.Info.CouldNotBeRecycled");
                                     throw new Exception(error + " " + path);
                                 }
-                                OnFileDeleted(path);
                             }
+                            OnFileDeleted(path);
```
Shenjoku: I'm a little unsure what how to handle OnFileDeleted(). Should it be inside the exist checks or outside? Technically if the file no longer exists then it _was_ deleted by the directory delete operation. So should it still call OnFileDeleted() on those files?

PR #1381, opened 2016-10-14
Improve completion after new

### LINKED ISSUE #71: No hint on new Vector<T>
When I declare a Vector variable and trying to initate it, the hint when I type new directs me to adobe.utils.CustomActions, when it should hint to Vector.

Similar forum post here: http://www.flashdevelop.org/community/viewtopic.php?f=6&t=11386

OS: Win7 x64
FD: 4.5.2.5
Project: AS3 Project

Neverbirth:
- If we are instantiating an nonexistent item add it to the list and notify about its missing state.
  - If we are instantiating a generic type, add it with the index type to the list.
This closes #71.

Sample:

selectnew

EDIT: There are some artifacts in the GIF for some reason.

### DIFF
--- a/External/Plugins/ASCompletion/Completion/ASComplete.cs
+++ b/External/Plugins/ASCompletion/Completion/ASComplete.cs
@@ -2201,15 +2201,28 @@ namespace ASCompletion.Completion
                 ICompletionListItem newItem;
                 if (!aClass.IsVoid())
                 {
-                    if (!string.IsNullOrEmpty(aClass.IndexType))
+                    // AS2 special srictly typed Arrays supports
+                    int p = newItemType.IndexOf('@');
+                    if (p > -1)
                     {
-                        newItemType = !string.IsNullOrEmpty(aClass.IndexType) ? aClass.QualifiedName : aClass.Type;
+                        newItemType = newItemType.Substring(0, p);
+                        newItem = null;
+                    }
+                    else if (!string.IsNullOrEmpty(aClass.IndexType))
+                    {
+                        newItemType = aClass.QualifiedName;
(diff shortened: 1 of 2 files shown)


SlavaRa:
#1346 should be closed after merge this PR

Neverbirth:
Why? That PR may bring other UX improvements, I don't think they are incompatible.

SlavaRa:
oh sorry

Neverbirth:
I had to write the PR in a rush, so maybe some more details are welcome:

The first case is nice when prototyping or writing some code before the implementation.

The second case with just Map wouldn't select the proper item in the list right now.

I know the third case could be solved in a similar way to how Vector completion works in AS3, but I really believe in this being a better solution as the developer has more power of choice and a better view on what he selects.

I removed the code in CompletionList.Select because in the long run it doesn't make sense and was part of the problem with #71.

I checked AS3 Vector works fine, anyway you can check the build from the CI server.

SlavaRa:
@Neverbirth please check:
AS3
```var a:Array/*of int*/ = new |```
```var v:Vector.<*> = new |```

Neverbirth:
I think the second case is a problem in another part of code for what I already have a fix, but I'll take a look just in case. The first one is a missing case.

Neverbirth:
I see the first case also had problems back in the day, due to its special handling. It would probably make more sense to refactor that part and make it behave like Vector and Haxe generics, but will leave it for the moment as is.

I see Vector.<*> is a completely unrelated problem... it's been wrongly parsed for quite a long time (every version I've tried so far). I'll hack this one for now.

PR #1389, opened 2016-10-24
Wrap ShowDialog usages with using where it makes sense to better free resources

Neverbirth:
Middle-sized refactoring. Doesn't solve any problem, but former code was worse as it could keep some stuff in memory longer than really needed, as forms don't free resources when clicking on a "Close" button (unless we were manually calling Close in them) in modal mode.


Meychi:
@Neverbirth, could you check the conflicts?

PR #1391, opened 2016-10-24
Improve OutputPanel

### LINKED ISSUE #1148: Output panel text sometimes invisible until scrolling
I fairly frequently have the issue that the Output panel is completely blank after a build. As soon as I start scrolling however, the output shows up.
I think this generally only happens when the panel is not active (I usually have the Results panel active in the same "container" and then switch to Output after a build). Maybe a refresh is missing when switching the active panel?

Neverbirth:
- Fix resizing in panel. Fully fixes #1148.
- Do not enable scrollTimer by default, unneeded and sometimes when debugging could throw cross thread exceptions.
- Fix panel visibility when autoscroll was disabled. It was making wrong and unneeded operations.

I never realized how buggy the RichTextBox resizing/scrolling logic was... even using the latest RichEdit control that was released in Windows XP (newer than the one used in .NET) shows this problem. I don't know if it's some kind of hidden setting. The most obvious problem (but there are others, like the fixed problem): when you have the caret at the bottom of the control, while having a scrollbar, resize the control, the scrollbar will resize, and the text will move upwards leaving a lot of blank space below, when doing something the content will jump to the proper bounds. With this change everything seems to work in a more proper way.

Fixing this was a bit painful, and didn't find any fully managed way to do it. Another option could have been to use custom scrollbars, but it would require more logic to get the same behavior with probably bugs or worse performance, and still requiring interop.

### DIFF
--- a/External/Plugins/OutputPanel/PluginUI.cs
+++ b/External/Plugins/OutputPanel/PluginUI.cs
@@ -16,7 +16,7 @@ namespace OutputPanel
     public class PluginUI : DockPanelControl
     {
         private Int32 logCount;
-        private RichTextBox textLog;
+        private RichTextBoxEx textLog;
         private PluginMain pluginMain;
         private String searchInvitation;
         private System.Timers.Timer scrollTimer;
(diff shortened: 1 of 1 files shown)


Neverbirth:
After some testing I see this doesn't work right with custom scrollbars. But even without this patch they don't work as expected when resizing the RichTextBox...

I'll try to look into this soon.

Neverbirth:
Since the custom scrollbars have problem even without this change I'll merge this during this week.

PR #1406, opened 2016-11-27
Fixed fails after renaming when renaming function parameters in some cases. part2 #467

SlavaRa:
code for example:
```actionscript
package {
	import flash.display.Sprite;
	
	public class Main extends Sprite {
		
		public function Main(value:String) {
			v = value;
			trace(v);
		}
		
		private var v:String;
	}
}
```
Expected result after rename `value` to `v`:
```actionscript
package {
	import flash.display.Sprite;
	
	public class Main extends Sprite {
		
		public function Main(v:String) {
			this.v = value;
			trace(this.v);
		}
		
		private var v:String;
	}
}
```
Actual result after rename `value` to `v`:
```actionscript
package {
	import flash.display.Sprite;
	
	public class Main extends Sprite {
		
		public function Main(v:String) {
			this.v = value;
			trace(this.v);
		}
		
		private var this.v:String;// <-- BUG: this.v should be v
	}
}
```

### DIFF
--- a/External/Plugins/CodeRefactor/Commands/Rename.cs
+++ b/External/Plugins/CodeRefactor/Commands/Rename.cs
@@ -240,10 +240,11 @@ namespace CodeRefactor.Commands
                 var targetMatches = entry.Value;
                 if (isParameterVar)
                 {
-                    var replacement = string.Empty;
                     var lineFrom = Target.Context.ContextFunction.LineFrom;
+                    var lineTo = Target.Context.ContextFunction.LineTo;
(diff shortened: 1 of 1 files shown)


Neverbirth:
Unit tests?

SlavaRa:
nope(( because `BackgroundWorker`(in `FRRunner`) does not work when it run in unit test

Neverbirth:
Googling around suggests BackgroundWorker is testable, but I would prefer to see some abstraction for this finding refererences logic.

Well, merging.

PR #1409, opened 2016-11-28
[Haxe 3.3.0] Support of `--wait stdio`

SlavaRa:
https://github.com/HaxeFoundation/haxe/pull/5188

### DIFF
--- a/PluginCore/PluginCore/Resources/zh_CN.resx
+++ b/PluginCore/PluginCore/Resources/zh_CN.resx
@@ -1,17 +1,17 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <root>
-  <!--
-    Microsoft ResX Schema
-
+  <!-- 
+    Microsoft ResX Schema 
+    
     Version 2.0
-
-    The primary goals of this format is to allow a simple XML format
-    that is mostly human readable. The generation and parsing of the
-    various data types are done through the TypeConverter classes
+    
+    The primary goals of this format is to allow a simple XML format 
+    that is mostly human readable. The generation and parsing of the 
+    various data types are done through the TypeConverter classes 
     associated with the data types.
-
+    
     Example:
-
+    
     ... ado.net/XML headers & schema ...
     <resheader name="resmimetype">text/microsoft-resx</resheader>
     <resheader name="version">2.0</resheader>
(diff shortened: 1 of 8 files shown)


Neverbirth:
This should be configurable, because keeping the old method of working is useful for getting compile information and in some minor cases to reuse the same haxe instance.

SlavaRa:
Okey.
`Enable stdtio with false by default` or `Disable stdio with false by default` or another variant?

Neverbirth:
I'd go with the first one because it's the classical behavior, but I don't care that much really. It would be more for @elsassph and @Meychi to decide maybe.

elsassph:
We're still using Haxe 3.1.3 here; backward compatibility is important. You should be able to get the compiler version I think.

elsassph:
And it should probably be possible to disable it in general (who knows what problems people can run into - we also need to disable the compiler server)

elsassph:
That's very cool though!

SlavaRa:
@elsassph `Enable stdtio with false by default` or `Disable stdio with false by default` or another variant?

Neverbirth:
@elsassph Unless I miss something backwards compatibility is already kept. And since this is using the compiler server, it can be disabled in the same way as the old server.

elsassph:
I'd enable by default (bool prop to disable it so default value is false) if you properly detect that it is supported by the compiler.

SlavaRa:
I added the setting

Neverbirth:
Will review today (hope not to forget!), but if someone else reviews and merge before than I do I'm OK with it.

SlavaRa:
@Neverbirth thanks

PluginCore/PluginCore/Resources/zh_CN.resx
```diff
@@ -6112,4 +6112,8 @@ AND - All conditions are met.</value>
     <value>Defines which method should be used to clean the contents of the output panel.</value>
     <comment>Added after 5.2.0</comment>
   </data>
+  <data name="HaXeContext.Description.CompletionServerWaitStdio" xml:space="preserve">
+    <value>Makes haxe completion server use stdin and stderr instead of sockets for communication with IDE. (only applies to CompletionServer Completion Mode).</value>
```
Neverbirth: Not that my English is perfect, but I'd say:

Makes Haxe completion server to use stdin and stderr instead of sockets for communication with the IDE (only applies to CompletionServer Completion Mode and Haxe 3.3+).

Neverbirth:
Is this PR really complete? From what I read of this new mode it looks like it isn't. `CompletionServerCompletionHandler` still behaves the same way, so it's using TCPClient to send queries instead of stdin, and haxeProcess_ErrorDataReceived should be changed to read the compiler results. Isn't it like this?

PluginCore/PluginCore/Resources/zh_CN.resx
SlavaRa: ok, I will fix it today

Neverbirth:
So what's the plan on this?

SlavaRa:
I rediscovered PR today or tomorrow

PR #1414, opened 2016-12-02
 Added `Haxe - HashLink Project`. Fixes #1400

### LINKED ISSUE #1400: Add HashLink support
Using `haxe -hl ouput.hl` 
Default test with `hl.exe $outputfile` (should be in PATH)

You can pass window size as `-D windowSize=1280x768` (and store it in HXML as well if generated, same as SWF target which generates swf-header), it will be used as a standard on the platform.

SlavaRa:
Fixes #1400


ncannasse:
Can we have a solution that does not require the creation of a .bat ? That would be nice for neko too (we should allow toggle on/off prompt before exit at project level)

SlavaRa:
@Meychi @Neverbirth @elsassph ?

Neverbirth:
I'm OK with it, and I already did some work that would help (but is not finished nor submitted anywhere :/, I showcased it in WWX), but this is a quite subjective topic. Which kind of behavior and/or user experience would you exactly expect?

ncannasse:
When target is neko/hl, the "Test Project" would be set by default to "Run" with a checkbox (default ON) "Prompt before exit".

When running the project with this configuration, it would either run "neko $(output)" or "hl $(output)".

If "Prompt before exit" is checked, it would do that by generating a .bat in Windows/Temp:
 - if the command is successful, print "The program execution was successful"
 - if the error code with not 0, "The program exit with error code XXX"

Additionally you could also specify start directory (relative to project) and commandline arguments

Neverbirth:
Looks like something completely possible with the stuff I was doing... will try to resume it "soon".

PR #1422, opened 2016-12-20
Use `Search->Back From Declaration` instead of `Seach->Back From Type Declaration`

SlavaRa:


SlavaRa:

PR #1423, opened 2016-12-21
Fixes #1370

### LINKED ISSUE #1370: Find All References - wrong results if function and param share same name
Hello,

It's a small bug, the "Find All References" feature makes mistakes about internal and public scopes.

Example :

```
File 1:
public function Main () : void {
    randomName(1);
}

File 2:
public function randomName (randomName:int) : void { // try Find All References on function name here
    randomName++;
}
```

Results gives the 3 occurrences of "randomName" in file 2 while they are not referring to the same object, and no allusion to the call in file 1.

SlavaRa:
@elsassph @Neverbirth please review this PR


Neverbirth:
Will do it along with your other PR during this week.

Neverbirth:
It looks good, but sadly is not simple to fully verify as there are many cases to check. We should look into tests for these cases.

PR #1444, opened 2017-02-14
If we have an incomplete model set PrivateSectionIndex to + 1 like when having a proper model in ActionScript

Neverbirth:
This fixes issues where importing a type from the last line would add the import to a wrong location. Specially noticeable in Haxe when starting to write a new type at the end of the file.


Gama11:
Wow, I think this might fix that import insertion bug I got every so often and was never able to reproduce properly... :D

Issue #1154, opened 2016-03-31
hxml support: classpaths after --next not added


Gama11:
```
-cp src
--next
-cp test
```

Only `src` is added to the classpaths, which is annoying because I can't use QuickNavigate's type explorer to open test files. Of course this also leads to completion not working properly in those files.

I know `--next` isn't supported properly (#455), but FD could still add all the classpaths for convenience anyway (don't think there's any harm in that).

Issue #455, opened 2014-10-21
Haxe multi-targets


ncannasse:
Haxe allows several compilations to be performed in a single command by using `--next`

For instance the following HXML builds 3 different targets:

```
-swf out.swf
-main Main
-lib mylib

--next
-js out.js
-main Main
-lib mylib
-D fdLaunch=index.html

--next
-cpp bin
-main Main
-lib mylib
```

FlashDevelop could split the different compilations ocurring between `--next`, analyse the target which is compiled, and propose it as part of a drop down menu similar to NME target one.

In the previous HXML that would be:

```
All (default, current behavior)
Flash
JavaScript
C++
```

In case the same target is used in several parts of the compilation, we could add the output file to differentiate between them (eg `C++ (bin)`)

When testing the project, if the target is set to `All` then we can use `Test Project` as we do already.

However if we have a target selected or if our Test Project is set to OpenDocument with documentName="", we can use the following behavior:
- if we have `-D fdLaunch=<file>` we can use that as OpenDocument
- if not, we use default behavior on a per-target basis:
- on Flash, play the output SWF with Flash Player
- on C++, run the `<output-dir>\<main-class>[-debug].exe`
- on other targets print an error saying missing `-D fdLaunch=<document>`

SlavaRa:
via #2663

PR #1445, opened 2017-02-14
[Haxe] Parse `-lib` and `-cp` after `-next`. fixes issue #1154

SlavaRa:


Neverbirth:
I think there were some reasons against this. @elsassph ?

SlavaRa:
My working project consists of a single file .hxml, for several targets.
Running the specific target occurs via gradle.
I have a special platform for such projects(Settings\Platforms\Haxe\).
But I can not work comfortably due to this issue. So I have a separate FD project for each platform and i have to switch between them

elsassph:
The rationale was that it could cause problems because conflicting libs/cps could cause compiler errors during completion if your different compilations target different platforms.

This limitation was meant to be lifted by allowing the user to choose one of the targets of the hxml when multiple are available.

Neverbirth:
Oh, yeah. Right.

So I would prefer to see [this feature request](https://github.com/fdorg/flashdevelop/issues/455) implemented rather than this. In fact, it's on my todo, but I would be glad if someone is able to look into it before heh.

PR #1454, opened 2017-02-18
Assign statement of callback to var local variable

SlavaRa:
Before fix:
```actionscript
function test() {
  foo | <-- cursor here
}
function foo():Number {return 1;}
```
Result
```actionscript
function test() {
  var f:Function = foo
}
function foo():Number {return 1;}
```
Actual for the current release


Neverbirth:
I guess the Result you typed here is wrong, isn't it? I guess you meant:

Result
```actionscript
function test() {
  var f:Function = foo
}
function foo():Number {return 1;}
```

The code and tests also point to it, but I want to double confirm before merging.

SlavaRa:
Oh.. Yes you are right

Issue #1452, opened 2017-02-18
cpp.Void prioritized over Void


Yanrishatum:
When targeting C++ target in Haxe, autocomplete prioritizing `cpp.Void` completion over `Void`, resulting in autogeneration of `cpp.Void` import, and breaking compilation as `cpp.Void` is a return type.   
Current workaround:
Write `Void` and the press escape to cancel autocomplete.

Visual representation of the problem:
https://i.imgur.com/WNgcehA.gif
![example](https://i.imgur.com/WNgcehA.gif)

Neverbirth:
Maybe you would like to try the build for [this PR](https://github.com/fdorg/flashdevelop/pull/1346) and give your feedback to include it or further improve it:

https://ci.appveyor.com/project/FlashDevelop/flashdevelop-devel/build/5.2.1.19-hsflhfjf/artifacts

Yanrishatum:
@Neverbirth Yes, this PR do fix that problem. I've talked with Slava about this problem last year, what actually resulted in that PR. Talked with him right now, and he said he'll fix merge conflicts soon.
![img](https://i.imgur.com/f7tyuVw.png)

PR #1455, opened 2017-02-18
Fixes #812 and #1452

### LINKED ISSUE #812: Auto class import priority for global class

SlavaRa:


Yanrishatum:
Tested the DLL (Slava sent me compiled version before PR) and everything works properly.

PR #1462, opened 2017-03-02
Fixed generation and override of functions with type parameters.

Neverbirth:
Import of types used in constraints and spaces between params may be removed, but it's not considered as important right now.

@SlavaRa  Any concern with this?


SlavaRa:
Looks good

PR #1473, opened 2017-03-13
If the target we are trying to rename exists as a private class or interface or enum or typedef or abstract we only need to search the current file

SlavaRa:


SlavaRa:
@Neverbirth @elsassph please review this PR

SlavaRa:
@Neverbirth thanks

PR #1459, opened 2017-02-23
Fixes issue #1387

SlavaRa:
I'll add tests after the merger #1458

PR #1478, opened 2017-03-14
Fixes issue #1459

SlavaRa:


Neverbirth:
Personally I would prefer some property of ContextFeatures, as we have for other stuff like dot, hasGenerics, etc, etc. I don't know what others like @elsassph thinks.

SlavaRa:
Well, I'll try to do otherwise

SlavaRa:
I think we could add a list of postfix operators to ContextFeatures. @Neverbirth what do you think about that?

Neverbirth:
I would be ok with it.

Issue #1463, opened 2017-03-07
Improve haxe errors parsing


yar3333:
Sometimes haxe output error message in several lines:
```
src/HaqComponent.hx:8: characters 7-16 : You cannot access the js package while targeting neko (for js.JQuery)
src/HaqComponentTools.hx:7: characters 7-35 :     referenced here
src/HaqTemplateManager.hx:9: characters 7-40 :     referenced here
src/Build.hx:68: characters 34-52 :     referenced here
```

FlashDevelop split this into several lines (this is not a problem) and sort them (this is not good!):
![1](https://cloud.githubusercontent.com/assets/900330/23678648/12f643ae-0396-11e7-9918-5f4fad991f53.png)

Neverbirth:
Actually this is already planned from my side.

Neverbirth:
In progress by a colleague.

PR #1479, opened 2017-03-16
Result grouping

chipshort:
Allows to group the items in the ResultsPanel by other columns than file.

screenshot1
screenshot2
screenshot3


SlavaRa:
Looks like good, but please use int instead of Int32(you have Int32 and int)

Neverbirth:
The PR is still missing a thing plus there is a couple of details to fix.

chipshort:
I just changed the behaviour to show no grouping if you click the column for the third time. Instead everything is shown in the order it showed up in the output:

nogrouping

Neverbirth:
This last change should fix #1463.

PR #1484, opened 2017-03-17
Remove bad code

SlavaRa:
@Neverbirth please review this PR and merge him :)

### DIFF
--- a/External/Plugins/StartPage/PluginMain.cs
+++ b/External/Plugins/StartPage/PluginMain.cs
@@ -125,7 +125,7 @@ namespace StartPage
                     {
                         case ProjectManagerEvents.Project :
                             // Close pluginPanel if the user has the setting checked and a project is opened
-                            if (de.Data != null & this.startPage != null)
+                            if (de.Data != null && this.startPage != null)
                             {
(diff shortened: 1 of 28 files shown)


External/Plugins/StartPage/PluginMain.cs
```diff
@@ -128,15 +128,12 @@ public void HandleEvent(Object sender, NotifyEvent e, HandlingPriority priority)
                             if (de.Data != null & this.startPage != null)
```
Neverbirth: de.Data != null & this.startPage != null this should be && heh.

SlavaRa: Fixed

Neverbirth:
Please, revert the change to PluginCore/BBCode/BottomUpParser.cs and all the ones inside PluginCore/DockPanelSuite/*.

Since they don't really add improvements it's better to avoid those ones for possible future improvements (like updating DockPanelSuite), and because BBCode from my side is removed and hopefully I can see it replaced soon...

SlavaRa:
One moment

SlavaRa:
Done

PR #1485, opened 2017-03-17
[AppMan] Update FlashPlayer and AIR to version 25

SlavaRa:
Guys please create http://www.flashdevelop.org/appman/frameworks_flexsdk_9-25.zip
http://www.adobe.com/support/flashplayer/debug_downloads.html


Meychi:
The file is now in place. Could you point the links also to that file and i will merge it then.

SlavaRa:
Done

PR #1489, opened 2017-03-21
Add EditorConfig and Yeoman plugin to appman

chipshort:


Neverbirth:
Ampersands need to be escaped in the link to the forum, otherwise the xml file breaks.

PR #1495, opened 2017-03-24
[Haxe] Updade FileParser. part 1.

SlavaRa:


External/Plugins/ASCompletion/Model/ASFileParser.cs
```diff
@@ -1112,6 +1112,16 @@ public void ParseSrc(FileModel fileModel, string ba, bool allowBaReExtract)
                     else if (inType)
                     {
                         foundColon = false;
+                        if (haXe && c1 == '>' && ba[i - 2] == '-' && inAnonType)
```
Neverbirth: Since the first condition is haxe, it looks like this if can be moved some lines down.

SlavaRa: one minute please

SlavaRa: fixed

PR #1496, opened 2017-03-25
Fixes NPE after renaming class name from ProjectManager. (fixes #1464)

### LINKED ISSUE #1464: Refactor rename class error
Sometimes after renaming a class I get this error.  I don't notice the actual refactors/renames was failed, I just get this message.

image

I use a quite recent dev build (https://github.com/fdorg/flashdevelop/commit/120f067995):
image

SlavaRa:


External/Plugins/CodeRefactor/Provider/ICommandFactory.cs
```diff
@@ -122,23 +122,23 @@ public interface ICommandFactory
         /// Outputs found results.
         /// Uses the current text location as the declaration target.
         /// </summary>
-        Command CreateRenameCommand();
+        Command CreateRenameCommandAndExecute();
```
Neverbirth: This change feels strange. Why AndExecute? Is the command executed by itself? At least in the Factory itself is not run. I would say it's not right if this command behaves in a way all the other commands in a different one.

SlavaRa: Yes, it starts execution right after the creation. This appeared after adding "Inline renaming". I'll fix it, but it takes time

Neverbirth:
Conflicts now.

SlavaRa:
5 minutes please

SlavaRa:
fixed

External/Plugins/CodeRefactor/Provider/ICommandFactory.cs
Neverbirth: Ok, I will merge it, but I would say it definitely needs to be changed.

SlavaRa: I completely agree with you, I will correct in the near future

PR #1497, opened 2017-03-25
Added `ContextFeatures.SpecialPostfixOperators`. (fixes #1475)

### LINKED ISSUE #1475: Add version check for PR #1459
Reminder @SlavaRa 

Quite probably having some ContextFeatures property that is enabled if Haxe >= 3.3.

SlavaRa:


External/Plugins/HaXeContext/Context.cs
```diff
@@ -241,13 +241,15 @@ public void SetHaxeEnvironment(string sdkPath)
...
+            path = path.TrimStart(';');
             path = sdkPath + ";" + path;
-            if (neko != null) path = neko.TrimEnd(new char[] { '/', '\\' }) + ";" + path;
+            if (neko != null) path = neko.TrimEnd('/', '\\') + ";" + path;
             Environment.SetEnvironmentVariable("PATH", path);
             currentEnv = sdkPath;
 
             LoadMetadata();
+            if (GetCurrentSDKVersion().IsGreaterThanOrEquals(new SemVer("3.3.0"))) features.SpecialPostfixOperators = new[] {'!'};
+            else features.SpecialPostfixOperators = new char[0];
```
Neverbirth: It seems the else is not needed since it's already initialized with an empty array in ContextFeatures itself, isn't it?

External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -3496,22 +3497,18 @@ private static ASExpr GetExpression(ScintillaControl sci, int position, bool ign
...
-                                expression.Separator = ' ';
-                                break;
-                            }
-                            genCount--;
+                            position--;
+                            expression.Separator = ' ';
+                            break;
                         }
-                        else sb.Insert(0, c);
+                        genCount--;
```
Neverbirth: Why this changes in this `else if` condition? Seems out of the scope of this PR. What does it improve?

External/Plugins/HaXeContext/Context.cs
SlavaRa: Due to the possibility of switching to a version of the SDK version below 3.3.0

External/Plugins/ASCompletion/Completion/ASComplete.cs
SlavaRa: Sorry, Apparently accidentally got into this commit, but in any case, in the construction ```else if (hasGenerics && (genCount> 0 || c == '<'))``` genCount can not be greater than zero

External/Plugins/HaXeContext/Context.cs
Neverbirth: Ah, I see.

PR #1499, opened 2017-03-26
[UnitTests][CodeRefactor] Small cleanup...

SlavaRa:


Tests/External/Plugins/CodeRefactor.Tests/Commands/CodeRefactorTests.cs
```diff
@@ -17,9 +17,9 @@ namespace CodeRefactor.Commands
     [TestFixture]
     class CodeRefactorTests
     {
-        private MainForm mainForm;
-        private ISettings settings;
-        private ITabbedDocument doc;
+        MainForm mainForm;
+        ISettings settings;
+        ITabbedDocument doc;
```
Neverbirth: I prefer to keep the private. This is not Haxe :P, in .NET the standard is to specify. Maybe we should look into CodeClimate or similar.

PR #1510, opened 2017-03-31
Create Manager to add more operations to BatchProcessDialog

chipshort:


External/Plugins/CodeRefactor/Controls/BatchProcessDialog.cs
```diff
@@ -232,6 +242,10 @@ private void DoProcess(ITabbedDocument document)
                     document.SciControl.ConvertEOLs(document.SciControl.EOLMode);
                     break;
                 }
+                default: // Custom BatchProcessor
+                    var i = this.operationComboBox.SelectedIndex - 4;
```
Neverbirth: This sounds quite error prone heh.

Neverbirth:
@SlavaRa ?

I wouldn't make the BatchProcessManager with all static methods... but since it's quite basic stuff I'm not sure it would bring limitations or problems when writing possible tests in the future?

I would also rewrite the current batch processors implementing this new stuff, but maybe not so important for a first PR.

External/Plugins/CodeRefactor/Controls/BatchProcessDialog.cs
SlavaRa: yes it is no good

SlavaRa:
@Neverbirth I think that now it is not bad, but it is very likely to be remade in the future

chipshort:
@Neverbirth I rewrote the current processors to use the new stuff. What do you think?

Neverbirth:
Uhm... this is missing something important: adding and removing processors at any moment on runtime. It wouldn't be strange things like wanting to do it depending on project types. Anyway, I will merge so you can go on with your stuff.

chipshort:
Well, you can add processors any time unless the BatchProcessDialog is already opened. For removing a processor depending on the current project, I would use the IsAvailable property. However, being able to actually remove a Processor is probably needed too, so I will add that.

PR #1512, opened 2017-03-31
[Haxe][ASGenerator] Improved callback code generation. (fixes #103)

### LINKED ISSUE #103: [haxe] callback code generation
It would be cool if we can get improved code generation on callbacks in Haxe
Lets say you have this class

```
class MyClass {
   public function onClick(event:Event->Void)
   {

   }
}

// implementation somewhere:
var instance = new MyClass();
instance.onClick(onInstanceClick // ctrl-shift-1 > create private function here!
```

gives this:

```
function onInstanceClick () {

}
```

I'd love to see this as:

```
function onInstanceClick (:Event):Void {

}
```

Optional, it would be even more cool to have some kind of documentation-rule for the parameter names

```
/**
* @param event (myNameHere:Event)->Void
*/
public function onClick(event:Event->Void)
   {

   }
```

So the function could be generated like this:

```
function onInstanceClick (myNameHere:Event):Void {

}
```

SlavaRa:


SlavaRa:

SlavaRa:
@Neverbirth what do you think about this?

Neverbirth:
It's Saturday 9pm and I'm doing stuff for work XD. But I'm sure this will get merged soon.

External/Plugins/ASCompletion/Completion/ASGenerator.cs
```diff
@@ -2815,12 +2816,114 @@ private static void GenerateFunctionJob(GeneratorJobType job, ScintillaControl s
...
+                                parameter = $"?{parameter}";
+                                type = type.TrimStart('?');
+                            }
+                            functionParameters.Add(new FunctionParameter(parameter, type, type, callerExpr));
+                            startPosition = i;
+                        }
+                    }
+                    newMemberType = s.Substring(endPosition);
+                }
+            }
```
Neverbirth: Some notes:
- It would make sense to have this extracted outside in a reusable way so others could also parse function types too. 
- I have the feeling `while (pos > 0)` can be optimized. Although it should never reach the exit condition I think.
- This wouldn't cover cases with generics like `function test<K>(arg:K->String):Void`, where you would like to create a new function also with the `<K>` type parameter, but step by step.

Anyway, good to merge I would say.

SlavaRa: Good comments
I will try to improve this functionality in the near future
Thanks

PR #1519, opened 2017-04-01
AddClosingBraces fixes

### LINKED ISSUE #1514: Error when editing AddClosingBracesOptions
Editor is trying to add/remove/edit a fixed `Array` instead of a `List<>`.

image

### LINKED ISSUE #1515: Closing paren won't overwrite after string - AddClosingBraces
With AddClosingBraces on, even with the correct options, a closing brace won't overwrite after a string.

image

wise0704:
Closes #1514
Closes #1515


Neverbirth:
Would be interesting to have tests for HandleAddClosingBraces.

Tests/External/Plugins/ASCompletion.Tests/Completion/ASCompleteTests.cs
```diff
@@ -606,6 +606,155 @@ public ComaExpression Haxe(string text)
...
+                sci.SetSel(cursor, cursor + 1);
+                sci.ReplaceSel("");
+                sci.Colourise(0, -1);
+                int c = addedChar ? sci.CharAt(cursor - 1) : sci.CurrentChar;
+                ASComplete.HandleAddClosingBraces(sci, (char) c, addedChar);
+                if (!addedChar)
+                {
+                    sci.DeleteBack(); // Backspace is handled after HandleAddClosingBraces(), so mimic that behaviour
+                }
+                return sci.GetCurLine(sci.CurLineSize - 1); // Ignore the last new line character
```
wise0704: Needs #1530 for this line to work properly

wise0704:
AS3 tests are running really slowly. `ASContext.Context.SetAs3Features()` takes around 300ms to run on my computer. Perhaps I should only have the Haxe context since the context wouldn't really matter?

Tests/External/Plugins/ASCompletion.Tests/Completion/ASCompleteTests.cs
```diff
@@ -719,17 +719,23 @@ public IEnumerable<TestCaseData> TestCases
             [Test, TestCaseSource(nameof(TestCases))]
             public string AS3(string text)
             {
-                sci.ConfigurationLanguage = "as3";
-                ASContext.Context.SetAs3Features();
-                return Common(text, sci);
+                if (sci.ConfigurationLanguage != "as3")
```
SlavaRa: What is it?

SlavaRa: Maybe I'd better speed up the `Set*Features` method?

wise0704: Nah I think it's fine. Those methods took around 300ms per run, and in practice, the context doesn't change this frequently and I believe `Set*Features()` aren't called this frequently either.

Tests/External/Plugins/ASCompletion.Tests/Completion/ASCompleteTests.cs
```diff
@@ -606,6 +606,162 @@ public ComaExpression Haxe(string text)
...
+                    yield return new TestCaseData("\"\"(+\"\""). Returns("\"\"(\"\"").SetName(prefix + "Open ( after 'String' before 'String'");
+                    yield return new TestCaseData("\"\")+)").    Returns("\"\")").    SetName(prefix + "Close ) after 'String' to overwrite )");
+                    yield return new TestCaseData("\"\")+)\"\"").Returns("\"\")\"\"").SetName(prefix + "Close ) after 'String' before 'String' to overwrite )");
+                    yield return new TestCaseData("\"\"(-)").    Returns("\"\"").     SetName(prefix + "Delete ( after 'String' to delete )");
+                    yield return new TestCaseData("\"\"(-)\"\"").Returns("\"\"\"\""). SetName(prefix + "Delete ( after 'String' before 'String' to delete )");
+
+                    yield return new TestCaseData("\"(+\""). Returns("\"(\""). SetName(prefix + "Open ( inside a string");
+                    yield return new TestCaseData("\")+)\"").Returns("\"))\"").SetName(prefix + "Close ) inside a string");
+                    yield return new TestCaseData("\"(-)\"").Returns("\")\""). SetName(prefix + "Delete ( inside a string");
```
Neverbirth: The formatting looks ugly to me. But if @SlavaRa is ok I will merge.

SlavaRa: For me as well) but I think that we can then bring everything to one kind

wise0704: If you're talking about the `\"` in particular, I agree. I had some trouble writing those strings myself.

Neverbirth: From my side I was talking about having everything in a single line and trying to have in the same column the different method calls but then using different columns for different blocks of code.

wise0704: Ah, should I have them all in the same column or just no spaces at all?

wise0704:
I'll me modifying and adding some tests in #1520, so if there isn't any problem with the first two commits (a3e35e5 and 3be4db4), we can merge this and continue in #1520.

Neverbirth:
This has conflicts

Tests/External/Plugins/ASCompletion.Tests/Completion/ASCompleteTests.cs
SlavaRa: You could make a many sets of test cases, instead of using regions
like as https://github.com/SlavaRa/fdplugin-postfix-code-completion/blob/master/PostfixCodeCompletionTests/Completion/PostfixGeneratorTests.cs#L668

PR #1520, opened 2017-04-01
Improved automatic closing brace insertion rules & settings interface.

wise0704:
Users are able to set multiple rules per set of braces, with a better interface.

**Edit:** updated image
image


External/Plugins/ASCompletion/Controls/AddClosingBracesRulesEditorForm.cs
```diff
@@ -0,0 +1,804 @@
...
+        #endregion
+
+        #region Initialization
+
+        private void InitializeGraphics()
+        {
+            colDelete.Image = PluginBase.MainForm.FindImage16("153", false);
+            btnUp.Image = PluginBase.MainForm.FindImage16("74", false);
+            btnDown.Image = PluginBase.MainForm.FindImage16("60", false);
+            btnAddRule.Image = PluginBase.MainForm.FindImage16("33", false); // Seriously no plus icon?
```
wise0704: Can we get a better plus icon?

Neverbirth: I guess if you find free to use pictures and they are nice @Meychi will be ok with adding them.

Neverbirth:
I see the picture and I don't understand what are the fields in that grid ^^U.

wise0704:
Yeah, that configuration has always been so hard to make it fully customisable and also user friendly at the same time.

Basically, you read everything from the column headers. For example from the picture above:
After (no character matches) **Or** After { Type } **And** Not Before { '<' } **And** Not Before { Identifier, Type }
or
After { '.' } **And** After { Operator } **And** Not Before { '<' } **And** Not Before { Identifier, Type }

Basically the 1st rule matches for `Type<|` not followed by '<', identifier or type. Result is `Type<|>`. | is the cursor.
2nd rule matches for `Type.<|` not followed by '<', identifier or type. Note that since "After '.' AND After Operator", it doesnt match for `100.<`.

wise0704:
Needs to wait for #1519 and after resolve merge conflicts.

wise0704:
With ae28d74 I've added a green plus image to Images.png and Images32.png (drew myself):

16x16
image

32x32
image

@Meychi

Neverbirth:
We should write some wiki entry explaining how to setup this stuff.

PR #1521, opened 2017-04-01
Improved SurroundWith feature

wise0704:
Instead of attaching it into a ToolStripMenu, make it into a list of CompletionListItem.

Now you can set a shortcut for the SurroundWith feature (default is None).

image

image


PluginCore/PluginCore/Controls/CompletionList.cs
```diff
@@ -364,7 +364,7 @@ static private void CLDrawListItem(Object sender, System.Windows.Forms.DrawItemE
                 g.InterpolationMode = System.Drawing.Drawing2D.InterpolationMode.HighQualityBicubic;
                 g.DrawImage(item.Icon, 1, e.Bounds.Top + ((e.Bounds.Height - newHeight) / 2), newHeight, newHeight);
                 int p = item.Label.LastIndexOf('.');
-                if (p > 0 && !selected)
+                if (p > 0 && !selected && !(item is ICompletionListSpecialItem))
```
wise0704: Added a new interface instead of adding a new property to the existing interface, since that will cause all other classes implementing it to break.

SlavaRa:
Looks good

Neverbirth:
I would like to check that if in `CompletionList` more in depth before merging. Will try to do soon...

Meychi:
@Neverbirth, if something needs fixing, PR them :)

PR #1522, opened 2017-04-01
Extended Shortcuts

### LINKED ISSUE #1198: F2 as default shortcut for Refactor Rename
Refactor>Rename is a super useful command and mostly bug free. Why is F2 not the default shortcut for the command? Can it be added or is it hard since its exposed by a plugin and not part of the default command set?

### LINKED ISSUE #1302: Advanced Shortcut - default main menu shortcuts not working

wise0704:
Manually resolving merge conflicts of 600+ files was definitely not fun.

Closes #1198
Closes #1302


wise0704:
Previously there had been a suggestion, that when the input of the first key sequence does not match any first parts of all shortcuts, FD should not register the second key sequence as the second part but rather as a new sequence. 

I have not implemented it nor think it should be the default behaviour for 2 reasons:

 - It can hinder (although minor) performance. It means looping through the list of all shortcuts for every unhandled first sequence inputs.
 - It can be confusing for the users and cause unexpected behaviour. For example, supposed the user wants to press `Ctrl+E, Ctrl+S`, but presses `Ctrl+R, Ctrl+S`. If `Ctrl+R` does not exist, it will unexpectedly invoke `Ctrl+S`. In my opinion, status bar saying `Ctrl+R, Ctrl+S` sequence does not exist is more expected.

So I'm thinking, that it is rather better to be added as a setting. What are your thoughts?

Neverbirth:
For me the user experience is not really any better, and makes the code more complex and therefore more error prone.

wise0704:
Since next is a major upgrade to 6.0, may I just break backwards compatibility, such as method signatures and property types?
It's extremely bug-prone and unnecessarily complicated trying to keep the backwards compatibility.

Btw this PR is quite broken atm. Fixes depend on the answer to above.

FlashDevelop/MainForm.cs
```diff
@@ -192,17 +194,28 @@ public ToolStrip ToolStrip
...
         }
 
         /// <summary>
+        /// Sets the text of the <see cref="StatusLabel"/>.
+        /// Use this method instead of directly accessing the <code>Text</code> property of <see cref="StatusLabel"/>.
+        /// </summary>
+        public string StatusLabelText
+        {
+            get { return this.StatusLabel.Text; }
+            set { if (!lockStatusLabel) StatusLabel.Text = value ?? " "; }
```
wise0704: @elsassph https://github.com/fdorg/flashdevelop/pull/1245#issuecomment-225375273
>when starting a shortcut (eg. Ctrl+E), I see in the status bar that FD is waiting for another key; but if I click in the editor (to cancel the sequence), I see a strange relayout whereby the Scintilla editor grows over the bottom tabs and quickly goes back to normal size.

Apparently in HDPI (I assume) having 0-length text (`null` gets converted to `""`) in status label resizes the control.
`value ?? " "` fixes the issue, while still allowing the option to set 0-length text that resizes the control.

An alternative is just `string.IsNullOrEmpty(value) ? " " : value`.

wise0704:
77e8e2e is like a preview of what will happen if backward compatibility is to be dropped:
- Properties marked with `[Obsolete("Type is ToolStripMenuItemEx")]` will have their types changed.
- Usage of above properties that required casting to `ToolStripMenuItemEx` will have the casting removed.
- Methods marked with `[Obsolete("This method has been deprecated.", true)]` will be removed.

wise0704:
This should wait for #1578, since it's easier to resolve conflicts from this PR.
Also this PR is not fully completed - needs more reviews, mainly regarding the backward compatibility.

wise0704:
maintenance not possible

PR #1530, opened 2017-04-02
Small fixes

wise0704:
Fix ScintillaControl.CurLineSize and GetCurLine()
Change CommandFactoryProvider.GetFactoryFromLanguage to GetFactory in
CodeRefactorTests.cs

Required for #1519

### DIFF
--- a/PluginCore/ScintillaNet/ScintillaControl.cs
+++ b/PluginCore/ScintillaNet/ScintillaControl.cs
@@ -997,7 +997,7 @@ namespace ScintillaNet
         {
             get
             {
-                return SPerform(2027, 0, 0);
+                return SPerform(2027, 0, 0) - 1; //ignore the terminating null character
             }
         }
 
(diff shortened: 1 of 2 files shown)


PluginCore/ScintillaNet/ScintillaControl.cs
```diff
@@ -3190,10 +3190,9 @@ public void GotoPos(int pos)
         /// </summary>
         unsafe public string GetCurLine(int length)
         {
-            int sz = SPerform(2027, length, 0);
-            byte[] buffer = new byte[sz + 1];
```
wise0704: was meant to be `sz` without `+1`

PluginCore/ScintillaNet/ScintillaControl.cs
```diff
@@ -3190,10 +3190,9 @@ public void GotoPos(int pos)
...
+            byte[] buffer = new byte[length + 1];
             fixed (byte* b = buffer) SPerform(2027, length + 1, (uint)b);
-            return Encoding.GetEncoding(this.CodePage).GetString(buffer, 0, sz - 1);
+            return Encoding.GetEncoding(this.CodePage).GetString(buffer, 0, length);
```
Neverbirth: This could cause problems as the returned value is different now, I'm OK with the change but I think it would be better to do:

```
        unsafe public string GetCurLine(int length)
        {
            int sz = SPerform(2027, 0, 0);
            byte[] buffer = new byte[length + 1];
            fixed (byte* b = buffer) SPerform(2027, length + 1, (uint)b);
            return Encoding.GetEncoding(this.CodePage).GetString(buffer, 0, Math.Min(length, sz - 1));
        }
```

At least this way when we try to get more chars than available we won't get back null characters at the end. Which some plugins devs may not expect with high length values.

wise0704: Aha, you're right. 
Also I think it would be better to cap `length` to `sz - 1` at the start instead, so `SPerform` also doesn't waste memory&time by writing extra null characters that will be ignored.

PR #1535, opened 2017-04-03
Make IBatchProcessor removable

chipshort:
@Neverbirth Regarding #1510

### DIFF
--- a/External/Plugins/CodeRefactor/Managers/BatchProcessManager.cs
+++ b/External/Plugins/CodeRefactor/Managers/BatchProcessManager.cs
@@ -17,6 +17,14 @@ namespace CodeRefactor.Managers
         {
             processors.Add(processor);
         }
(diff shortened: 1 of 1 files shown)


Neverbirth:
@SlavaRa ? For me it looks like EventManager is not needed and I would go with just a normal class exposed by the CodeRefactor plugin and normal events. But you decide.

SlavaRa:
Please give me 30 minutes to study the code :)

SlavaRa:
I, too, do not see the point in these events, but maybe the @chipshort will tell we why they are needed?

chipshort:
@SlavaRa Well, I added the events because @Neverbirth asked for them.

Neverbirth:
Yeah, I think having events is useful as there may be the need to refresh information when processors are added or removed.

PR #1536, opened 2017-04-03
[Haxe][ASGenerator] Improved callback code generation.

SlavaRa:


External/Plugins/ASCompletion/Completion/ASGenerator.cs
```diff
@@ -2817,65 +2817,70 @@ private static void GenerateFunctionJob(GeneratorJobType job, ScintillaControl s
...
+                else if (c == ']') arrCount++;
+                else if (c == '[') arrCount--;
+                else if (c == '}') braCount++;
+                else if (c == '{') braCount--;
+                else if (c == '>') genCount++;
+                else if (c == '<') genCount--;
+                else if (dquCount == 0 && c == '"' && sci.CharAt(pos - 1) != '\\') dquCount++;
+                else if (dquCount > 0 && c == '"' && sci.CharAt(pos - 1) != '\\') dquCount--;
+                else if (quoCount == 0 && c == '\'' && sci.CharAt(pos - 1) != '\\') quoCount++;
+                else if (quoCount > 0 && c == '\'' && sci.CharAt(pos - 1) != '\\') quoCount--;
```
wise0704: Wouldn't this behave unexpectedly when there is an escaped backslash before the quotes like `\\"`?
`ASComplete` has `IsEscapedCharacter(ScintillaControl sci, int position)` method which might be useful if so.

SlavaRa: Just a moment

SlavaRa: @wise0704 thanks, fixed

PR #1540, opened 2017-04-04
Add LintingManager

chipshort:
The idea is to have one manager to handle linting / code analysis.
I would like to get some comments on this.
It would be worth considering if this belongs into CodeRefactor plugin, for example.
Other than that, what about the overall architecture?

An example for a plugin using it: https://github.com/DoclerLabs/hexDevelop/tree/haxecheckstyle/Plugins/HaxeCheckStyle

### DIFF
--- a/External/Plugins/LintingHelper/Helpers/DictionaryHelper.cs
+++ b/External/Plugins/LintingHelper/Helpers/DictionaryHelper.cs
@@ -1,0 +1,30 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace LintingHelper.Helpers
+{
+    class DictionaryHelper
+    {
+        /// <summary>
+        /// Helper method for a dictionary containing lists,
+        /// Returns either the list of the given <paramref name="key"/> or a new list, if the key does not exist.
+        /// </summary>
+        public static List<T> GetOrCreate<S, T>(Dictionary<S, List<T>> dict, S key)
+        {
+            List<T> list;
+            if (dict.ContainsKey(key))
+            {
+                list = dict[key];
+            }
+            else
+            {
+                list = new List<T>();
+                dict[key] = list;
+            }
+
+            return list;
+        }
+    }
+}
--- a/External/Plugins/LintingHelper/Helpers/DictionaryHelper.cs
+++ b/External/Plugins/LintingHelper/Helpers/DictionaryHelper.cs
@@ -5,20 +5,16 @@ using System.Text;
 
 namespace LintingHelper.Helpers
 {
-    class DictionaryHelper
+    static class DictionaryHelper
     {
         /// <summary>
         /// Helper method for a dictionary containing lists,
         /// Returns either the list of the given <paramref name="key"/> or a new list, if the key does not exist.
         /// </summary>
-        public static List<T> GetOrCreate<S, T>(Dictionary<S, List<T>> dict, S key)
+        public static List<T> GetOrCreate<S, T>(this Dictionary<S, List<T>> dict, S key)
         {
             List<T> list;
-            if (dict.ContainsKey(key))
-            {
-                list = dict[key];
-            }
-            else
+            if (!dict.TryGetValue(key, out list))
             {
                 list = new List<T>();
                 dict[key] = list;
--- a/External/Plugins/LintingHelper/Managers/LintingManager.cs
+++ b/External/Plugins/LintingHelper/Managers/LintingManager.cs
@@ -1,11 +1,8 @@
 ﻿using LintingHelper.Helpers;
 using PluginCore;
 using PluginCore.Managers;
-using System;
 using System.Collections.Generic;
-using System.IO;
 using System.Linq;
-using System.Text;
 
 namespace LintingHelper.Managers
 {
@@ -30,6 +27,8 @@ namespace LintingHelper.Managers
             {
                 list.Add(provider);
             }
+
+            EventManager.DispatchEvent(provider, new DataEvent(EventType.Command, "LintingManager.LinterRegistered", language));
         }
 
         /// <summary>
@@ -45,6 +44,8 @@ namespace LintingHelper.Managers
             {
                 linters[language].Remove(provider);
             }
+
+            EventManager.DispatchEvent(provider, new DataEvent(EventType.Command, "LintingManager.LinterUnregistered", language));
         }
 
         /// <summary>
@@ -87,6 +88,7 @@ namespace LintingHelper.Managers
                 linter.LintAsync(files, (results) =>
                 {
                     ApplyLint(files, results);
+                    EventManager.DispatchEvent(linter, new DataEvent(EventType.Command, "LintingManager.FilesLinted", files));
                 });
             }
         }
--- a/External/Plugins/LintingHelper/Managers/LintingManager.cs
+++ b/External/Plugins/LintingHelper/Managers/LintingManager.cs
@@ -65,12 +65,14 @@ namespace LintingHelper.Managers
         public static void LintFiles(string[] files, string language)
         {
             language = language.ToLower();
-            //PluginBase.MainForm.CallCommand("PluginCommand", "ResultsPanel.ClearResults"); //TODO: fix this
+            //PluginBase.MainForm.CallCommand("PluginCommand", "ResultsPanel.ClearResults"); //TODO: add stuff to output / results panel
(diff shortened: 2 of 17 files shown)


Neverbirth:
@SlavaRa ?

External/Plugins/LintingHelper/Managers/LintingManager.cs
```diff
@@ -0,0 +1,200 @@
...
+        public static void LintFiles(string[] files)
+        {
+            //detect languages
+            var filesByLang = new Dictionary<string, List<string>>();
+            foreach (var file in files)
+            {
+                var lang = PluginBase.MainForm.SciConfig.GetLanguageFromFile(file).ToLower();
+                var list = DictionaryHelper.GetOrCreate(filesByLang, lang);
+
+                list.Add(file);
```
SlavaRa: It looks very not obvious

External/Plugins/LintingHelper/Managers/LintingManager.cs
```diff
@@ -0,0 +1,200 @@
...
+        /// <param name="provider">the linter</param>
+        public static void RegisterLinter(string language, ILintProvider provider)
+        {
+            language = language.ToLower();
+
+            var list = DictionaryHelper.GetOrCreate(linters, language);
+
+            if (!list.Contains(provider))
+            {
+                list.Add(provider);
```
SlavaRa: It looks very not obvious

External/Plugins/LintingHelper/Managers/LintingManager.cs
```diff
@@ -0,0 +1,200 @@
...
+                    switch (result.Severity)
+                    {
+                        case LintingSeverity.Error:
+                            color = 0x00000080;
+                            break;
+                        case LintingSeverity.Warning:
+                            color = 0x00008080;
+                            break;
+                        case LintingSeverity.Info:
+                            color = 0x00008000;
```
SlavaRa: Maybe it makes sense to put it in the settings?

Neverbirth: Good catch, but I don't think it should be a setting, it should come from the theme or be derived from it.

External/Plugins/LintingHelper/Helpers/DictionaryHelper.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace LintingHelper.Helpers
+{
+    class DictionaryHelper
```
SlavaRa: I think it's better to change this class to an extension like as:
```csharp
static class DictionaryExtensions
{
    /// <summary>
    /// Helper method for a dictionary containing lists,
    /// Returns either the list of the given <paramref name="key"/> or a new list, if the key does not exist.
    /// </summary>
    public static List<TValue> GetValueOrCreate<TKey, TValue>(this Dictionary<TKey, List<TValue>> @this, TKey key)
    {
        if (!@this.ContainsKey(key)) @this[key] = new List<TValue>();
        return @this[key];
    }
}
```

Neverbirth: I think using `@this` is not very nice heh.

Please, do not use `dict.ContainsKey` in a case like this. Use `dict.TryGetValue`.

```
        List<T> list;
        if (!dict.TryGetValue(key, out list)) dict[key] = list = new List<TValue>();
        return list;
```

Separate that dict setting into 2 lines if you think it's cleaner, personally it's the same for me.

External/Plugins/LintingHelper/Managers/LintingManager.cs
chipshort: @Neverbirth I agree, but I am not sure how to do it.

External/Plugins/LintingHelper/Helpers/DictionaryHelper.cs
SlavaRa: Looks good

External/Plugins/LintingHelper/Managers/LintingManager.cs
Neverbirth: @Meychi Do we already have colors that we can reuse here? if not, which ones should we use for blending and/or deriving proper ones?

External/Plugins/LintingHelper/Managers/LintingManager.cs
```diff
@@ -0,0 +1,200 @@
...
+
+            return linters.ContainsKey(language) && linters[language].Count > 0;
+        }
+
+        /// <summary>
+        /// Gets the linters for a given <paramref name="language"/>.
+        /// </summary>
+        public static List<ILintProvider> GetLinters(string language)
+        {
+            if (linters.ContainsKey(language))
```
SlavaRa:
```csharp
return listeners.GetOrCreate(language)
```
 ?

chipshort: Not sure how I could overlook that one xD

SlavaRa: :)

SlavaRa:
@Neverbirth Very broadly written, but it looks good

chipshort:
Why are the tests failing?

chipshort:
I just noticed a problem with this. When the auto completion triggers, the file is saved. Therefore, the LintingManager runs the linters and adds the results to the log. But this hides the completion list, so code completion does not work correctly anymore.
@Neverbirth proposed it would be possible to add a reason or something similar to the Save event as a possible solution.

External/Plugins/LintingHelper/Managers/LintingManager.cs
Meychi: Output panel colors might be good for this.

Meychi: Also we have syntax theme settings that may match these better.

Neverbirth: Which ones?

SlavaRa:
@chipshort what do you think about c#6 and ReSharper?

External/Plugins/LintingHelper/Helpers/DictionaryHelper.cs
wise0704: It could've been in one line
```c#
return dict.TryGetValue(key, out List<T> list) ? list : dict[key] = new List<T>();
```

wise0704:
To be honest, I don't really know how to feel about an extension method that only targets specific type arguments.

Neverbirth:
It can be changed to nearest interface possible. It's true it's always better to do it like that, but at any rate it's not exposed publicly so I don't think it's too much of a problem.

chipshort:
@SlavaRa I'm not using ReSharper right now, but I know Jetbrains usually does great stuff. About C# 6, I think I like the null conditional operator the most. Why do you ask?

SlavaRa:
@chipshort I can give you a license for a ReSharper. I ask because your code can be done in 2-3 less)

chipshort:
@SlavaRa Ah, thank you, but it's okay. I should be able to get a student license. :)

SlavaRa:
JettBrains gives us(fdteam) licenses for all products

PluginCore/PluginCore/Resources/zh_CN.resx
```diff
@@ -6120,4 +6120,16 @@ AND - All conditions are met.</value>
...
+    <value>Runs the available linters when a file is saved.</value>
+	<comment>Added after 5.2.1</comment>
+  </data>
+  <data name="LintingHelper.Description.LintOnOpen" xml:space="preserve">
+    <value>Runs the available linters when a file is opened.</value>
+	<comment>Added after 5.2.1</comment>
+  </data>
+  <data name="LintingHelper.Label.RunLinters" xml:space="preserve">
+    <value>Run Linters</value>
+	<comment>Added after 5.2.1</comment>
```
wise0704: Current version is 5.2.0, isn't it?

chipshort: Made me wonder too, but @Neverbirth told me to put this in there. Maybe it was just a mistake. Could you clarify @Neverbirth ?

Neverbirth: Yep, if you check there are already a lot of entries like that heh. It was planned to make 5.2.1 happen, but sadly due to lack of time it never happened.

PR #1543, opened 2017-04-06
Add Images for Organise & Truncate Imports refactor menu

wise0704:
image

Plus just some minor code cleanup.

### DIFF
--- a/External/Plugins/CodeRefactor/Controls/RefactorMenu.cs
+++ b/External/Plugins/CodeRefactor/Controls/RefactorMenu.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Drawing;
 using System.Windows.Forms;
 using ASCompletion;
 using ASCompletion.Context;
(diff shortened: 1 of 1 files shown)


Neverbirth:
@Meychi up to you this one. I don't specially care about this heh.

Meychi:
Hmm... me either, @wise0704 whats the reasoning for this? As in VS?

wise0704:
Not really from VS, but it had been in my plan to add meaningful icons for some contextual refactor menu when I added those refactoring menus. I think it makes it look even more fancy. 

Imports are for packages, so I tried to find package icons before when I first added the refactor menus, but couldn't figure out where the icons were then haha. Organise&truncate packages are kind of separated from the rest of the menus.

Neverbirth:
For organize maybe there are more meaningful icons like the typical sort icon with A->Z inside or similar.

wise0704:
How about this?
image

Neverbirth:
Looks definitely better.

PR #1544, opened 2017-04-06
Ignore haxe tests if HAXEPATH not exists

SlavaRa:


Neverbirth:
Seems quite drastic to me... not all tests calling SetHaxeFeatures need Haxe. Also, this could lead to PRs breaking something without realizing it.

From my side I would go with calling Assert.Fail.

PR #1547, opened 2017-04-10
Add argument to Save

chipshort:
Needed to fix a bug in #1540

### DIFF
--- a/FlashDevelop/MainForm.cs
+++ b/FlashDevelop/MainForm.cs
@@ -1713,6 +1713,24 @@ namespace FlashDevelop
             }
         }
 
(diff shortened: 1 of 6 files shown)


FlashDevelop/MainForm.cs
```diff
@@ -1713,6 +1713,24 @@ public void OnDocumentModify(ITabbedDocument document)
             }
         }
 
+        public void OnFileSave(ITabbedDocument document, string oldFile, string reason)
```
Neverbirth: Why did you decide to not change the old OnFileSave without reason to just call this one?

chipshort: Ah, I think I did it for testing and forgot about it. Will fix

PR #1551, opened 2017-04-11
Diagnostics Completion

chipshort:
Adds methods to HaxeComplete for `@diagnostics` display mode and a linter that uses it.


External/Plugins/HaXeContext/Completion/HaxeComplete.cs
```diff
@@ -529,7 +582,21 @@ public enum HaxeCompilerService
...
         /// https://haxe.org/manual/cr-completion-top-level.html
         /// </summary>
-        TOP_LEVEL
+        TOP_LEVEL,
+
+        /// <summary>
+        /// Since Haxe 3.?.?
+        /// no manual available
+        /// </summary>
+        DIAGNOSTICS
```
SlavaRa: v 3.3.0-rc1
https://github.com/HaxeFoundation/haxe/commit/8e7f7327

chipshort: Thank you.
I was just about to ask this.

External/Plugins/HaXeContext/Completion/HaxeComplete.cs
```diff
@@ -111,9 +119,8 @@ void Notify<T>(HaxeCompleteResultHandler<T> callback, T result)
 
         protected virtual string[] BuildHxmlArgs()
         {
-            // check haxe project & context
-            if (PluginBase.CurrentProject == null || !(PluginBase.CurrentProject is HaxeProject)
-                || !(ASContext.Context is Context))
+            // check haxe project
+            if (PluginBase.CurrentProject == null || !(PluginBase.CurrentProject is HaxeProject))
```
SlavaRa:
```csharp
if (!(PluginBase.CurrentProject is HaxeProject)) return null;
```
because null cannot be HaxeProject

External/Plugins/HaXeContext/Linters/DiagnosticsLinter.cs
```diff
@@ -0,0 +1,77 @@
...
+    {
+        private static Dictionary<HaxeDiagnosticsSeverity, LintingSeverity> conversionDict =
+            new Dictionary<HaxeDiagnosticsSeverity, LintingSeverity>()
+            {
+                {HaxeDiagnosticsSeverity.ERROR, LintingSeverity.Error},
+                {HaxeDiagnosticsSeverity.INFO, LintingSeverity.Info},
+                {HaxeDiagnosticsSeverity.WARNING, LintingSeverity.Warning}
+            };
+
+        public void LintAsync(string[] files, LintCallback callback)
```
SlavaRa: This functionality is already used somewhere?

chipshort: @SlavaRa I'm not sure I understand the question. Could you elaborate, please?

SlavaRa: @chipshort Before using this functionality, you need to check the compiler version to convince that "@diagnostics" is available

chipshort: @SlavaRa Good catch, I'll fix that.

External/Plugins/HaXeContext/Linters/DiagnosticsLinter.cs
```diff
@@ -33,44 +34,51 @@ public void LintAsync(string[] files, LintCallback callback)
...
+                                Severity = conversionDict[res.Severity]
+                            };
+
+                            switch (res.Kind)
+                            {
+                                case HaxeDiagnosticsKind.UNUSEDIMPORT:
+                                    result.Description = "Unused import";
+                                    break;
+                                case HaxeDiagnosticsKind.UNUSEDVAR:
+                                    result.Description = "Unused variable";
```
Neverbirth: I guess this should be localized. And maybe we should create some `Kind`, `ResultId` or similar property for `LintingResult` to be able to maybe link it in the future with some other stuff? Although with proper implementation case I guess adding this already will surely mean it will be changed in the future at the first time we will really need it.

Neverbirth: Also something like native description could be useful if you want to link linting results to compiler results...

chipshort: I will localize them. About the LintingResult, I think it should be added in the future when there is an actual need for it.

chipshort: What do you mean with native description? Are you referring to the id in your first comment?

Neverbirth:
There are still improvements possible for this that are in the works, merging now to also be able to get more feedback from other users.

PR #1554, opened 2017-04-12
Improvements to LintingHelper

chipshort:


External/Plugins/LintingHelper/LintingCache.cs
```diff
@@ -17,6 +17,14 @@ public LintingCache()
             results = new Dictionary<string, List<LintingResult>>();
         }
 
+        /// <summary>
+        /// Returns all cached results of all files that are still in the cache
+        /// </summary>
+        public List<LintingResult> GetAllResults()
+        {
+            return new List<LintingResult>(results.Values.SelectMany(i => i));
```
SlavaRa:
```csharp
public List<LintingResult> GetAllResults()
{
    return results.Values.SelectMany(i => i).toList();
}
```
or c# 6

```csharp
public List<LintingResult> GetAllResults() => results.Values.SelectMany(i => i).toList();
```

External/Plugins/LintingHelper/Managers/LintingManager.cs
```diff
@@ -180,23 +179,38 @@ static void ApplyLint(string[] files, string language, List<LintingResult> resul
...
                             color = lang.editorstyle.ErrorLineBack;
+                            id = 3;
                             break;
                         case LintingSeverity.Warning:
                             color = lang.editorstyle.DebugLineBack;
+                            id = 4;
                             break;
                         case LintingSeverity.Info:
                             color = lang.editorstyle.HighlightWordBackColor;
+                            id = 5;
```
SlavaRa: Magic numbers?

SlavaRa: Magic numbers?

chipshort: Well, most numbers should work fine here, except 0, 1 and 2, because those are already used in various places:
https://github.com/fdorg/flashdevelop/blob/development/PluginCore/ScintillaNet/ScintillaControl.cs#L5719
https://github.com/fdorg/flashdevelop/blob/development/PluginCore/ScintillaNet/ScintillaControl.cs#L6800
https://github.com/fdorg/flashdevelop/blob/development/External/Plugins/AS3Context/Context.cs#L650
(to only name some of them)
But I think it would be better to have some api or something to handle this differently. From what I can tell, the only thing that counts, is that the indicators are different from each other. Otherwise they override each other.

External/Plugins/LintingHelper/LintingCache.cs
chipshort: @SlavaRa Ah, yes. I knew about it actually, but wasn't sure how far I should take it.

SlavaRa: I think that you should use ```results.Values.SelectMany(i => i).toList();``` instead of ```new List<LintingResult>(results.Values.SelectMany(i => i));```

PR #1557, opened 2017-04-12
[Linux] Added SetVersion.sh

SlavaRa:


FlashDevelop/FlashDevelop.csproj
```diff
@@ -257,13 +257,20 @@
...
+  <PropertyGroup Condition=" '$(OS)' == 'Unix' ">
+    <PreBuildEvent>"$(SolutionDir)\SetVersion.sh" "$(SolutionDir)"</PreBuildEvent>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(OS)' != 'Unix' And '$(Platform)' == 'AnyCPU' ">
     <PreBuildEvent>"$(SolutionDir)\SetVersion.bat" "$(SolutionDir)" &amp;&amp; ^
 IF EXIST "$(TargetPath)" move /Y "$(TargetPath)" "$(TargetDir)$(TargetName)86$(TargetExt)"</PreBuildEvent>
   </PropertyGroup>
+  <PropertyGroup Condition=" '$(OS)' == 'Unix' And '$(Platform)' == 'AnyCPU' ">
+    <PreBuildEvent>"$(SolutionDir)\SetVersion.sh" "$(SolutionDir)" &amp;&amp; ^
+IF EXIST "$(TargetPath)" move /Y "$(TargetPath)" "$(TargetDir)$(TargetName)86$(TargetExt)"</PreBuildEvent>
```
chipshort: Is this supposed to run on a unix machine? If so, what about the move and IF EXIST. Those are windows specific. Or does mono do some magic here?

SetVersion.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+target_dir=$1
+cd $target_dir
+mono External/Tools/SetVersion/SetVersion.exe FlashDevelop/Properties/AssemblyInfo.cs "%APPVEYOR_BUILD_NUMBER%"
```
chipshort: I'm not sure about the `%APPVEYOR_BUILD_NUMBER%` wouldn't it be "$APPVEYOR_BUILD_NUMBER" on linux / unix?

FlashDevelop/FlashDevelop.csproj
SlavaRa: all ok https://travis-ci.org/SlavaRa/flashdevelop

SetVersion.sh
SlavaRa: all ok https://travis-ci.org/SlavaRa/flashdevelop

FlashDevelop/FlashDevelop.csproj
Neverbirth: You sure you end up with both FlashDevelop and FlashDevelop64?

SlavaRa: At the moment I have a build of 64 bit version disabled

Neverbirth: Heh, so you are not running that IF EXIST and move then.

SlavaRa: Travis-CI config https://github.com/SlavaRa/flashdevelop/blob/development-linux/.travis.yml
I can add a build there for a 64-bit system, but this has nothing to do with the configuration of the project, it is made in analogy with the Windows version
I do not understand, what is the problem now?

Neverbirth: It seems the solution wouldn't successfully compile when involving both x86 and AnyCPU builds as that command is specific to Windows. Could you check? Even if we don't have both builds to at least to not have the AnyCPU one to fail if it's the case right now.

SlavaRa: AnyCPU is used for PluginCore https://github.com/SlavaRa/flashdevelop/blob/development-linux/.travis.yml#L15
But in order to use it for the editor when build on Linux there will be some necessary modifications
Of course I can remove the lines 270-273 as unused

SlavaRa: @Neverbirth ?

Neverbirth: Which modifications are needed for AnyCPU?

SlavaRa: If I was sure about what to do, I would have done it already)
This PR allows my team to run the assembly now without any unnecessary actions. I think that the guys in the process of work will adjust the build for AnyCPU

Neverbirth: I see, I'm sure one of the changes is that IF EXIST and move part heh.

PR #1559, opened 2017-04-13
Fix NullReferenceException

chipshort:


External/Plugins/CodeRefactor/Controls/MoveDialog.cs
```diff
@@ -174,5 +176,10 @@ protected override void OnKeyDown(KeyEventArgs e)
                     break;
             }
         }
+
+        private void OnTreeSelectedIndexChanged(object sender, EventArgs e)
+        {
+            this.processButton.Enabled = true;
```
Neverbirth: I'm not sure this really adds too much value, because it would be interesting to disable it when no item is selected, but when the list becomes empty and no item is selected the even doesn't trigger. It would be nicer to fully make it enabled or disabled on every case.

PS: I didn't even try this dialog ever before nor knew where it was displayed it heh.

External/Plugins/CodeRefactor/Controls/MoveDialog.cs
```diff
@@ -121,7 +121,15 @@ void FillTree()
...
             if (classpaths.Count > 0) tree.Items.AddRange(classpaths.ToArray());
-            if (tree.Items.Count > 0) tree.SelectedIndex = 0;
+            if (tree.Items.Count > 0)
+            {
+                tree.SelectedIndex = 0;
+                this.processButton.Enabled = true;
+            }
+            else
+            {
+                this.processButton.Enabled = false;
```
Neverbirth: Remove this. to be consistent with the rest of the file. Also, it seems to me this could be rewritten to:

```
if (classpaths.Count > 0)
{
    tree.Items.AddRange(classpaths.ToArray());
    tree.SelectedIndex = 0;
    this.processButton.Enabled = true;
}
else this.processButton.Enabled = false;
```
I don't think it's possible for `classpaths` to have at least 1 item and then tree be empty heh.

PR #1560, opened 2017-04-13
LintHelper should not work at the time of searching for all references

SlavaRa:


Neverbirth:
I don't think this is the way to solve this issue. There is already work in progress for this and other similar UX problems with the Results panel. Hopefully tomorrow will be pushed.

SlavaRa:
When using the find all references, files are opened this triggers diagnostics. 
Which in turn breaks the output of the result and leads to the closing of the editor, because on my project the answer from the compiler will be about two minutes later, and the file will be already closed.

Neverbirth:
I see, for that one there is also another work in progress. But I will merge this in the meantime.

SlavaRa:
Thanks

PR #1562, opened 2017-04-13
[Haxe] Ability to install missing libraries from haxelib. Part 1

SlavaRa:


Neverbirth:
Nice, i was wondering about this one recently. I think your other PR will wait a bit more heh

External/Plugins/HaXeContext/PluginMain.cs
```diff
@@ -166,6 +170,30 @@ public void HandleEvent(Object sender, NotifyEvent e, HandlingPriority priority)
...
+                        foreach (var pattern in patterns)
+                        {
+                            var match = Regex.Match(item.Message, pattern);
+                            if (match.Success) libs.Add(match.Groups[2].Value);
+                        }
+                    }
+                    logCount = TraceManager.TraceLog.Count;
+                    if (libs.Count == 0) return;
+                    var label = "Error";
+                    var text = $"Libraries {string.Join(", ", libs.ToArray())} is not installed\nDo you want to install their using haxelib?";
```
SlavaRa: I think we need to put it in the localization, @Neverbirth please help me come up with the text (if this one does not fit) and the key

SlavaRa:
Part 2: install libraries from git|hg
Part 3: UI to display the installation progress

External/Plugins/HaXeContext/Context.cs
```diff
@@ -1668,8 +1668,26 @@ public override bool BuildCMD(bool failSilently)
...
+        #region haxelib
+
+        internal void Install(List<string> libraries)
+        {
+            var haxePath = PathHelper.ResolvePath(GetCompilerPath());
+            if (!Directory.Exists(haxePath) && !File.Exists(haxePath))
+            {
+                ErrorManager.ShowInfo(TextHelper.GetString("Info.InvalidHaXePath"));
+                return;
+            }
```
Neverbirth: I would say that Haxelib detection should be extracted and centralized for this and `LookupLibrary`, also if done in a proper way, [Haxe Shim](https://github.com/lix-pm/haxeshim) or others could be supported as well.

External/Plugins/HaXeContext/PluginMain.cs
```diff
@@ -8,6 +8,7 @@
 using PluginCore.Utilities;
 using PluginCore;
 using System.Text.RegularExpressions;
+using System.Windows.Forms;
```
Neverbirth: It would be nice to not add direct `Windows.Forms` dependencies here.

External/Plugins/HaXeContext/PluginMain.cs
Neverbirth: Uhm... not sure...  printing all libraries may look really ugly in projects with tons of missing dependencies. Maybe I would use something like "Missing libraries detected. Would you like to try to install them through the package manager?" Also, I wouldn't use a `MessageBox` for this since it's modal and I think the user experience is not the best for this. The key could just be "Info.MissingLib" or similar.

External/Plugins/HaXeContext/Context.cs
SlavaRa: I think it will be better to do this separately from this task

External/Plugins/HaXeContext/PluginMain.cs
SlavaRa: I added localization.
In occasion of UI/UX it is necessary to experiment, I wanted to make it separate task, when the basic functionality will be ready

Neverbirth: OK, but I think it should be really changed in the near future.

SlavaRa: All depends on the speed of accepting PRs :)

External/Plugins/HaXeContext/Context.cs
Neverbirth: Yes, adding support for other package managers would be a separate task of course. Just commenting of an improvement it can bring to just extract the haxelib resolution for both parts of code (which that one could be done now).

At any rate I have the feeling decoupling from Haxelib may be really needed in the future. Also, in my company, and I'm sure in others, we have our own private Haxelib repository, so for example this PR may be a bit of a pain in the ass right now with the `MessageBox` (of course it's a nice addition and for most people welcome).

External/Plugins/HaXeContext/PluginMain.cs
Neverbirth: Heh, I never said I won't merge the PR as-is. In fact I already decided yesterday at night before commenting I was going to do it now. Although maybe now I won't do it with your comment... ;P.

PR #1564, opened 2017-04-14
Multiple ResultsPanels

chipshort:
Adds support for multiple results panels. Also changed CodeRefactor and "Find and Replace in Files" to use it.

### DIFF
--- a/External/Plugins/ResultsPanel/PluginMain.cs
+++ b/External/Plugins/ResultsPanel/PluginMain.cs
@@ -7,6 +7,7 @@ using PluginCore.Helpers;
 using PluginCore.Localization;
 using PluginCore.Managers;
 using PluginCore.Utilities;
+using ResultsPanel.Managers;
 using WeifenLuo.WinFormsUI.Docking;
 
 namespace ResultsPanel
@@ -22,7 +23,7 @@ namespace ResultsPanel
         private Settings settingObject;
         private String settingFilename;
         private PluginUI pluginUI;
-        private Image pluginImage;
+        internal Image pluginImage;
 
         // Shortcut management
         public const Keys CopyEntryKeys = Keys.Control | Keys.C;
@@ -145,6 +146,7 @@ namespace ResultsPanel
                     break;
 
                 case EventType.Trace:
+                    ResultsPanelManager.OnTrace();
                     this.pluginUI.AddLogEntries();
                     break;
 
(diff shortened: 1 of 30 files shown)


Neverbirth:
Conflicts

chipshort:
Was already on it

External/Plugins/ResultsPanel/PluginMain.cs
```diff
@@ -122,13 +125,26 @@ public void HandleEvent(Object sender, NotifyEvent e, HandlingPriority priority)
                     if (evnt.Action == "ResultsPanel.ClearResults")
                     {
                         e.Handled = true;
-                        this.pluginUI.ClearOutput();
+
+                        if (evnt.Data == null)
```
Neverbirth: Could you make the deault results panel also like another panel inside the panel helper? Having this kind of differences makes code slightly more complex and harder to read.

wise0704:
This broke the shortcuts.

wise0704:
I've started working on the fixes, but I'm not sure if this PR was really worth.

Firstly, it's using the concept of "groups", meaning I can't have 2 windows of Find All References, which is really what I thought of as useful. Maybe I can fix this while working on the other fix.

Secondly, as @Neverbirth said, the layout is not saved. However, since they are more like "temporary panels", I don't really think the layout data is necessary to be saved.

Thirdly, it (will) make the shortcuts ambiguous. I might press F8 for next entry of lists I have in Find All References, but depending on the last active window, it can go to the next entry of Rename.

Neverbirth:
> I've started working on the fixes, but I'm not sure if this PR was really worth.

I consider this PR an improvement over what FD had before, if it motivates you to make something better then doubly welcome. What is your proposal? The shortcuts may be a problem with any other solution, as you have to set different ones for any of them anyway, or either set an active panel (something similar to tabs in the results panel? but that seems quite dirty to me as you would group multiple unrelated things together) and only enable the shortcuts for it, and both solutions could be applied here equally.

> Firstly, it's using the concept of "groups", meaning I can't have 2 windows of Find All References, which is really what I thought of as useful. Maybe I can fix this while working on the other fix.

Why is it a problem? I would say it's just an implementation detail on when the groups are registered. You could change it to create everytime different panels, add the option to the Find and Replace dialog to send the result to new panels (or multiple predefined ones as in VS), etc.

> However, since they are more like "temporary panels", I don't really think the layout data is necessary to be saved.

Even if some of them could be considered temporary, I think the last layout information should be saved for them, it's an annoyance that on every program run you have to position again the panels to your liking.

PR #1567, opened 2017-04-14
[Haxe][ASGenerator] Improved property override

SlavaRa:


External/Plugins/ASCompletion/Completion/ASGenerator.cs
```diff
@@ -4135,41 +4136,40 @@ public static void GenerateOverride(ScintillaControl Sci, ClassModel ofClass, Me
...
                 var name = newMember.Name;
-                var parameters = member.Parameters;
                 if (parameters != null && parameters.Count == 1) type = parameters[0].Type;
                 type = FormatType(type);
                 if (type == null && !features.hasInference) type = features.objectKey;
                 newMember.Type = type;
-                if (ofClass.Members.Search(name, FlagType.Getter, 0) != null && (!IsHaxe || parameters[0].Name == "get"))
+                var currentClass = context.CurrentClass;
+                if (ofClass.Members.Search(name, FlagType.Getter, 0) != null
+                    && (!IsHaxe || (parameters?[0].Name == "get" && currentClass.Members.Search($"get_{name}", FlagType.Function, 0) == null)))
```
Neverbirth: Can `parameters` be null here? The old check always assumed in cannot be.

SlavaRa: becase https://github.com/fdorg/flashdevelop/blob/development/External/Plugins/ASCompletion/Completion/ASGenerator.cs#L4152

SlavaRa: PVS-Studio does not believe that there can not be null ;)

PR #1568, opened 2017-04-16
SetVersion: Path.DirectorySeparatorChar instead of hardcoded backslash

RealyUniqueName:
Should i also commit compiled SetVersion.exe?


Neverbirth:
Yes, SetVersion.exe should be committed too.

PR #1569, opened 2017-04-16
Fix build errors on Linux

RealyUniqueName:


Neverbirth:
OK, this broke AnyCPU build on Windows.

PR #1570, opened 2017-04-17
Closing brace inside strings should not unindent

### LINKED ISSUE #489: Typing } in a string eats a tab
Seems to happen in 4.6.4 as well, so it's not a recent regression (if it is one).

wise0704:
Closes #489

Doesn't solve the second issue in the comments of the referenced issue.

### DIFF
--- a/PluginCore/ScintillaNet/ScintillaControl.cs
+++ b/PluginCore/ScintillaNet/ScintillaControl.cs
@@ -9,6 +9,7 @@ using System.Runtime.InteropServices;
 using System.Text;
 using System.Windows.Forms;
 using ScintillaNet.Configuration;
+using ScintillaNet.Lexers;
 using PluginCore.FRService;
 using PluginCore.Utilities;
 using PluginCore.Managers;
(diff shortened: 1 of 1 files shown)


PluginCore/ScintillaNet/ScintillaControl.cs
```diff
@@ -5902,14 +5903,11 @@ private void OnSmartIndent(ScintillaControl ctrl, int ch)
...
-                            int previousIndent = GetLineIndentation(curLine - 1);
-                            int match = SafeBraceMatch(position - 1);
-                            if (match != -1)
+                            int position = CurrentPos - 1;
+                            int match = SafeBraceMatch(position); //SafeBraceMatch() calls Colourise(0, -1)
+                            if (match != -1 && !PositionIsInString(position))
                             {
-                                previousIndent = GetLineIndentation(LineFromPosition(match));
-                                IndentLine(curLine, previousIndent);
+                                IndentLine(LineFromPosition(position), GetLineIndentation(LineFromPosition(match)));
```
Neverbirth: You are now doing `CurrentPos - 1`, so the line to indent may be the previous one.

wise0704: `CurrentPos - 1` is always the index of `}`, which should be the same line as `CurrentPos`

Neverbirth: Right

PR #1575, opened 2017-04-21
Extra options for AddClosingBraces

wise0704:
## Part 1 - 3c3bf6f
Ignore Whitespace setting

Rather than ignoring whitespace globally, allow individual settings.
Features:
- Overwrite closing with whitespace in the middle (`' '` or `'\t'`)
**Add `}`:**

Result:  Before: 
- Delete matching closing brace with whitespace in the middle (`char.IsWhiteSpace()`)
**Backspace:**

Result:  Before: 
**Also deletes new line characters:**

Result:  Before: 

## Part 2 - 42a05a4
Add space setting

Insert a space after typing the opening brace. Only `{ Braces }` brace has this option checked by default.
Features:
- Inserting `{`:

- Separated haxe interpolation expression brace rule:

Which doesn't have the Add Space option checked:

## Part 3 - d96595d
Remove whitespaces on newline after opening brace.
Features:
- Pressing enter:

Result:  Before: 

---
**Miscellaneous**
Part 1 also contains:
- AddClosingBracesRulesEditorForm now uses settings default font, enabling font scaling.
- btnAddRule's width will also scale accordingly.

Part 2 also contains:
- In AddClosingBracesRulesEditorForm, removing items from ListBox at index 0 set selected index to -1. Now correctly sets selected index to 0.

### DIFF
--- a/External/Plugins/ASCompletion/Completion/ASComplete.cs
+++ b/External/Plugins/ASCompletion/Completion/ASComplete.cs
@@ -449,11 +449,20 @@ namespace ASCompletion.Completion
                 if (!(IsStringStyle(styleBefore) && IsStringStyle(styleAfter)) && !(IsCharStyle(styleBefore) && IsCharStyle(styleAfter))
                     || IsInterpolationExpr(sci, sci.CurrentPos - 1))
                 {
+                    char nextChar = sci.CurrentChar;
+                    int nextPos = sci.CurrentPos;
+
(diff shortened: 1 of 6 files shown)


Neverbirth:
The add space option seems like something that should be derived from formatting settings... although not sure how feasible it is...

wise0704:
Could use CodeFormatter "Spaces Inside Literal Braces".
But that means checking if it is '{', or matches hardcoded characters for any other braces if "Space Inside Parens", etc., were to be implemented too, which is actually worse.

PR #1578, opened 2017-04-23
Multiple ResultsPanel improvements

wise0704:
Summary:
- Fixed shortcuts
  - The active panel gets the input shortcuts.
  - Contextual shortcuts (Ctrl+C and Delete) are only executed when the active panel has focus.
- Fixed highlights (squiggles)
  - _Update:_ Added a setting `ResultsPanel.HighlightOnlyActivePanelEntries`
  - "Highlight Only Active Panel Entries" = false
    - Highlight all entries from all panels.
  - "Highlight Only Active Panel Entries" = true
    - Highlights appear when panel is active; disappear when another results panel is active.
    - The main panel always shows highlights.
- Active panel management
  - The panel that got the focus most recently is the active panel, regardless of whether it currently has focus or not.
- _Update:_ panel layout data is stored just like any other panels.
- _Update:_ some panel groups (Find All References and Find in Files) have an option to lock the panels, so further operations that opens the panel will instead create new panels.
- _Update:_ panels that don't need the filter buttons (info, warnings and errors) won't show them.

**Issues:**
- [x] Focus management with floating panels is really buggy. It sometimes fires `Enter` event, but most of the time it doesn't.

**Todo:**
- [x] Save layout data.
- [x] Allow multiple windows for Find All References.
- [x] Add a way to "lock" Find All References panel and Find in Files panel, to allow multiple panels.
- [x] Add a setting for whether to highlight entries from all panels or just the active and main panels.
- [x] Let certain groups to omit info, warning and error filter buttons.


External/Plugins/ResultsPanel/PluginMain.cs
```diff
@@ -271,17 +251,16 @@ public void CreatePluginPanel()
...
+            ResultsPanelHelper.Initialize(this, this.pluginUI);
         }
         
         /// <summary>
         /// Opens the plugin panel if closed
         /// </summary>
         public void OpenPanel(Object sender, EventArgs e)
         {
-            this.pluginPanel.Show();
+            ResultsPanelHelper.ActiveUI.ParentPanel.Show();
```
wise0704: This is when the user uses the shortcut "ViewMenu.ShowResults", or clicks View > Results Panel.
Would it be better to change it back to `this.pluginUI.ParentPanel.Show()`?

wise0704:
Should there be a setting for whether to allow multiple windows for Find All References?
Wouldn't be much of a hassle to add.

Neverbirth:
From my side I don't think I would ever switch back to the old behavior... I guess for some people it may be faster to just have one results panel, but then it's most times a PITA to work like that... you lose your results on so many unwanted occasions...

Neverbirth:
Why show squiggles only on active panel?

wise0704:
Not exactly the old behaviour. See how 'Find All References' creates a new panel each time. For example, in VS, the same panel for 'Find All References' is used unless you lock the panel; then multiple panels are created for each use.

Advantages of using that feature: it wouldn't be only 'Find All Reference', which is hardcoded to have multiple panels, to be able to have multiple panels for each use. Other panels will also be able to have multiple panels (which panels will have the option to let users allow multiple windows, will however be hardcoded, since you wouldn't want multiple linting results panels).

If we set the default of these panels to allow multiple panels, there aren't any disadvantages of adding this.

wise0704:
Those squiggles are more like highlights, and before I used to allow all highlights to be visible, and while testing, I figured that it looks confusing and not very useful.

The default panel, which shows the errors and warnings, will always have them visible.

Neverbirth:
OK, sorry, I understood your message as a setting for the old behavior. In that case I think it may be useful.

I would prefer to have all squiggles visible at any time. I know it may be confusing to see squiggles and not know where the problem comes from, but IMHO that's better than having some issue and not knowing you have it because you didn't activate the corresponding panel.

I guess we should display issues when hovering some squiggle, but that's another subject.

External/Plugins/ResultsPanel/Helpers/ResultsPanelHelper.cs
```diff
@@ -1,83 +1,249 @@
...
+using System.Collections.Generic;
 using PluginCore;
 using PluginCore.Localization;
 using PluginCore.Managers;
 using WeifenLuo.WinFormsUI.Docking;
 
 namespace ResultsPanel.Helpers
 {
-    class ResultsPanelHelper
+    internal static class ResultsPanelHelper
```
Neverbirth: I have mixed feelings about making this fully static, but I don't think I will make it a blocker.

External/Plugins/ResultsPanel/ListViewEx.cs
```diff
@@ -1,94 +1,71 @@
...
-using System.Drawing.Imaging;
-using System.Linq;
-using System.Text;
 using System.Windows.Forms;
 using System.Windows.Forms.VisualStyles;
 
 namespace ResultsPanel
 {
-    class ListViewEx : System.Windows.Forms.ListViewEx
+    internal class ListViewEx : System.Windows.Forms.ListViewEx
```
Neverbirth: Why the change of all the access modifiers to internal? Also, since you already marked the class as internal itself it won't make much difference to change the fields from public to internal.

External/Plugins/ResultsPanel/PluginMain.cs
```diff
@@ -19,17 +19,12 @@ public class PluginMain : IPlugin
         private String pluginHelp = "www.flashdevelop.org/community/";
         private String pluginDesc = "Adds a results panel for console info to FlashDevelop";
         private String pluginAuth = "FlashDevelop Team";
-        private DockContent pluginPanel;
+        //private DockContent pluginPanel;
```
Neverbirth: Just remove it if it's not needed anymore.

External/Plugins/ResultsPanel/Settings.cs
```diff
@@ -25,6 +26,14 @@ public GroupingMethod DefaultGrouping
             get { return defaultGroup; }
             set { defaultGroup = value; }
         }
+
+        [DisplayName("Highlight Only Active Panel Entries")]
+        [LocalizedDescription("ResultsPanel.Description.HighlightOnlyActivePanelEntries"), DefaultValue(true)]
```
Neverbirth: Shouldn't the default be false as the backing field? If you really intended it to be true, I prefer it to be false.

External/Plugins/ResultsPanel/PluginMain.cs
```diff
@@ -19,17 +19,12 @@ public class PluginMain : IPlugin
...
         private Settings settingObject;
         private String settingFilename;
-        private PluginUI pluginUI;
+        internal PluginUI pluginUI;
```
Neverbirth: This change to internal seems to be unneeded?

PluginCore/PluginCore/Managers/TraceManager.cs
```diff
@@ -102,28 +102,82 @@ public static void AddAsync(String message, Int32 state)
...
+            }
+            return groupId + ":" + string.Join(",", args);
+        }
+
+        /// <summary>
+        /// Creates a string for group data with a unique id argument at the end. Used for groups with <see cref="TraceGroup.AllowMultiplePanels"/> set to <see langword="true"/>.
+        /// <para/>
+        /// <paramref name="groupId"/>:arg1,arg2,<paramref name="args"/>,uniqueId
+        /// </summary>
+        public static string CreateGroupDataUnique(string groupId, params string[] args)
```
Neverbirth: It feels too me like this is some special behavior that would be better to leave to the registerers of groups, but not sure it matters too much.

PluginCore/DockPanelSuite/Docking/DockPanel.FocusManager.cs
```diff
@@ -416,13 +416,13 @@ internal void SetActiveContent()
                     return;
 
                 if (m_activeContent != null)
-                    m_activeContent.DockHandler.IsActivated = false;
+                    m_activeContent.DockHandler.SetIsActivated(false);
```
Neverbirth: Have you  compared the behavior with the latest DPS? Do they do the same or do they have the same problem? If the latter you may want to create a PR for them.

Neverbirth:
Merging to let progress  go on, but left some comments.

External/Plugins/ResultsPanel/Helpers/ResultsPanelHelper.cs
wise0704: Static class is much better in the cases of:

- When the class only needs one instance
- The class needs to be initialised as soon as the functionality is needed
- The class lasts for the entire duration of the assembly's life
- The class doesn't need derived types or interface

If a singleton was to be used in this case, it's actually worse, because:

- It requires the class to have a constructor
- There is a cost of class instantiation
- There is the class definition, and also the instance object definition lying in the memory
- The object lies in the part of the memory which is targeted by the GC, meaning although negligible, it's a permanent object never to be GC'ed until the assembly unloads, only hindering the GC process by having to be moved around.

External/Plugins/ResultsPanel/ListViewEx.cs
wise0704: I see the `internal` on this line is only about code style.
But the members are kind of different when it comes to reflection, because the members are non-public. And I think it is better to keep every access modifier to as restricted as possible, as they aren't designed to be accessible (class is marked internal).

External/Plugins/ResultsPanel/PluginMain.cs
wise0704: It's used in PluginUI.

PluginCore/PluginCore/Managers/TraceManager.cs
wise0704: I didn't have it at the start, but as I was writing implementations, such as in Find All References, it was quite inconvenient, since i had to manually have a static variable for each implementation, and increment and pass the value.

One reason was to have only one static variable for the unique token.
Second reason was that the callers never really need to access this unique token unless wanted, and the value of the token is meaningless. It means they wouldn't really want to set or store the unique token themselves.

PluginCore/DockPanelSuite/Docking/DockPanel.FocusManager.cs
wise0704: I haven't yet compared with the latest DPS.

Side note is that, the choice to introduce a method SetIsActivated() instead of a setter for IsActivated was to follow the other conventions in the package - like the dock panel had IsActivated getter and SetIsActivated().

wise0704: Just checked.
It seems they still have it the way it was 
https://github.com/dockpanelsuite/dockpanelsuite/blob/master/WinFormsUI/Docking/DockPanel.FocusManager.cs#L468
https://github.com/dockpanelsuite/dockpanelsuite/blob/master/WinFormsUI/Docking/DockContentHandler.cs#L706

External/Plugins/ResultsPanel/Helpers/ResultsPanelHelper.cs
Neverbirth: All of the negative points you mention have a negative impact of only nanoseconds of difference even on several years old computers (there are benchmarks out there, some of them by MS), and most of them are impacts that happen only once, I would say the last point you mention is the more important one. I care more about the negative effects that static classes bring like not possible to switch one implementation with other, making it sometimes harder to refactor code, to test, of course in a class that keeps state it's impossible to create multiple instances, etc.

Since this is an internal class most of the negative points get diminished, but there are still some valid concerns.

External/Plugins/ResultsPanel/ListViewEx.cs
wise0704: By reflection, I mean for example, `dynamic` keyword.

External/Plugins/ResultsPanel/PluginMain.cs
Neverbirth: I see, I tried to look for references and didn't find any, but GitHub is not the best tool for that kind of operation.

External/Plugins/ResultsPanel/PluginMain.cs
wise0704: I had already removed it in the other branch :P

External/Plugins/ResultsPanel/Settings.cs
wise0704: Fixed in #1623

PR #1585, opened 2017-04-24
Speed up tests

wise0704:
capture capture2


wise0704:
Look at that

image

My other PR:

image

PR #1589, opened 2017-04-27
Clipboard History

wise0704:
image
image

### DIFF
--- a/FlashDevelop/Dialogs/ClipboardHistoryDialog.cs
+++ b/FlashDevelop/Dialogs/ClipboardHistoryDialog.cs
@@ -1,0 +1,275 @@
+﻿using System;
+using System.Drawing;
+using System.Windows.Forms;
+using PluginCore.Localization;
+using PluginCore.Managers;
+
+namespace FlashDevelop.Dialogs
+{
+    /// <summary>
+    /// A dialog displaying the clipboard history from <see cref="ClipboardManager"/>.
+    /// </summary>
+    public class ClipboardHistoryDialog : Form
+    {
+        private SplitContainer splitContainer;
+        private ListBox listBox;
+        private Button btnPaste;
+        private Button btnCancel;
+        private Button btnClear;
+        private RichTextBox richTextBox;
+
+        private ClipboardTextData selectedData;
+
+        /// <summary>
+        /// Creates a new instance of <see cref="ClipboardHistoryDialog"/>.
+        /// </summary>
+        public ClipboardHistoryDialog()
+        {
+            InitializeComponent();
+            InitializeFont();
+            InitializeLocalization();
+            InitializeListBox();
+        }
+
+        #region Windows Form Designer generated code
+
+        /// <summary>
+        /// Required method for Designer support - do not modify
+        /// the contents of this method with the code editor.
+        /// </summary>
+        private void InitializeComponent()
+        {
+            splitContainer = new SplitContainer();
+            btnClear = new Button();
+            listBox = new ListBox();
+            richTextBox = new RichTextBox();
+            btnCancel = new Button();
+            btnPaste = new Button();
+            splitContainer.Panel1.SuspendLayout();
+            splitContainer.Panel2.SuspendLayout();
+            splitContainer.SuspendLayout();
+            SuspendLayout();
+            // 
+            // splitContainer
+            // 
+            splitContainer.Dock = DockStyle.Fill;
+            splitContainer.FixedPanel = FixedPanel.Panel1;
+            splitContainer.Location = new Point(0, 0);
+            splitContainer.Name = "splitContainer";
+            splitContainer.Orientation = Orientation.Horizontal;
+            splitContainer.Panel1MinSize = 100;
+            splitContainer.Panel2MinSize = 100;
+            splitContainer.Size = new Size(482, 453);
+            splitContainer.SplitterDistance = 171;
+            splitContainer.TabIndex = 0;
+            // 
+            // splitContainer.Panel1
+            // 
+            splitContainer.Panel1.Controls.Add(btnClear);
+            splitContainer.Panel1.Controls.Add(listBox);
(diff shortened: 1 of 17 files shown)


wise0704:
An alternative to the listbox texts is to wrap the `ClipboardTextData` in a class, and override `ToString()`
```c#
if (data.IsTextFormat)
{
    int lineBreak = data.Text.IndexOfAny(new[] { '\r', '\n' });
    return "[" + data.Format + "]" + (lineBreak >= 0 ? data.Text.Remove(lineBreak) : data.Text);
}
return "[" + data.Format + "]";
```

PluginCore/PluginCore/Resources/en_US.resx
```diff
@@ -6169,4 +6169,8 @@ AND - All conditions are met.</value>
     <value>Paste History</value>
     <comment>Added after 5.2.0</comment>
   </data>
+  <data name="FlashDevelop.Description.ClipboardHistorySize" xml:space="preserve">
+    <value>The number of items to store in the clipboard history. Minimun value is 1.</value>
```
wise0704: "minimu**n**"...

wise0704:
Also, `ClipboardHistoryDialog` can derive `PluginCore.Controls.SmartForm` instead, then it will save its window size to a file (Settings\FormStates\\%FormGuid%.fdb).
But it will also do probably unwanted things like `ScaleHelper.AdjustForHighDPI()`.

Neverbirth:
Some random thoughts:

- I'm a bit concerned about using a hook. Starting from Windows Vista MS added some protections that may result in applications being removed from a global hook like this, and you won't know about it, It used to happen in an application from my previous company; there is a registry entry for this, although it requires admin rights and would feel hacky and bad for us having to change it. I don't think it would happen in a case like this, because the protection is mainly about a program taking too long to process a message and pass it to the next listener in the chain, but who knows. 

- Why do you even put things that cannot be pasted into the list? I don't think it's worth it. But by your comments I understand this PR is still WIP.

I will try to play with this more tomorrow. I checked VS out of curiosity and it doesn't seem to use a global clipboard hook? I will try IJ tomorrow.

wise0704:
- https://msdn.microsoft.com/en-us/library/ms649016(v=vs.85).aspx#dataxchg_Monitoring_Clipboard_Contents
  >There are three ways of monitoring changes to the clipboard. The oldest method is to create a clipboard viewer window. Windows 2000 added the ability to query the clipboard sequence number, and Windows Vista added support for clipboard format listeners. Clipboard viewer windows are supported for backward compatibility with earlier versions of Windows. New programs should use clipboard format listeners or the clipboard sequence number.

  I think the hook you're referring to is [SetClipboardViewer](https://msdn.microsoft.com/en-us/library/ms649052(v=vs.85).aspx), which is the window handle chain method. The one I'm using is [AddClipboardFormatListener](https://msdn.microsoft.com/en-us/library/ms649033(v=vs.85).aspx):

  >This method is recommended over creating a clipboard viewer window because it is simpler to implement and avoids problems if programs fail to maintain the clipboard viewer chain properly or if a window in the clipboard viewer chain stops responding to messages.

- I was thinking of letting users re-copy contents from the history, back into the clipboard to allow users the "restore previous clipboard data" feature. 
That wasn't implemented yet mainly because I couldn't be bothered with inserting bitmap as RTF in the rich text box (users need to at least see the preview of the content they're about to restore).
But as you said, non-textual data can just be ignored. I'll leave the decision to you.

- I'd say, if possible, limiting the clipboard history to only the copy operations from scintilla control is not suitable. For example, you could want to copy&paste a file name from the project panel. A normal Ctrl+V would paste the content, but if it doesn't appear in the history, it wouldn't make too much sense.

Neverbirth:
1. Could be, I will have to check this in depth. In the meantime I tried IJ, IJ is using a global clipboard checker for its history.
2. I don't get what you mean with letting users re-copy back into the history :S. From my side I would just display and paste RTF text in plain format, not RTF... it will be a pain in the ass for some stuff, also the RichText control wrapped by Windows Forms is an old one, you can change it to a newer one, but even so I wouldn't go that way. IJ is only capturing normal text, not even RTF.
3. VS is doing it that way heh, Although again, not IJ. Personally I don't have a big preference for one or the other.

wise0704:
For 2., okay, I'll display only textual data in the history.

FlashDevelop/Dialogs/ClipboardHistoryDialog.cs
```diff
@@ -0,0 +1,310 @@
...
+
+            if (selectedData != null)
+            {
+                if (selectedData.Format == DataFormats.Text)
+                {
+                    richTextBox.Text = selectedData.Text;
+                }
+                else if (selectedData.Format == DataFormats.Rtf)
+                {
+                    richTextBox.Rtf = selectedData.Rtf;
```
Neverbirth: If I understood correctly you are not copying RTF anymore, so I guess this if is not needed.

FlashDevelop/Dialogs/ClipboardHistoryDialog.cs
```diff
@@ -0,0 +1,298 @@
...
+
+        #endregion
+
+        #region Events
+
+        private void ListBox_DrawItem(object sender, DrawItemEventArgs e)
+        {
+            e.DrawBackground();
+
+            string text = (e.Index + 1) + "    " + listBox.GetItemText(listBox.Items[e.Index]).TrimStart();
```
wise0704: Currently the ListBox only displays the first line:
image
If I change the code to
```c#
string[] lines = listBox.GetItemText(listBox.Items[e.Index]).Split('\n');
for (int i = 0; i < lines.Length; i++)
{
    lines[i] = lines[i].Trim();
}
string text = (e.Index + 1) + "    " + string.Join(" ", lines);
```
then it merges all lines together:
image

Which would be better?

PR #1590, opened 2017-04-28
[Haxe][ASGenerator] Improved callback code generation.

SlavaRa:


External/Plugins/ASCompletion/Completion/ASGenerator.cs
```diff
@@ -2832,6 +2832,8 @@ private static void GenerateFunctionJob(GeneratorJobType job, ScintillaControl s
             }
             if (caller?.Parameters != null && caller.Parameters.Count > 0)
             {
+                Func<string, string> cleanType = null;
+                cleanType = s => s.StartsWith("(") && s.EndsWith(')') ? cleanType(s.Trim('(', ')')) : s;
```
Neverbirth: I don't know if this simple check is good enough? Maybe you should count the number of open and closing parenthesis to make sure they are the same. Because you could generate the callback from wrong code that doesn't have the last closing parenthesis but has others for example.

What do you think of this?

SlavaRa: I think that if there is an incorrect type here, then the error was earlier, when determining the type

Neverbirth: But it can happen. Sadly I don't have the exact test case here, but I saw it yesterday while playing around. Anyway, merging this for now, I don't think it's that important.

SlavaRa: Thanks

PR #1591, opened 2017-04-28
ScintillaControl CopyRTF Improvement

wise0704:


PluginCore/ScintillaNet/Configuration/Scintilla.cs
```diff
@@ -191,7 +191,7 @@ public KeywordClass GetKeywordClass(string keywordName)
 
         public Language GetLanguage(string languageName)
         {
-            return GetLanguages().Find(language => language.name == languageName);
+            return Array.Find(AllLanguages, language => language.name == languageName);
```
Neverbirth: Why this change? It seems to be unrelated to the PR?

Neverbirth:
Just curious, did you try looking at latest Scintilla.NET (both the old one In CodePlex and the new one in GitHub) to see how they do it? maybe they have something better, or even if not, provide some helpful tips.

Neverbirth:
I would prefer if you revert the changes to ScintillaControl for the time being as they are not needed here.

wise0704:
I did.
https://github.com/jacobslusser/ScintillaNET/blob/0c308e88cb67609a7444648c7a9f4271bad92b7e/src/ScintillaNET/Helpers.cs#L456

Firstly, they have a tab size definition, which I think is unnecessary. VS doesn't have it implemented either.
And they seem to have a different style values management.

Side note is that [this line](https://github.com/jacobslusser/ScintillaNET/blob/0c308e88cb67609a7444648c7a9f4271bad92b7e/src/ScintillaNET/Helpers.cs#L602) in their code will put an unexpected space in text when all `if` cases right above it were skipped.
I believe the [`@"\par"`](https://github.com/jacobslusser/ScintillaNET/blob/0c308e88cb67609a7444648c7a9f4271bad92b7e/src/ScintillaNET/Helpers.cs#L666) without any whitespace after it will also cause trouble too.

Anyway that aside, my implementation seems to produce really similarly to VS.

PluginCore/ScintillaNet/ScintillaControl.cs
```diff
@@ -725,14 +725,17 @@ private void SetLanguage(String value)
                     }
                     catch (Exception ex)
                     {
+                        String info;
                         if (lexerType == null)
                         {
-                            ErrorManager.ShowWarning($"Lexer '{lang.lexer.name}' ({lang.lexer.key}) unknown.", ex);
+                            info = String.Format("Lexer '{0}' ({1}) unknown.", lang.lexer.name, lang.lexer.key);
+                            ErrorManager.ShowWarning(info, ex);
```
SlavaRa: @wise0704 Why did you need to change this code?

wise0704: I changed it while browsing the file, and added in the first commit. That's a revert change you've commented on.
>I would prefer if you revert the changes to ScintillaControl for the time being as they are not needed here. - @Neverbirth

SlavaRa: Ah, ok

PR #1594, opened 2017-04-29
[Haxe] Added setting "Enable Compiler For Find All References"

SlavaRa:


Neverbirth:
Oh, @chipshort is also working on something like this, although a bit better. I will merge it, but it will get replaced sooner or later.

SlavaRa:
@Neverbirth Let's somehow synchronize, so as not to do each other's work?

Neverbirth:
Yes, when I saw the PR I thought the same. As soon as I have some time I will try to collect my list of things and sync up with you.

SlavaRa:
@chipshort did a lot, but there are many problems. I will describe them after the weekend. @Neverbirth go to chat

Neverbirth:
Well, some problems are known and in the works, also, most of them are mainly coming from other stuff in FD than his changes.

Please create issues as you see just in case something has been missed.

markknol:
Nice! Will this feature also refactor using the compiler services?

Neverbirth:
Yes, although in my experience it's not so good when it comes to some stuff... I prefer to use the built-in find references in a lot of cases in my main project codebase...

markknol:
Maybe it could use some kind of mixed results, just like the completion? Other than that, the team working on vscode support also heavily depend on the compiler services so pretty sure it will improve if it misses stuff.

PR #1602, opened 2017-05-02
ESC to Close Panel

### LINKED ISSUE #1347: Close unpinned panels with Esc key
Colleague request: Right now unpinned panels may take a lot of time to hide, plus if the mouse is over them, or the they have the focus, there is no way to close them by the keyboard. Using Escape, or Shift+Escape, would be a good way to improve the UX of the IDE, and I think it won't conflict with any other part of the application.

wise0704:
Closes #1347


Neverbirth:
Since AFAIK this doesn't work 100% maybe we should decline the PR? Or do you prefer to keep it open just as a reminder of looking this more in depth in the future?

wise0704:
This does work, but this is going the opposite direction of the way I prefer keys should be handled. I've pushed the commits just in case, but I don't mind closing this PR. If the other shortcut PR's are finished, the whole code written here would be deleted anyway.

wise0704:
Update:
This change is now also relevant for my other shortcut branch since https://github.com/wise0704/FlashDevelop/commit/5928a702d8de1b4ebbf61bd76e590a4d346ea86b.

PR #1604, opened 2017-05-03
Resolve Ctrl+Shift+V default shortcut conflicts

wise0704:


Neverbirth:
Just extra information: the Ctrl+Shift+V was previously set by default for Close All Windows, and it was conflicting with the new Clipboard history feature. Since Ctrl+Shift+V is standard in other IDEs for the clipboard history we made the same here and redefined Close All Windows to Ctrl+Alt+W.

PR #1607, opened 2017-05-04
Fixed "Sequence contains no elements"

SlavaRa:
```
   at System.Linq.Enumerable.Last[TSource](IEnumerable`1 source)
   at ASCompletion.Completion.ASGenerator.GetStatementReturnType(ScintillaControl sci, ClassModel inClass, String line, Int32 startPos)
   at ASCompletion.Completion.ASGenerator.AssignStatementToVar(ClassModel inClass, ScintillaControl sci, MemberModel member)
   at ASCompletion.Completion.ASGenerator.GenerateJob(GeneratorJobType job, MemberModel member, ClassModel inClass, String itemLabel, Object data)
   at ASCompletion.Completion.GeneratorItem.get_Value()
   at PluginCore.Controls.CompletionList.ReplaceText(ScintillaControl sci, String tail, Char trigger)
   at PluginCore.Controls.CompletionList.HandleKeys(ScintillaControl sci, Keys key)
   at PluginCore.Controls.UITools.HandleKeys(Keys key)
   at PluginCore.Controls.UITools.HandleEvent(Object sender, NotifyEvent e, HandlingPriority priority)
```

### DIFF
--- a/External/Plugins/ASCompletion/Completion/ASGenerator.cs
+++ b/External/Plugins/ASCompletion/Completion/ASGenerator.cs
@@ -3239,11 +3239,17 @@ namespace ASCompletion.Completion
                             pos = startPos + line.Length;
                             break;
                         }
-                        pos = position;
-                        var lineFromPosition = sci.LineFromPosition(pos);
+                        var lineFromPosition = sci.LineFromPosition(position);
                         startPos = sci.PositionFromLine(lineFromPosition);
-                        line = sci.GetLine(lineFromPosition);
-                        line = line.Substring(0, pos - startPos);
+                        var tmpLine = sci.GetLine(lineFromPosition);
+                        tmpLine = tmpLine.Substring(0, position - startPos);
+                        if (string.IsNullOrEmpty(tmpLine.TrimStart()))
+                        {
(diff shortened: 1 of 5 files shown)


Neverbirth:
I think I've never used `Last()` before... safer to use `LastOrDefault()`, if the collection is empty it won't throw exception.

SlavaRa:
Then I would not have caught the error :)

Neverbirth:
Not the best way to find out in production with such an error heh. Depending on the case you can still catch and log this errors in a nicer way wrapping in proper try catch, checking if there is any element in a collection, check if the returned value of `LastOrDefault` is the default value, etc.

By the way, I just saw in `ParseFunctionParameters` this:

```
                            try
                            {
                                d = double.Parse(trimmed, CultureInfo.InvariantCulture);
                            }
                            catch (Exception)
                            {
                            }
```

Why don't you use `if (!double.TryParse())`?

Neverbirth:
And maybe the `trimmed.Equals("true") || trimmed.Equals("false")` check could be first before trying the conversion to double?

SlavaRa:
I'll fix it a little later

PR #1614, opened 2017-05-06
Fixed ArgumentNullException

SlavaRa:
```
Parameter name: input
   at System.Text.RegularExpressions.Regex.Match(String input)
   at HaXeContext.PluginMain.HandleEvent(Object sender, NotifyEvent e, HandlingPriority priority)
```

### DIFF
--- a/External/Plugins/HaXeContext/PluginMain.cs
+++ b/External/Plugins/HaXeContext/PluginMain.cs
@@ -172,19 +172,19 @@ namespace HaXeContext
                     break;
                 case EventType.Trace:
                     if (settingObject.DisableLibInstallation) return;
-                    var nameToVersion = new Dictionary<string, string>();
-                    var length = TraceManager.TraceLog.Count - logCount;
-                    for (var i = 0; i < length; i++)
+                    var patterns = new[]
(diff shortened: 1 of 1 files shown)


External/Plugins/HaXeContext/PluginMain.cs
```diff
@@ -172,19 +172,19 @@ public void HandleEvent(Object sender, NotifyEvent e, HandlingPriority priority)
                     break;
                 case EventType.Trace:
                     if (settingObject.DisableLibInstallation) return;
+                    var patterns = new[]
+                    {
+                        "Library \\s*(?<name>[^ ]+)\\s*?(\\s*version (?<version>[^ ]+))?",
+                        "Could not find haxelib\\s*(?<name>\"[^ ]+\")?(\\s*version \"(?<version>[^ ]+)\")?"//openfl project
+                    };
```
Neverbirth: Why not making the regular expressions compiled instead of having them run for every trace message?

SlavaRa: Thanks, I will fix this in the near future

PR #1619, opened 2017-05-07
[ASGenerator] Improved Assign statement to var

SlavaRa:


SlavaRa:
@Neverbirth please review this PR and merge

External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -2667,20 +2673,15 @@ private static ASResult EvalExpression(string expression, ASExpr context, FileMo
...
                     head = EvalVariable(token, context, inFile, inClass);
                 }
             }
-            else if (token == "\"") // literal string
-            {
-                head = new ASResult();
-                head.Type = ASContext.Context.ResolveType(ASContext.Context.Features.stringKey, null);
-            }
-            else if (token.Contains("<")) head = new ASResult {Type = ASContext.Context.ResolveType(token, inFile)};
+            else if (token.Contains("<")) head = new ASResult {Type = ctx.ResolveType(token, inFile)};
```
Neverbirth: You added some lines above `else if (features.hasE4X && token == "</>") ` so it looks like it will be overriden by this line?

SlavaRa: It is needed to process new <int>...
I will make it more obvious later, otherwise there will be a lot of changes now

External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -3465,26 +3476,22 @@ private static ASExpr GetExpression(ScintillaControl sci, int position, bool ign
                         sb.Insert(0, "RegExp.#" + (subCount++) + "~");
                     }
                     // array access
+                    else if (c == ']') arrCount++;
```
Neverbirth: This change is strange... `if (c == '[')` can happen at any time but not the opposite? Why?

External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -3540,27 +3547,75 @@ private static ASExpr GetExpression(ScintillaControl sci, int position, bool ign
...
                         }
-                        braceCount++;
+                        parCount++;
                     }
                     else if (c == '>')
                     {
-                        if (haXe && position - 1 > minPos && (char)sci.CharAt(position - 1) == '-') { }
+                        if (haXe && position - 1 > minPos && (char) sci.CharAt(position - 1) == '-')
+                        {
+                        }
```
Neverbirth: ? Completely empty block.

External/Plugins/ASCompletion/Completion/ASComplete.cs
SlavaRa: This is necessary for the correct counting of parentheses, otherwise problems arise, this code already works in the ASGenerator.GetStartOfStatement In the future I want to get rid of it and keep the starting position in ASExp

External/Plugins/ASCompletion/Completion/ASComplete.cs
SlavaRa: yes

SlavaRa:
What about this?

Neverbirth:
I want to check in depth some parts of it :/. I think it will get fully reviewed by the end of tomorrow.

SlavaRa:
Ok, thanks

SlavaRa:
what's up

SlavaRa:
@Neverbirth Thanks

Neverbirth:
@SlavaRa The changes to ASComplete in this PR broke some stuff... for example trying to resolve some class is returning std\Class.hx instead of the class itself.

Neverbirth:
If it was not clear, I mean something like having

```
function test(class:Class<Dynamic>) {}

test(SomeClass);
```

`SomeClass` resolves to Class, not to SomeClass.

SlavaRa:
@Neverbirth thanks, I will fix it in the near future

SlavaRa:
@Neverbirth Could you create issue?

Neverbirth:
Done.

Issue #1601, opened 2017-05-02
Linter panel feedback


markknol:
I have a new "linter" panel, just wanted to share my feedback:
1.  The panel should also be in the "view" top menu, to show/hide them I'm my layout when I want. Now it just appear. 
1. The unused import/variables warnings are really nice/helpful in general!
1. The unused import doesn't work when you have conditional compilation `#if macro`
1. In macro code, I get fairly random errors, I think it doesn't understand this context at all. 
1. I also have projects where variables are generated by macros, It doesn't seem to understand that. Are compiler services used for this?
1. The results are send to both output and the linter panel. Maybe the linter panel is enough?
1. The errors/warnings don't go away after I save/resolve them. While I have my setting for this turned on.
1. The errors are empty, see screenshot
![image](https://cloud.githubusercontent.com/assets/576184/25609227/9956af20-2f1e-11e7-80aa-3dd7f6dd3c4e.png)

I use dev build  https://github.com/fdorg/flashdevelop/commit/0709ecbc3816c0c5b34886a2188f25c1f08a0444

Neverbirth:
Thanks for your feedback!

1. Yeah, this is in progress.
2. Thanks again!
3. The information is coming from the Haxe compiler service... so I'm afraid this is related to them...
4. The same as 3 I guess.
5. Same as 3 again.
6. This was done on purpose to follow the same behavior as compiler output, find references, refactor, etc, but it could be changed if more people are against it.
7. They should go away after you save, if they don't then there may be an error here.
8. @chipshort ?

markknol:
6. In that case, it would be nice if there was an option to ignore the errors/warnings (from the settings). Now you can hide them by deselecting the icon in the panel, but it still logs to output panel.

chipshort:
Okay, so here is what I think:
3. I could imagine this being a problem with the way we call the compiler service, since we specify the target, but I'm not completely sure.
For 8. It seems like there are more diagnostics types than I know of. At the moment, only unused variable and unused import are implemented. I recently found out there is one for cases in a switch that will never match, but there does not seem to be any documentation for any of them.
I guess, for now, I will make sure unsupported results are not shown.

Neverbirth:
@markknol Could you try the build for PR #1627?

https://ci.appveyor.com/api/buildjobs/yf46b9shf9wipe6f/artifacts/FlashDevelop%2FInstaller%2FBinary%2FFlashDevelopPR_1627.zip

[...]

PR #1623, opened 2017-05-08
ResultsPanel Improvements v2

### LINKED ISSUE #1596: New panels Refactor/References
In latest update appears 2(maybe more) panels Refactor/References, their purpose not clearly understand - early all showed in Results panel and works ok.
In addition they don't save state where it pinned, and after restart appears in default position.
Also they don't available at View menu.

wise0704:
- Added `KeepResultsByDefault` setting.
- View Menu for multiple panels as a drop down.
  - Note that hidden panels are present in the drop down list. However closed panels are not.
- Sorting improved:
  - You can now sort by line numbers.
  - Sorting by line numbers or descriptions does not group the entries.
    - This is because their values are likely to be all unique, thus it will only create groups with only one or two entries otherwise.
- `DefaultGrouping` setting does not affect panels that already exist anymore.
- `HighlightOnlyActivePanelEntries` setting defaults to `false`.
- Bug fixes:
  - `RefreshSquiggles()` should respect the settings.
  - Only refresh when related settings have changed.

Closes #1596 
Related to issue 1 in #1601

### DIFF
--- a/External/Plugins/ResultsPanel/Helpers/ResultsPanelHelper.cs
+++ b/External/Plugins/ResultsPanel/Helpers/ResultsPanelHelper.cs
@@ -160,7 +160,7 @@ namespace ResultsPanel.Helpers
                 }
             }
         }
-        
+
         /// <summary>
         /// Creates a new results panel.
         /// </summary>
@@ -205,7 +205,7 @@ namespace ResultsPanel.Helpers
                 var traceGroup = TraceManager.GetTraceGroup(groupId);
                 ui.GroupData = groupData;
                 ui.Text = string.Format(traceGroup.Title ?? TextHelper.GetString("Title.PluginPanel"), args);
-                ui.ParentPanel.Text = ui.Text; 
+                ui.ParentPanel.Text = ui.Text;
             }
         }
 
--- a/External/Plugins/ResultsPanel/Helpers/ResultsPanelHelper.cs
+++ b/External/Plugins/ResultsPanel/Helpers/ResultsPanelHelper.cs
@@ -102,8 +102,10 @@ namespace ResultsPanel.Helpers
             foreach (var pluginUI in pluginUIs)
             {
                 pluginUI.ApplySettings();
+                pluginUI.ClearSquiggles();
             }
-            
+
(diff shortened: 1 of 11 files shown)


Neverbirth:
> Note that hidden panels are present in the drop down list. However closed panels are not.

What is a closed panel? because if I press the X button I see the panels still in the dropdown and can open them. But for me it's fully OK.

Neverbirth:
By the way, I don't agree that grouping for Description is going to mostly have single entries for most groups, at least on the kind of projects I tend to work I often get tons of groups with the same messages and I often sort by it so I can first concentrate of problems of the same kind.

But again, I'm ok with keeping your change, not needed to revert it.

wise0704:
By default, all panels are hidden automatically when X is pressed.

However, because of the layout data, multiple hidden panels with the same persist string will of course cause multiple panel definitions with the same persist string to be written in the XML. This doesn't really cause problems other than having more lines saved in the layout data. When the XML is read, I ensured that only the first panel data with the same persist string is used and others are ignored.

What I've done is, that
- At least one panel exists (not closed) at any given time. This is to ensure that the panel layout data with the persist string (group id) is written in the XML.
- A panel stays hidden only when there isn't any panels with the same persist string. This is to ensure the above.
- The hidden panel is closed when a new panel with the same persist string is open. This is to ensure that the layout data of invisible panels - which are closed in the perspective of the user -  are not saved in the XML.

Also, when the panels don't get closed when X is pressed, it will eventually lead to a growing number of panels that can't be closed, unless another method to explicitly close these panels is implemented.

This means that, if you open multiple Find All References window, and close them all, the panel closed last will be hidden, and the rest are closed. In the view menu, you'll only be able to access the lastly closed panel.

One way to slightly change the behaviour is to:
- Prevent closing panels that have 'Keep Results' toggled.
- Prevent showing hidden panels in the View menu that don't have 'Keep Results' toggled.

This will result in the ultimate effect of the X button to be clearer to the users:
- Closing locked panels mean hiding the panels, and the panels can be re-opened using View menu.
- Closing unlocked panels mean literally closing panels.

PR #1631, opened 2017-05-20
Refactoring for highlights management

wise0704:
Instead of directly accessing scintilla highlighting api from LintingHelper, let ResultsPanel manage all highlighting.

Also change the default colour of warning (`DebugLineBack`) to orange from yellow, since it was barely visible with a default white background.

### DIFF
--- a/External/Plugins/HaXeContext/Completion/HaxeComplete.cs
+++ b/External/Plugins/HaXeContext/Completion/HaxeComplete.cs
@@ -258,7 +258,6 @@ namespace HaXeContext
                         Errors = lines;
                         return HaxeCompleteStatus.ERROR;
                     }
-                    break;
                 default:
                     try
                     {
--- a/External/Plugins/LintingHelper/Managers/LintingManager.cs
+++ b/External/Plugins/LintingHelper/Managers/LintingManager.cs
@@ -161,7 +161,7 @@ namespace LintingHelper.Managers
             var cachedResults = Cache.GetAllResults();
             foreach (var result in cachedResults)
             {
-                string chars = null;
+                string chars;
                 if (result.Length > 0)
                 {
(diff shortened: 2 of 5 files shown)


External/Plugins/LintingHelper/Managers/LintingManager.cs
```diff
@@ -172,60 +161,46 @@ static void ApplyLint(string[] files, string language, List<LintingResult> resul
...
+                        break;
+                    default:
+                        continue;
+                }
+                TraceManager.Add(message, state, TraceGroup);
             }
 
             PluginBase.RunAsync(() =>
             {
                 PluginBase.MainForm.CallCommand("PluginCommand", "ResultsPanel.ShowResults;" + TraceGroup);
```
wise0704: In my opinion, this line should be better deleted.
Currently, the linter results panel will jump up every time linting is done (file save, file open), even if the user closed the panel, which should then be accessible using the view menu.

If this line is deleted, on the very first creation of the linter results panel, the panel will be visible. Then, it won't be "forced shown". In other words, it doesn't jump up, nor shows when manually closed by the user.

Neverbirth: Sounds ok to me.

External/Plugins/HaXeContext/Completion/HaxeComplete.cs
```diff
@@ -258,7 +258,6 @@ HaxeCompleteStatus ParseLines(string lines)
                         Errors = lines;
                         return HaxeCompleteStatus.ERROR;
                     }
-                    break;
```
Neverbirth: I know it's not needed, but feels strange to remove it to me. Anyway, not really important I guess.

wise0704: vs gives you the unreachable code warning though

PR #1632, opened 2017-05-20
AssemblyInfo consistency

wise0704:
These changes wouldn't cause any problems, would they?


Neverbirth:
I didn't see any change that could cause problems, but I leave this one to @Meychi

PR #1633, opened 2017-05-20
Scintilla MoveLineUp/Down rewritten

### LINKED ISSUE #1626: Move line up/down buggy with Arrays of literal objects
Haxe code:
```haxe
class Main 
{	
	static function main() 
	{
		var a:Array = [
			{value: 'a'},
			{value: 'b'},
			{value: 'c'},
			{value: 'd'},
			{value: 'e'},
		];
	}	
}
```

Put the cursor on line with value `e`, press Ctrl-Alt-Up.
Expected result: object values in order a,b,c,e,d
Actual result: object values in order a,b,d,e,c

Problem seems to be triggered by curly braces.

wise0704:
- Make the operation simpler by not taking the language syntax into account, also avoiding numerous bugs. Closes #1626.
- Keep the current cursor selection after the operation.

Things to note:
- The last line is not movable.
- Our version of scintilla does not store cursor position in undo history, meaning undo will set the cursor to the first operation, which is `Clear()`, rather than back to the original position.
  - After moving `aaaaaa` down: 
image
Pressing undo will put cursor after `bbbbbb`:
image
  - After moving `bbbbbb` up:
image
Pressing undo will put cursor after `aaaaaa`:
image
  - Redo is equally weird.

### DIFF
--- a/PluginCore/ScintillaNet/ScintillaControl.cs
+++ b/PluginCore/ScintillaNet/ScintillaControl.cs
@@ -6935,7 +6935,7 @@ namespace ScintillaNet
         /// </summary>
         public void MoveLineUp()
         {
-            MoveLine(-1);
+            this.MoveLine(-1);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


Neverbirth:
I guess current implementation tries to be smart and move up/down selected members/blocks of code? we should include some operation like this, I already got the request in the past, but it's not so trivial to get correctly every edge case fine.

wise0704:
The thing is I can always just select the desired lines myself and then move up/down. Just like in VS, when I want to move a specific line, I expect just that line to move.

For e.g. I want to move the open brace up above the if,
```
...
if (x)
{
   ...
}
```
but it moves the if too. If i wanted that, I would've selected the two lines.

Meychi:
@Neverbirth, good to merge?

PR #1637, opened 2017-05-23
Apply theme to the Resources dialog

### LINKED ISSUE #1179: "Open resource" dialog - add theme support
The `Project > Open resource` dialog remains one of the few commonly-used features lacking theme support.
It would look better if it supported themes (e.g. `Type Explorer` does).

wise0704:
Closes #1179

### DIFF
--- a/External/Plugins/ProjectManager/Controls/OpenResourceForm.cs
+++ b/External/Plugins/ProjectManager/Controls/OpenResourceForm.cs
@@ -153,12 +153,7 @@ namespace ProjectManager.Controls
         /// </summary>
         private void InitializeGraphics()
         {
-            ImageList imageList = new ImageList();
-            imageList.ColorDepth = ColorDepth.Depth32Bit;
-            imageList.ImageSize = ScaleHelper.Scale(new Size(16, 16));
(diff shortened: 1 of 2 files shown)


wise0704:
This is the same pattern I used for clipboard history dialog too.

Meychi:
Could this be done to all basic dialogs at the same time?

wise0704:
This pattern to all basic dialogs individually?
Or something like override `Form.OnShown` in `SmartForm` and handle it there?
One small problem to the latter is that form back colours and label&checkbox fore colours are not defined in the theme files, so I have to manually fetch `"MenuStrip.BackColor"` and `"MenuStrip.ForeColor"` as done in this PR, and loop through the child controls to apply them.

Meychi:
Hmm, maybe we should add proper support for basic forms with SmartForm and define the missing colors to the themes. We can default to current functionality by adding a setting that enables global coloring. Also we need to remember to add handling for theme event when the theme gets applied.

Do you have time for this or should i jump on it?

wise0704:
No I don't really have time...

Meychi:
Ok, i will check this.

Meychi:
See #1719

PR #1642, opened 2017-05-31
Fix error accessing disposed scintilla when document closed during linting

wise0704:


wise0704:
@Neverbirth maybe check 0564819 separately...

PR #1644, opened 2017-05-31
Haxe formatter

chipshort:
Adds a settings panel for Haxe formatting using AStyle.


Neverbirth:
Some initial comments:

- If you don't have any file open it will throw an exception.
- Maybe loading the current open file as a preview of the formatter should be optional and we should display some default files with the minimum amount of code to let the user see the effects of the formatting changes.
- When changing a setting it's annoying that the displayed file scroll backs to the top.
- Not too important right now: The location of the menu it's not the best IMHO. This should be located in the settings panel I think.

I guess it's still missing tons of options... I never got to check AStyle tho. Did you check latest AStyle version?

wise0704:
My comments:
 - I know the dialog is pretty nice, with the preview and stuff, but imo the whole setting should be in the settings dialog like other language settings (AS3 and MXML). If you want to keep the dialog, though, there's an option. Make the whole settings into a serialisable class object, and make the dialog the editor using the `EditorAttribute` attribute (see [example](https://github.com/fdorg/flashdevelop/blob/development/External/Plugins/ASCompletion/Settings/GeneralSettings.cs#L439)).
    - This also means no extra file to save the settings.
 - I believe we try to keep control/form classes in one file; that is, we don't need that empty resource file, and usually merge the contents of the .Designer.cs in the actual file.
 - On the same page as what @Neverbirth said; maybe instead of using the current file, it's better to use a predefined snippet, which has all the configurable parts of the code present (i.e. has the codes that change for every setting changes, so we know which setting does which changes).
 - Conflicts with the first point I've said, but I think the shortcut Ctrl+Shift+3 doesn't make sense. Also that it's between "Code Formatter" and "Code Generator" doesn't make sense either.

External/Plugins/CodeFormatter/Dialogs/HaxeAStyleDialog.cs
```diff
@@ -0,0 +1,427 @@
...
+            { "Banner", "banner" },
+            { "GNU", "gnu" },
+            { "Linux", "linux" },
+            { "Horstmann", "horstmann" },
+            { "One True Brace", "otbs" },
+            { "Google", "google" },
+            //{ "Mozilla", "mozilla" }, //not supported by old version of AStyle
+            { "Pico", "pico" },
+            { "Lisp", "lisp" },
+        };
```
wise0704: I'd say try to avoid this design pattern. It just keeps the dictionary permanent in the memory during the lifetime of the program.
Using `switch` statement is much better. The compiler decides whether to use `if-else` statements, or a one-time dictionary, depending on which it believes is faster.
Even a manual one-time dictionary every time the mapping is needed is still better, since instantiation doesn't cost much, and is better than having dictionaries stuck in the memory.

External/Plugins/CodeFormatter/PluginMain.cs
```diff
@@ -234,7 +241,15 @@ public void SaveSettings()
...
         private const int TYPE_CPP = 3;
         private const int TYPE_UNKNOWN = 4;
-        
+
+        /// <summary>
+        /// Opens the Haxe AStyle settings dialog.
+        /// </summary>
+        public void ShowSettings(Object sender, EventArgs e)
+        {
+            new HaxeAStyleDialog().ShowDialog();
```
wise0704: This needs a `using` block.

External/Plugins/CodeFormatter/CodeFormatter.csproj
```diff
@@ -71,26 +71,49 @@
...
+    </ProjectReference>
     <ProjectReference Include="..\..\..\PluginCore\PluginCore.csproj">
       <Project>{61885F70-B4DC-4B44-852D-5D6D03F2A734}</Project>
       <Name>PluginCore</Name>
       <Private>False</Private>
     </ProjectReference>
+    <ProjectReference Include="..\AirProperties\AirProperties.csproj">
+      <Project>{F2CED639-1DA7-489D-9DBB-909F7B590C3D}</Project>
+      <Name>AirProperties</Name>
+    </ProjectReference>
```
wise0704: These references seem unnecessary... I can compile with all of the newly added ones removed (namely, ProjectManager, System.Data, System.Web, System.Xml, FlashDevelop(why was this added?), and AirProperties).

chipshort: Not sure, where those come from. I guess I was testing some stuff. I will remove them.

chipshort:
@Neverbirth
1 / 2. I think a predefined file probably makes more sense and will also prevent the problem with the exception.
3. I also noticed that, but didn't know how to fix it. I will have to look into it.
4. I think @wise0704 is right in that it should be in the settings dialog.
5. There aren't that many options for AStyle, especially if you keep in mind, that some of them do not work with Haxe and could potentially break code. I did check the latest version, it has a few more options, but not many. Back in the day, I could not compile it at work. I will look into it.

@wise0704 
1. I think the dialog should be kept, as it is really annoying to find out what the options do without immediate feedback, but putting it into the settings is a good idea.
2. I noticed that already. I will merge them when all missing options are added. It is just easier to edit everything using the designer.
3. (see above)
4. That was mostly just for testing. We won't need it anymore, after I add it to the settings.

wise0704:
>I noticed that already. I will merge them when all missing options are added. It is just easier to edit everything using the designer.

As long as you have the `InitializeComponents()` method, VS designer works even if you don't have the .Designer.cs file.

External/Plugins/CodeFormatter/Resources/AStyleExample.hx
```diff
@@ -0,0 +1,43 @@
...
+				trace("a"); trace("another statement");
+			case "b":
+			{
+				trace("b");
+			}
+		}
+
+		if (true 
+			&& false)
+			trace("Inconsistent system")
```
wise0704: semicolon?

Meychi:
Cound you check the conflicts?

SlavaRa:
WIP

PR #1648, opened 2017-06-02
Fix new results panel affecting other panel's layout

chipshort:
When a new DockablePanel is added, it shown immediately before it's DockState is set. This causes it to be added to the right (where the Outline, Project, etc. panel are in the default layout) and then moved to the correct place immediately. However, this means, the focus of the "tabbed" panels on the right will change to the first one (instead of where it was before).
This PR should fix this behaviour.

### DIFF
--- a/FlashDevelop/Docking/DockablePanel.cs
+++ b/FlashDevelop/Docking/DockablePanel.cs
@@ -24,7 +24,6 @@ namespace FlashDevelop.Docking
             this.HideOnClose = true;
             this.Controls.Add(ctrl);
             Globals.MainForm.ThemeControls(this);
-            this.Show();
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


Neverbirth:
Sadly I'm not sure this is the best solution because this is quite a breaking change... although maybe nobody creates DockablePanel directly? maybe you could create a constructor overload where you pass the DockState?

@wise0704 Do you have time to check?

wise0704:
DockablePanel is a part of FlashDevelop instead of PluginCore, so I don't think any 3rd party has gone against our rule of not referencing FD.exe and C#'s rule of avoiding referencing an exe.
Even if they did (although there is really no reason to) I don't think we should care about those disadvised cases.

I'm fine with the changes here.

Neverbirth:
That's why I was asking that maybe nobody was creating it directly. Fair enough.

PR #1649, opened 2017-06-03
Prevent dynamic persist panel from force-activating

wise0704:
Copy paste from the commit message:

>Panels created using CreateDockablePanel() are expected to be created
during plugin initialization, which is right before
LayoutManager.BuildLayoutSystems(). That method ensures that all panels
are in the correct place, and shows unregistered panels as floating,
essentially making the call to Show() in CreateDockablePanel()
unnecessary.
Panels created using CreateDynamicPersistDockablePanel() are also called
LayoutManager.SetContentLayout(), which ensures the panel is in the
correct place if registered. Adding Show() at the end ensures the panel
is shown when the panel information isn't stored in the layout file.

`dockablePanel.Show()` in `CreateDynamicPersistDockablePanel()` was causing the panel to be activated upon creation, even when the panel was registered.

### DIFF
--- a/FlashDevelop/MainForm.cs
+++ b/FlashDevelop/MainForm.cs
@@ -576,7 +576,6 @@ namespace FlashDevelop
             try
             {
                 DockablePanel dockablePanel = new DockablePanel(ctrl, guid);
-                dockablePanel.Show();
                 dockablePanel.Image = image;
                 dockablePanel.DockState = defaultDockState;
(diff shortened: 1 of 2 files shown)


Neverbirth:
I think this PR (if not #1648, didn't fully check) caused a problem: when running FD for the first time now the panels designed to go to the right by default are stacked instead of grouped:

image

PR #1651, opened 2017-06-05
[ASGenerator] Improved Generate Function. fixes #1645

### LINKED ISSUE #1645: [AS3] Wrong function generation behavior
Let's have
```
private function f1(f:Function):void{}

private function f2():void
{
	f1(f3);
}
```
when generate new function f3 we got
`private function f3():Function`
expected
`private function f3():void`

SlavaRa:


SlavaRa:
@Neverbirth please review this PR and merge :)

External/Plugins/ASCompletion/Completion/ASGenerator.cs
```diff
@@ -2846,67 +2846,75 @@ private static void GenerateFunctionJob(GeneratorJobType job, ScintillaControl s
...
                 else
                 {
-                    var parCount = 0;
-                    var braCount = 0;
-                    var genCount = 0;
-                    var startPosition = 0;
-                    var typeLength = parameterType.Length;
-                    for (var i = 0; i < typeLength; i++)
+                    var voidKey = ASContext.Context.Features.voidKey;
+                    if (parameterType == "Function") newMemberType = voidKey;
```
Neverbirth: What if the user creates a custom Function type in Haxe?

SlavaRa: Hmm.. Yes, I should fix it. Thanks

SlavaRa: Fixed

External/Plugins/ASCompletion/Completion/ASGenerator.cs
```diff
@@ -2841,72 +2841,91 @@ private static void GenerateFunctionJob(GeneratorJobType job, ScintillaControl s
             if (caller?.Parameters != null && caller.Parameters.Count > 0)
             {
                 Func<string, string> cleanType = null;
-                cleanType = s => s.StartsWith("(") && s.EndsWith(')') ? cleanType(s.Trim('(', ')')) : s;
+                cleanType = s => s.StartsWith("(") && s.EndsWith(')') ? cleanType(s.Trim('(', ')')) : s;
```
wise0704: I don't understand why the recursion is needed, since `Trim('(', ')')` would already make it that it doesn't start or end with '(' and ')'.

Maybe it's meant to be
```c#
s => s.StartsWith('(') && s.EndsWith(')') ? cleanType(s.Substring(1, s.Length - 2)) : s
```
or better without the recursion or useless string creations
```
s => {
    int start = 0, end = s.Length - 1;
    while (start < end && s[start] == '(' && s[end] == ')') {
        start++; end--;
    }
    return s.Substring(start, end + 1 - start);
}
```

Why is it a local delegate instead of a method though?

SlavaRa: @wise0704 Well, I'll make this place more obvious

PR #1652, opened 2017-06-05
Error tooltip

chipshort:


wise0704:
I like this change, letting results panel handle what it should. I haven't tested it yet however.

Just a few suggestions for performance...

In `Scintilla_OnMouseHover()`:
I think `PluginUI.GetResultsAt()` should take the `List<string>` as a parameter instead, and add results to it (it's an internal method anyway, so yeah just make it do the specific job we want efficiently without worrying about whether the name actually matches what it does). That will save list instantiation for each call, as well as `foreach` loop for every panel.
At the end, check for `results.Count > 0` instead, then do
```c#
string desc = string.Join(Environment.NewLine, results.ToArray());
```
I don't think `if (!string.IsNullOrEmpty(result))` is necessary, but if it is, it's better to move it to `PluginUI.GetResultsAt()`.

In `PluginUI.GetResultsAt()`:
I think `GetPosition()` would be better with `TryParse` pattern: return `bool`, and `start` and `end` as out parameters. That will save an array instantiation and array access.
After that just do `start += lineStart; end += lineStart`.
(By the way, I think you meant `lineStart` instead of `listStart`.)

wise0704:
Ah, in `GetResultsAt()`, you need to check for the entry types, and only show warnings and errors. Find and replace in files > "i" gives me a chaos.

wise0704:
You also need to check for the file name.

PR #1653, opened 2017-06-05
EventManager refactored

wise0704:
Fixes the bug on `RemoveEventHandler()` which checks for equality of two instances of `EventObject` which would never be `true`.
Also, the `try...catch` block is moved to wrap each `HandleEvent()` calls to prevent an exception from stopping the whole event dispatch process.
Includes some more refactoring.

### DIFF
--- a/PluginCore/PluginCore/Managers/EventManager.cs
+++ b/PluginCore/PluginCore/Managers/EventManager.cs
@@ -1,68 +1,29 @@
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 
 namespace PluginCore.Managers
 {
-    public class EventManager
+    /// <summary>
+    /// Provides an application-wide event management methods.
+    /// </summary>
+    public static class EventManager
     {
-        /// <summary>
-        /// Properties of the class
-        /// </summary>
-        private static List<EventObject> highObjects;
-        private static List<EventObject> normalObjects;
-        private static List<EventObject> lowObjects;
+        private const int PriorityCount = 3;
+        private static List<EventObject>[] eventObjects;
 
-        /// <summary>
-        /// Static constructor of the class
-        /// </summary>
         static EventManager()
         {
-            highObjects = new List<EventObject>();
-            normalObjects = new List<EventObject>();
-            lowObjects = new List<EventObject>();
-        }
-
-        /// <summary>
-        /// Gets the event lists as an array
-        /// </summary>
-        private static List<EventObject>[] GetObjectListCollection()
-        {
-            List<EventObject>[] collection = new List<EventObject>[3];
-            collection.SetValue(highObjects, 0);
-            collection.SetValue(normalObjects, 1);
-            collection.SetValue(lowObjects, 2);
-            return collection;
-        }
-
-        /// <summary>
-        /// Adds a new event handler with a specific HandlingPriority
-        /// </summary>
-        public static void AddEventHandler(IEventHandler handler, EventType mask, HandlingPriority priority)
-        {
-            try
+            eventObjects = new List<EventObject>[PriorityCount]
             {
-                EventObject eo = new EventObject(handler, mask, priority);
-                switch (priority)
-                {
-                    case HandlingPriority.High:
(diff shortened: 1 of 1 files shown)


Meychi:
These kind of internal changes need extensive testing with performance and validation that nothing can break in plugins and in other 3rd party code.

wise0704:
I've been using this version of EventManager for quite a while, while I was working on other branches.

Behaviour changes:
 - `AddEventHandler()` will now throw an `InvalidEnumArgumentException` for invalid `HandlingPriority` values.
 - `RemoveEventHandler()` will now actually remove the handler. Previously both signatures used `List.Remove()` which would fail with the equality check.
 - `DispatchEvent()` continues the loop after exception. This part would need more testing?
 - `DispatchEvent()` checks `(obj.Mask & e.Type) == e.Type` rather than `(obj.Mask & e.Type) > 0`.
    - I don't actually need this change now: I needed this pattern for a "deprecated event", where for e.g. `EventA=0b0001` is deprecated and `EventB=0b1001` is used instead, but `EventA` handlers should handle `EventB` whereas `EventB` handlers shouldn't handle `EventA`. Since I don't actually need it now, it can just be changed back to `(obj.Mask & e.Type) > 0`.

Performance changes:
 - `AddEventHandler()` has an extra array index operation with an `Enum` to `int` conversion.
 - `DispatchEvent()` skips an array creation & initialisation. Loops with a constant as the length rather than `Array.Length`.

Neverbirth:
> `DispatchEvent()` checks `(obj.Mask & e.Type) == e.Type` rather than `(obj.Mask & e.Type) > 0`.

This sounds like a potential regression, isn't it?

> `DispatchEvent()` skips an array creation & initialisation. Loops with a constant as the length rather than `Array.Length`.

Looping with a constant value is in this case (or implementation) actually bad. Some handler could unregister itself in the middle, throwing an exception that would be now unhandled.
You could seal the collection but I would say it's also prone to errors (you would also need to have a global try catch and/or locking system because of multi threading, and you need extra logic in most of the other methods).
I would say that the best option would be to work with a copy of the list.

wise0704:
>I'm surprised nobody ever had issues with this.

Well because`RemoveEventHandler()` never worked :P

I agree. Now that listeners can be removed, maybe creating a snapshot of the list is a good idea.

Neverbirth:
Although maybe the snapshot can be optimized with just setting some flag on removing and after the dispatching check it? I wonder if using some of the blocking collections of .NET 4 would also be a solution..

wise0704:
I don't know if there is a more optimised way... 
Also, addition of items need to be considered - they should also be applied after dispatching like the removal. 

Maybe some ideas:
 1. Have two list variables and a boolean "dispatching" flag
    - One keeps a list of event objects to be added after the dispatching, and one keeps a tuple of two integers (i, j) of the indices of objects to be removed after the dispatching.
    - The dispatching flag is checked during adding/removing event handlers.
 2. Make the iterators used during dispatching into variables instead.
    - Removing event handlers will check the iterator values. If removed item is after the iterator, do nothing. If before, decrease the iterator.
    - This changes the behaviours: the effect of adding and removing event handlers will be applied immediately during event dispatching.

Neverbirth:
I was thinking of something else when I made my comment. Right now there is this:

```
            var eventObjects = new EventObject[highObjects.Count + normalObjects.Count + lowObjects.Count];
            highObjects.CopyTo(eventObjects);
            normalObjects.CopyTo(eventObjects, highObjects.Count);
            lowObjects.CopyTo(eventObjects, highObjects.Count + normalObjects.Count);
```

So what I was thinking is having a class level flag, when adding or removing some listener set the flag, and only instantiate the objects when the flag is true, and revert.

But I'm wondering about multithreaded scenarios... but that one should have happened already.

wise0704:
Could you elaborate? What is the flag for? I don't quite understand:
>only instantiate the objects when the flag is true, and revert.

Neverbirth:
For me this would be OK to merge, let's see what @Meychi says... but I think I will make some deadline of some days...

PR #1654, opened 2017-06-06
Prevent documents from closing during layout changes

### LINKED ISSUE #1014: Cannot undo after build
I can't undo to previous task after build project. If I change code and then undo, it success, if i change code, build project and then undo, it failed.

wise0704:
`EventType.RestoreSession` will no longer be dispatched when changing layout, but I don't see why session needs to be "restored" during layout changes anyway.
@Meychi 

Closes #1014


Neverbirth:
@Meychi I'm ok with merging this, what do you say?

PR #1655, opened 2017-06-06
[ASGenerator] Improved Generate Function. Adds return statement with default value to body of function if needed and posible

SlavaRa:


External/Plugins/AS3Context/Context.cs
```diff
@@ -979,6 +979,19 @@ protected override void InitTopLevelElements()
...
+        public override string GetDefaultValue(string type)
+        {
+            if (string.IsNullOrEmpty(type) || type == features.voidKey) return null;
+            switch (type)
+            {
+                case "int":
+                case "uint": return "0";
+                case "Number": return "NaN";
+                case "Boolean": return "false";
+                default: return "null";
```
wise0704: When type is empty or `features.dynamicKey`, shouldn't it be "undefined"?

SlavaRa: For features.dynamicKey "undefined" is ok
But I doubt that it's good for an empty type

wise0704: In AS3 empty type and * are the same:
`var a:*` is the same as `var a;`

SlavaRa: hmm okay :)

SlavaRa: > In AS3 empty type and * are the same:
>var a:* is the same as var a;

I will do this individual PR next week

SlavaRa:
@Neverbirth what do you think about this PR?

Neverbirth:
I think this should be a setting. Some developers (me included) don't like this kind of default implementation as it may make code to implement to be forgotten, others prefer to throw exceptions as default implementations, etc.

image

Neverbirth:
In case it was not clear, I'm Ok with merging this, but after some setting is added for it. For instance, I know some of my colleagues would be happy with it, but again, not everyone.

SlavaRa:
Well, I'll try to add this setting today or tomorrow after the main job

SlavaRa:
@Neverbirth done, please check

External/Plugins/ASCompletion/Completion/ASGenerator.cs
```diff
@@ -2961,17 +2960,19 @@ private static void GenerateFunction(MemberModel member, int position, bool deta
...
-                    {
-                        var expr = inClass.InFile.Context.ResolveType(type, inClass.InFile);
-                        if ((expr.Flags & FlagType.Abstract) != 0 && !string.IsNullOrEmpty(expr.ExtendsType))
-                            type = expr.ExtendsType;
-                    }
-                    var defaultValue = inClass.InFile.Context.GetDefaultValue(type);
-                    if (!string.IsNullOrEmpty(defaultValue)) body = $"return {defaultValue};";
+                    case GeneratedMemberBodyStyle.ReturnDefaultValue:
+                        var type = member.Type;
+                        if (inClass.InFile.haXe)
```
Neverbirth: So `if ((member.Flags & FlagType.Constructor) == 0)` is not needed in the end?

PR #1662, opened 2017-06-19
Add EndAtLastLine setting to FlashDevelop Editor settings

### LINKED ISSUE #835: It's possible to scroll a file down too far down
A file can be scrolled down until it's last line is the only line visible on the screen (as first line), with empty space (and no line numbers either) covering the rest of the screen.
I think at the bottom-most scroll position the last line of the file should be at the bottom of the screen, not at the top.

image1

image2

FD 5.0.1 
.NET CLR 2.0.50727

http://www.flashdevelop.org/community/viewtopic.php?f=6&t=12362

gene-pavlovsky:
Closes #835 

Adds a new `EndAtLastLine` setting, controlling Scintilla's setting with the same name. Allows to disable over-scroll mode, introduced by [b21e2de](https://github.com/fdorg/flashdevelop/commit/b21e2de250b234ca29609979d73c8a0c870a733d).

Notes: 
- Needs translations for all languages other than `en_US`.
- Why `ScintillaControl.EndAtLastLine` is an `int`? The [Scintilla API](http://www.scintilla.org/ScintillaDoc.html#SCI_SETENDATLASTLINE) I could find lists it as a `bool`.


PluginCore/PluginCore/Resources/en_US.resx
```diff
@@ -6189,4 +6189,7 @@ AND - All conditions are met.</value>
     <value>Keep result panels locked by default.</value>
     <comment>Added after 5.2.0</comment>
   </data>
+  <data name="FlashDevelop.Description.EndAtLastLine" xml:space="preserve">
+    <value>Sets the scroll range so that maximum scroll position has the last line at the bottom of the view. Setting this to false allows scrolling one page below the last line.</value>
+  </data>
```
wise0704: missing comment

gene-pavlovsky: It's strange, I was sure I put a comment in there...

Neverbirth:
Custom scrollbars are not rendered correctly when changing the setting.

gene-pavlovsky:
Oh no! What exactly is the problem? Screenshot?

Neverbirth:
I deleted the build, but it looks like the custom scrollbar is fixed somehow to work in the current way, so the thumb size includes the extra page and scrolling with the keyboard to the bottom of a document doesn't move it to the bottom as expected.

gene-pavlovsky:
Remaining issue (but also in current build):
- On certain monitors the thumb is not aligned with the arrows. Probably related to ScaleHelper and different DPI.

Neverbirth:
I'm not really sure we need the viewportsize but this seems ok to me.

PR #1663, opened 2017-06-21
use the android sdk from flex sdk

matrix3d:


Neverbirth:
I guess this would be a regression for old SDKs? Because I never got problems when I used to make iOS and Android apps with FD :S.

Neverbirth:
Or maybe if the one in FlexSDK doesn't exist try out the old one in $(AppDir)\Tools\android

matrix3d:
the bat in the project use the fd android sdk.will get error.when copy the project file to others.
the others fd location maybe not the same.

PR #1666, opened 2017-06-29
Fixed wrong autocompletion. fixes #1664

### LINKED ISSUE #1664: Wrong autocompletion
This bug has come to my knowledge today. Having this AS3 code (I guess Haxe is similar but didn't check):

```
package {
	public class NewClass {
		function test():void
		{
		}
		
		func
	}
}
````

Trying to autocomplete after `func` is not working nicely, after checking in depth I've seen that the changes to `ASComplete` in #1619 also broke this, and `ASComplete` is returning that the tail of the expression in that position is `{}func` instead of just `func`.

Pinging @SlavaRa in case he is not aware of this case.

SlavaRa:


Neverbirth:
The changes to `HandleDeclarationCompletion` are unrelated, aren't they? I hope I will have the time to look into this today.

SlavaRa:
No, they are not related to HandleDeclarationCompletion

PR #1673, opened 2017-06-30
Improved Ctrl+R dialog

chipshort:
I actually did this quite a while ago.
I'm not so sure about the tests though. Comments please?


gene-pavlovsky:
Hey man, it's nice to see this being improved (I've requested fuzzy searching quite a while ago, but never got around to actually do it myself). Here are my comments:

- ProjectManager.dll in root folder, causing exceptions when opening a (Haxe) project, and an exception about loading ProjectManager.dll when restarting FlashDevelop
- In general, fuzzy searching seems to be working quite well
	+ Sort order is not clear. I suppose it's by relevance? But e.g. in this screenshot it's really not obvious why the results are sorted this way
	+ Mixed case search query should be improved
		* A major feature in the old Ctrl+R dialog is lost - being able to search for a camel-case-named file by typing only the capital letters (e.g. LOV should offer LastOnlineView before LoginView)
		* Searching for same query with lowercase and uppercase gives slightly different results, but the sort order is really not obvious
- (also on current release) If a file has a common name, but there are many of them in different folder (config.xml, Makefile, etc.). It's not possible to further filter by path.
- (also on current release) Type a search query, switch to another application (via Alt-tab or by clicking on FlashDevelop) - the query is cleared (bug or feature?)
 
 It's a great effort, but I think it should be improved before it's good enough to replace the old behavior (biggest gripe is losing the feature to search for camel-case-named files using first capital letters).
 Proposals/suggestions:
- Check out [fzf](https://github.com/me-and/fzf) and it's [Search syntax](https://github.com/me-and/fzf#search-syntax) for inspiration
	+ Smart case: Match case-sensitively if there are any uppercase letters in search term, case-insensitively otherwise - default in both fzf and ag (aka the silver searcher)
		* Note: this alone should solve my biggest gripe (LOV will match LastOnlineView, but not LoginView)
	+ Add (optional) flags, which are special characters at the beginning of a search term (multiple search terms, if necessary, are separated with a space). Proposed flags:
		* Default: fuzzy match
		* ^: search term should match at the beginning of the string (e.g. ^mot if you want Motivation but not InitMotivation)
		* $: search term should match at the end of the string (e.g. ss$ if you want to find .scss or .css files)
		* ': exact match (not fuzzy, consecutive letters) - for cases when fuzzy matching produces many extraneous results even when you already type exactly what you need
		* !: inverse fuzzy match (to filter out unwanted results)
		* /: search within file's dirname rather than basename
		* Most flags may be combined (e.g. !^'Mot filters out files starting with literal "Mot")
	+ Note: For simple queries, no flags and no spaces are necessary, so these extra features make life easier if you want to use them, and have no effect otherwise

gene-pavlovsky:
Cool! Much better now with support for exact matching of uppercase letters.
No more Project Manager.dll problem, and the query stays now, when switching to another app and back.
I'll play more with it but I think it's now better than the previous one.
Still missing somewhat the ability to narrow down by folder.

PR #1679, opened 2017-07-19
Update appman.xml: latest Flex SDK + AIR SDK, AIR SDK + ASC 2.0, Adob…

gene-pavlovsky:
…e AIR, Flash Player (*), TortoiseGit, Git For Windows, TortoiseSVN, SlikSVN

Note 1: here's the updated [frameworks_flexsdk_9-26.zip](https://github.com/fdorg/flashdevelop/files/1158631/frameworks_flexsdk_9-26.zip) which should be deployed to http://www.flashdevelop.org/appman/frameworks_flexsdk_9-26.zip

Note 2: For Flex SDK (OLD), should we use to `http://www.flashdevelop.org/appman/frameworks_flexsdk_9-26.zip`? Currently it uses `http://www.flashdevelop.org/appman/frameworks_flexsdk_9-21.zip`. I suppose people who still use the old Flex SDK may still want to target the latest player version.

Note 3: I didn't update TortoiseHg (latest is 4.2.2) because I'm not sure which build to use. Currently it's pointing to Mercurial-4.0.1-*.exe, however I'm not sure if that version includes TortoiseHg. Mercurial downloads page lists an msi installer as the one including TortoiseHg, but I can't check it at the moment.

Note 4: For some downloads e.g. Flash Player, it's rather pointless to mention a build number, since the download link stays the same even if Adobe releases a new build of the same version, the info in appman.xml would then be outdated.

### DIFF
--- a/appman.xml
+++ b/appman.xml
@@ -32,8 +32,8 @@
 	<Entry>
 		<Id>flexairsdk</Id>
 		<Name>Flex SDK + AIR SDK</Name>
-		<Version>4.6.0+25.0.0</Version>
-		<Build>23201B+134</Build>
+		<Version>4.6.0+26.0.0</Version>
+		<Build>23201B+118</Build>
 		<Desc>Adobe Flex SDK merged with Adobe AIR SDK</Desc>
 		<Info>http://www.adobe.com/devnet/air.html</Info>
 		<Group>Compilers</Group>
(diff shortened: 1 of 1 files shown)


Meychi:
Frameworks zip deployed, @Neverbirth could you finish this? After merge, remember to deploy from AppVeyor.

gene-pavlovsky:
:* Mika

gene-pavlovsky:
Heci :*

PR #1693, opened 2017-08-27
[ASGenerator][Haxe] Fixed of wrong auto import after override of function. fixes #1553

### LINKED ISSUE #1553: Wrong auto import after override of function
```haxe
class Foo {
    public function new() {}   
    function foo(c:Type.ValueType->String){}
}

class Bar extends Foo {
    override foo|
}
```
expected result:
```haxe
import Type.ValueType;

class Bar extends Foo {
    override function foo(c:Type.ValueType->String) {
        super.foo(c);
    }
}
```
actual result:
```haxe
import Type.ValueType->String;

class Bar extends Foo {
    override function foo(c:Type.ValueType->String) {
        super.foo(c);
    }
}
```

SlavaRa:


Meychi:
@Neverbirth, ok to merge?

External/Plugins/ASCompletion/Context/IASContext.cs
```diff
@@ -207,7 +206,14 @@ public interface IASContext
...
         bool UnsetOutOfDate();
 
+        /// <summary>
+        /// Returns the default value for the given type
+        /// </summary>
+        /// <param name="type"></param>
+        /// <returns>Default value</returns>
         string GetDefaultValue(string type);
+
+        List<string> DecomposeTypes(List<string> types);
```
Neverbirth: I would say that using `IEnumerable<string>`, at least as the returning type (but probably the argument too) is preferrable. I don't see you have any specific need for `List`.

SlavaRa: Fixed

PR #1707, opened 2017-09-06
Implementors / Overriders

chipshort:
Adds arrow markers next to functions that are overridden in a subclass, override a function in the super-class, implement a function of an interface or (for interfaces) are implemented by some class.

screenshot


chipshort:
Tests seem to also fail on development.

External/Plugins/ASCompletion/Helpers/ASTCache.cs
```diff
@@ -0,0 +1,264 @@
...
+        {
+            var set = GetOrCreateSet(dict, key);
+
+            set.UnionWith(value);
+        }
+
+        internal static ISet<T> GetOrCreateSet<S, T>(Dictionary<S, HashSet<T>> dict, S key)
+        {
+            HashSet<T> set;
+            dict.TryGetValue(key, out set);
```
SlavaRa: mb?
```csharp
HashSet<T> set;
if (!dict.TryGetValue(key, out set))
{
    set = new HashSet<T>();
    dict.Add(key, set);
}
return set;
```

External/Plugins/ASCompletion/Controls/ReferenceList.cs
```diff
@@ -0,0 +1,164 @@
...
+                    Tag = r,
+                    Group = group
+                };
+                listView.Items.Add(item);
+            }
+            
+        }
+    }
+
+    class Reference
```
SlavaRa: Is not it better to use the structure?

External/Plugins/ASCompletion/Helpers/ASTCache.cs
```diff
@@ -0,0 +1,264 @@
...
+                    PluginBase.RunAsync(new MethodInvoker(finished));
+                    return;
+                }
+
+                foreach (MemberModel cls in context.GetAllProjectClasses())
+                {
+                    if (PluginBase.MainForm.ClosingEntirely)
+                        return; //make sure we leave if the form is closing, so we do not block it
+
+                    var clas = GetClassModel(cls);
```
SlavaRa: line 18: cls
line 46: clas
Maybe it's worth using one name everywhere?

External/Plugins/ASCompletion/Helpers/ASTCache.cs
```diff
@@ -0,0 +1,264 @@
...
+
+                currentParent = currentParent.Extends;
+            }
+
+            return parentFunctions;
+        }
+
+        static CachedClassModel GetOrCreate(Dictionary<ClassModel, CachedClassModel> cache, ClassModel cls)
+        {
+            CachedClassModel cached;
```
SlavaRa: mb?
```csharp
HashSet<T> cached;
if (!cache.TryGetValue(key, out cached))
{
    cached = new CachedClassMode();
    cache.Add(key, cached);
}
return cached;
```

External/Plugins/ASCompletion/Controls/ReferenceList.cs
chipshort: I thought about that before.
I went with a class since the structs would become relatively big. Also they will be boxed anyway when they are assigned to the Tag of the item, so performance-wise there is probably not much difference.

External/Plugins/ASCompletion/Helpers/ASTCache.cs
chipshort: Fixed

Neverbirth:
Two things:
- In a file with multiple types it only displays the info for the last type.
- It should be switchable.

Neverbirth:
Conflicts

PR #1711, opened 2017-09-12
DiagnosticsLinter improvements

chipshort:
Worked on the diagnostics linter and LintingManager.

* Prevents CompletionList from closing when linting results appear
* Clears LinterPanel when switching projects and when compiling
* Lint files after changing project
* Changed linter api to have a seperate function to lint the whole project. (technically a breaking change, but I think nobody else is using it yet), also implemented global diagnostics and use it for this.
* Added a process queue for DiagnosticsLinter to prevent it from starting too many processes at once.


External/Plugins/CodeRefactor/BatchProcessors/ConsistentEOLProcessor.cs
```diff
@@ -14,13 +14,18 @@ class ConsistentEOLProcessor : IBatchProcessor
...
             foreach (var file in files)
             {
                 var document = PluginBase.MainForm.OpenEditableDocument(file) as ITabbedDocument;
                 document.SciControl.ConvertEOLs(document.SciControl.EOLMode);
             }
         }
+
+        public void ProcessProject(IProject project)
+        {
+
```
Neverbirth: Aren't you missing here the `BatchProcessManager.GetAllProjectFiles` code?

External/Plugins/HaXeContext/Linters/DiagnosticsLinter.cs
```diff
@@ -1,121 +1,216 @@
...
+                if (running.Count < maxRunning)
+                {
+                    running.Add(CreateTask(action));
+                }
+                else
+                {
+                    queue.Add(action);
+                }
+            }
+        }
```
Neverbirth: Just some API design comment. Personally I think that if you would return the Task you can make this a bit more elegant and nicer. But no need to change anything.

Issue #783, opened 2015-06-23
[Haxe] Organize imports shouldn't touch #elseif blocks


Gama11:
Organize imports ignores imports in blocks like these (as it should, too complex for FD to handle properly):

```haxe
import haxe.unit.TestCase;

#if flash
import flash.display.Sprite;
import flixel.FlxGame;
#end
```

However, this:

```haxe
import haxe.unit.TestCase;

#if flash
import flash.display.Sprite;
import flixel.FlxGame;
#elseif windows
import cpp.abi.Abi;
#end
```

Is turned into:

```haxe
#if flash
import flash.display.Sprite;
import flixel.FlxGame;
#elseif windows
#end
```

This is tested with none of the imported types being used in the file.

Neverbirth:
Fixed locally, I hope to be able to write some unit tests and make a PR during this week.

PR #1716, opened 2017-09-26
Detect lines inside #elseif directives (fixes issue #783).

Neverbirth:


Neverbirth:
I've had this on my local repo for more than 8 months and never got to commit it here...

However after writing the unit tests I found something funny... https://github.com/fdorg/flashdevelop/issues/1717

PR #1737, opened 2017-10-31
[AppMan] Update Haxe to 3.4.4

SlavaRa:


CI/tests.ps1
```diff
@@ -14,7 +14,7 @@ Param (
 
 If ($env:HAXEPATH -eq $null)
 {
-	cinst.exe haxe --version 3.4.3 -y
+	cinst.exe haxe --version 3.4.4 -y
```
Neverbirth: This cannot be merged, there is some problem in the Chocolatey package, I already reported to the Haxe guys, but I didn't get to look into why, maybe it's not setting some environment variable properly.

PR #1776, opened 2017-11-14
[Haxe][CodeRefactor] Fixed "Organize Imports" removes implicit imports of types within modules. fixes #191

### LINKED ISSUE #191: [haxe] "Organize Imports" removes implicit imports of types within modules
Organize imports does not seem to account for the fact that importing a module will also import also the types within.

Example with two files:

Test.hx

``` haxe
package;

class Test
{

}

typedef TestTypedef = Void->Void;
```

test/Import.hx

``` haxe
package test;

import Test.TestTypedef;

class Import
{
    public function new(test:TestTypedef) 
    {

    }
}
```

---

The explicit import of `TestTypedef` is not removed when organizing imports - when you change it to only `import Test;` however, it is removed, even though it is required for compilation.

Interestingly enough, another import has to be added to `Import.hx` - otherwise the Organize Imports option is greyed out.

SlavaRa:


Meychi:
Could you get around OrganizeImports : global::CodeRefactor.Commands.OrganizeImports decl?

SlavaRa:
?

Meychi:
Some how remove it, @Neverbirth said its possible.

SlavaRa:
```
using CodeRefactor.Commands.OrganizeImports;
HaxeOrganizeImports : OrganizeImports 
```
? 

What's wrong with the `global::CodeRefactor.Commands.OrganizeImports`?

Meychi:
Not consistent i guess.

SlavaRa:
I can do HaxeOrganizeImports if this will speed up the merge.

Meychi:
Yep.

SlavaRa:
Fixed

SlavaRa:
@Meychi now ok?

PR #1821, opened 2017-12-01
[ASFileParser][Haxe] Fixed parsing of regex in ternary operator. fixes #1814

### LINKED ISSUE #1814: Haxe completion breaks after a regex
Valid Haxe code:

```haxe
package;

class Test {
	public function new() {}

	function parseClassBody(text:String):String {
		var rx = ~/\bclass\b[^{]*{\s*\n/;
		return rx.match(text) ? ~/\s*}\s*$/.replace(rx.matchedRight(), '\n') : 'Error\n';
	}
}
```

Place the caret after the end of the function, and type `fu`.
Expected to see `function`, instead the completion list has `Float` etc.
The culprit that confuses the completion is the `}` in the second regexp (`~/\s*}\s*$/`). Looks like it is treated as `}` closing the function block.
Tried Compiler and FlashDevelop completion modes, same result.

SlavaRa:


SlavaRa:
done :)

PR #1822, opened 2017-12-03
[ProjectManager][HaxeProject] Fixed compile error when output path of project ends with '\' or '/'. fixes #1469

### LINKED ISSUE #1469: Cannot compile C# and C++ targets in Windows 7 x64
I've got the following error (similar for hxcpp and hxcs) for default projects in C++/C#.
OS Windows 7 x64.

Running process: C:\Program Files (x86)\HaxeDevelop\Tools\fdbuild\fdbuild.exe "E:\temp\New Project14\New Project14.hxproj" -ipc 7de8e6ea-dcd6-4496-acea-c62337a09d79 -version "3.4.0" -compiler "C:\HaxeToolkit\haxe" -notrace -library "C:\Program Files (x86)\HaxeDevelop\Library"
Building New Project14
haxe  -cp src -cpp "E:/temp/New Project14/bin" -main Main
Unix.Unix_error(21, "open_proc", "haxelib path hxcpp")
Build halted with errors (haxe.exe).
Done(1)

hxcpp and hxcs are installed. System environment variables are set.
HAXELIB_PATH=C:\HaxeToolkit\haxe\lib
HAXEPATH=C:\HaxeToolkit\haxe
HAXE_LIBRARY=C:\HaxeToolkit\haxe\std
%HAXEPATH% is appended to PATH variable value.

Compiling simple C# project without IDE (from cmd line) is ok: 
haxe build.hxml

where
build.hxml contents:
-cs test
-main Test
-D haxe3

SlavaRa:


Neverbirth:
Checking this one.

Neverbirth:
Sadly this is not 100% the proper fix. `OutputPathAbsolute` is used in several places, some of them doing `Path.GetDirectoryName()` which returns "C:\parentPath" in the case of "C:\parentPath\path".

I guess you already noticed this being somewhat wrong because you put that `Directory.Exists()` check, and in this case FlashDevelop should create it.

The thing is that if the directory doesn't exist this fix doesn't work and Haxe will still fail. 

Looking at the code I would say this has to be fixed in `HaxeProjectBuilder.DoBuild()`.

SlavaRa:
I put this check, because there may be a file

SlavaRa:
Thanks, fixed

Neverbirth:
I'm curious, have you tested this with Haxe 2+? I'm wondering if there is any Haxe version that requires the ending slash.

SlavaRa:
Haxe 2? No ;)
>I'm wondering if there is any Haxe version that requires the ending slash.
I do not know, but I know that because of her there are problems:
https://github.com/HaxeFoundation/haxe/issues/6768
https://github.com/HaxeFoundation/haxe/issues/6212

PR #1824, opened 2017-12-04
[WIP] Feature/1187 improve open resource dialog

### LINKED ISSUE #1187: "Open resource" dialog - improve search feature for files like a_b or a-b
Open resource dialog has really nice ability to quickly search for filenames using MixedCase.
It would be awesome if it would also start supporting camelCase, underscores and dashes filename styles.

Examples:
- `ACET` offers `AppControllerEventType.as` - it works and it's awesome!
- `gST` should offer `getStackTrace.as` - now it doesn't
- `s_m` should offer `solarized_mobile.xml` - now it doesn't
- `s-m` should offer `solarized-mobile.json` - now it doesn't
- `b.x` should offer `build.xml` - now it doesn't

Would that be possible to implement?

gene-pavlovsky:
Closes #1187 
Improve Open Resource dialog fuzzy matching.
According to my experience, significantly improves user experience (in my arrogant opinion ;) ).


External/Plugins/ProjectManager/Controls/OpenResourceForm.cs
```diff
@@ -492,45 +497,72 @@ public static List<string> getMatchedItems(List<string> source, string searchTex
             try
             {
                 searchFile = Path.GetFileName(searchText);
-                searchDir = Path.GetDirectoryName(searchText);
+                searchDir = Path.GetDirectoryName(searchText);
```
gene-pavlovsky: `Path.GetDirectoryName(searchText)` automatically converts `/` to `\\` (on Windows), so searchDir has `\\`, meanwhile `searchText` had `/`

External/Plugins/ProjectManager/Controls/OpenResourceForm.cs
```diff
@@ -542,73 +574,85 @@ public static List<string> getMatchedItems(List<string> source, string searchTex
...
-                while (j < text.Length && char.IsLower(text[j]))
-                {
-                    j++;
-                }
-                if (j == text.Length) return false;
-                if (pattern[i] != text[j]) return false;
-            }
-            return i == pattern.Length;
+            // Prefer shorter results. The effect is less pronounced when query length is increased.
+            return Score(str, query, caseMatters) / (str.Length + query.Length);
```
gene-pavlovsky, line 585: Previously score was divided by an arithmetic average of `str.Length` and `query.Length`. Just using their `sum` is as good, though, since scores are only compared to each other, any constant factor (`1/2`) doesn't matter.

gene-pavlovsky:
I'm thinking about if we really should favor short results.
Because currently score is divided by length of the result (more specifically, average of result length and query length),  in real-world usage I'm sometimes seeing not the best matches being on top just because they are shorter.

chipshort:
@gene-pavlovsky I am not sure how not dividing by length would impact the results. I could imaging that this would favour long results too much, because the longer the result, the more likely it is to have the query characters in it (in the right order).
Did you try it? How does it turn out?

gene-pavlovsky:
I will try tomorrow and report back. But my thinking is if we have e.g. one short and one long result with query chars in the right order, the scoring algorithm will still pick the best overall match (e.g. whichever one has the query chars falling on beginning of word boundaries, or consecutive position, etc.). If the scores are equal, the shorter match will show up first because final results are sorted by score then by length (then alphabetically).

gene-pavlovsky:
So far I can't say if it makes matters better or worse.
There are a few strange results I'm getting that I have to investigate.

gene-pavlovsky:
I have a few cases producing unexpected results. I will have to investigate more...

PR #1838, opened 2017-12-08
Fix implements arrow getting messed up when settings are closed

### LINKED ISSUE #1815: Closing settings dialog leads to show implements arrows disappearing and messes up lines background color
Pre-requisite: `Show implements` setting enabled (default).
Open any class implementing an interface, or an interface implemented by some class.
Open Tools > Settings, and then just close it without any changes.
The implements icons and their column disappear (from all open tabs), the lines of code which had those icons lose their background color (becomes white, I assume it is only noticeable when using a non-default theme).
Making any change to any file and saving it brings the implements icons back and redraws the background as well (for all open tabs).

@chipshort

chipshort:
Fixes #1815


Neverbirth:
I have the feeling this is not the best solution. Hopefully @Meychi can provide better feedback.

PR #1842, opened 2017-12-10
[Haxe][ASFileParser] Fixed parsing of regex after a 'switch', 'try'. fixes #1841

### LINKED ISSUE #1841: [Haxe][ASFileParser][Bug] Haxe completion breaks after a switch, try
```haxe
class Test {
	switch ~/\s*}\s*$/.match("}") {
            case true: trace("1");
            case false: trace(0);
        }
    }
}
```
Description #1814

SlavaRa:


External/Plugins/ASCompletion/Model/ASFileParser.cs
```diff
@@ -1705,8 +1705,18 @@ private bool LookupRegex(string ba, ref int i)
...
-                if ("=(,[{;?:+-*/|&~^><n!".IndexOf(c) >= 0) break; // ok
-                if (" \t".IndexOf(c) >= 0) continue;
+                if (c == '=' || c == '(' || c == '[' || c == '{' || c == ',' || c == ';'
+                    || c == '?' || c == ':'
+                    || c == '+' || c == '-' || c == '*' || c == '/'
+                    || c == '|' || c == '&' || c == '~' || c == '^'
+                    || c == '>' || c == '<' || c == '!'
+                    || c == 'n' /* return ~/regex/ or in ~/regex/ */
+                    || c == 'h' /* switch ~/regex/ */
+                    || c == 'y' /* try ~/regex/ */)
```
Neverbirth: I think this is way less readable than previous code, but it's not so big deal. The problem I see now is that this is not quite bulletproof (plus that single character check makes me feel a bit uneasy heh). For example, what if:

`function test():Void ~/regex/;`

In the same way it would be possible to do for example:

`function test(value:String):AnythingPossible ~/regex/.match(value) ? return whatever : return null;`

SlavaRa:
this is another more complicated problem, if I start solving them all at once, it will take a very long time and there will be a lot of change.

I know more than 10 of these options.

External/Plugins/ASCompletion/Model/ASFileParser.cs
SlavaRa: Yes, such cases are not currently supported.
The parsing of single-line functions I will be make separately.

SlavaRa:
Done.
The parsing of single-line functions I will be make separately.

Neverbirth:
This seems a bit hacky and quite probably there are better ways of solving this problem, but it could be investigated later.

PR #1844, opened 2017-12-11
 [Haxe] Support for 'import.hx'. fixes #1426

### LINKED ISSUE #1426: import.hx items are not highlighted nor shown in auto-completion
Static imports via `import.hx` are not highlighted the same way as static imports inside the classes are.

In addition, they are not shown in auto-completion menu, though the argument list is still displayed upon opening the bracket.
image
Sample: https://gist.github.com/YellowAfterlife/6f3410ead21f4c015e099e6d0134d046

SlavaRa:


External/Plugins/HaXeContext/Context.cs
```diff
@@ -819,7 +800,7 @@ public override MemberList ResolveImports(FileModel inFile)
                 foreach (ClassModel aClass in inFile.Classes)
                     if (aClass.Access != Visibility.Private) ResolveImport(aClass, imports);
             }
-
+            imports.Add(ResolveDefaults(inFile.Package));
```
Gama11: `import.hx` is "recursive".. e.g. if you have a file structure like this:

```
- src
    - import.hx
    - foo
        - import.hx
        - Bar.hx 
```

Then `Bar.hx` will have the imports of both `import.hx` files available. It doesn't look like this is covered by that? Also, it's not package, but directory-based (a file with the same package, but in another directory as an `import.hx`, will _not_ be affected by it).

SlavaRa:
@Gama11 all ok?

SlavaRa:
@Neverbirth ?

Yanrishatum:
@Neverbirth Bump. As I extensively use import.hx in my project since this feature was introduced, lack of proper support from FD gets very annoying.

Yanrishatum:
If that would speed things up: I did tested this PR, and it does improve my life considerably, It works as expected both with project-level `import.hx` and package-level `import.hx`.

External/Plugins/HaXeContext/Context.cs
```diff
@@ -1113,11 +1088,39 @@ protected override void InitTopLevelElements()
...
+            if (GetCurrentSDKVersion() < "3.3.0") return result;
+            var packagePath = string.IsNullOrEmpty(package) ? string.Empty : package.Replace('.', dirSeparatorChar);
+            while(true)
+            {
+                foreach (var it in classPath)
+                {
+                    if (!it.IsValid || it.Updating) continue;
+                    var path = Path.Combine(it.Path, packagePath, "import.hx");
+                    if (!File.Exists(path)) continue;
+                    var model = ASFileParser.ParseFile(new FileModel(path) {Context = this});
```
elsassph: Why parsing it? shouldn't it be already parsed by the classpath  exploration? If not, maybe an exception for this file could be made.

elsassph:
A bit wasteful as it will re-parse the file continuously instead of adding it to the model.

External/Plugins/HaXeContext/Context.cs
SlavaRa: Did you read https://haxe.org/blog/importhx-intro/ ?

SlavaRa: After parsing, these files will be cached.

SlavaRa: Or you mean that parsing needs to be moved to BuildClassPath?

elsassph: I don't think it gets cached - and yes it could be handled during classpath exploration and kept up to date like other classes.

SlavaRa: ok, I will try to do otherwise

SlavaRa:
> A bit wasteful as it will re-parse the file continuously instead of adding it to the model.

@elsassph fixed

PR #1848, opened 2017-12-11
[ASComplete] Show documentation of constructor when hovering on 'Some' for 'new Some()'. fixes #637

### LINKED ISSUE #637: [haxe] documentation of new Foo()
if I type : 

```
new h2d.Button();
```

The hovered completion hint fot the class h2d.Button and does not hint for the constructor documentation.

It would be better to mix both.

SlavaRa:


SlavaRa:
@Neverbirth ?

Neverbirth:
Note that this doesn't work properly when a constructor is not defined because of a parent class already defines one... but I see that case doesn't work properly in normal completion either.

SlavaRa:
Yes, this will be fixed soon

PR #1856, opened 2017-12-15
[ASGenerator] Suggest generate a constructor instead of a modification of declaration if the constructor is not declared in the class or in the parents. fixes #1738

### LINKED ISSUE #1738: [Haxe][ASGenerator][Feature] Offer to generate a constructor instead of a modification of declaration if the constructor is not declared in the class or in the parents
```haxe
package;

class Main {
	public function new() {
		new |Foo(); // | <-- cursor
	}
}

class Foo {
}
```
actual result:

SlavaRa:


Tests/External/Plugins/ASCompletion.Tests/Completion/ASGeneratorTests.cs
```diff
@@ -3334,6 +3335,66 @@ internal static void Common(string sourceText, string constructorArgs, Scintilla
...
+                public string Haxe(string fileName, GeneratorJobType job, bool hasGenerator) => HaxeImpl(sci, fileName, job, hasGenerator);
+
+                internal static string HaxeImpl(ScintillaControl sci, string fileName, GeneratorJobType job, bool hasGenerator)
+                {
+                    SetHaxeFeatures(sci);
+                    var sourceText = ReadAllTextHaxe(fileName);
+                    fileName = GetFullPathHaxe(fileName);
+                    fileName = Path.GetFileNameWithoutExtension(fileName).Replace('.', Path.DirectorySeparatorChar) + Path.GetExtension(fileName);
+                    fileName = Path.GetFullPath(fileName);
+                    fileName = fileName.Replace($"\\FlashDevelop\\Bin\\Debug\\{nameof(ASCompletion)}\\Test_Files\\", $"\\Tests\\External\\Plugins\\{nameof(ASCompletion)}.Tests\\Test Files\\");
```
SlavaRa: I have no idea how to do it better, but it will not work otherwise :)

PR #1858, opened 2017-12-18
Thread-safe EventManager.

Neverbirth:
This fixes some of the crashes people may be having.

Performance in single-threaded scenarios will be most times the same and at worst just some nanoseconds slower.

### DIFF
--- a/PluginCore/PluginCore/Managers/EventManager.cs
+++ b/PluginCore/PluginCore/Managers/EventManager.cs
@@ -15,6 +15,8 @@ namespace PluginCore.Managers
         private static EventObject[] eventObjectsSnapshot;
         private static bool snapshotInvalid;
 
+        private static readonly object eventLock = new object();
+
         static EventManager()
         {
             highObjects = new List<EventObject>();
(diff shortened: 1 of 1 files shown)


PluginCore/PluginCore/Managers/EventManager.cs
```diff
@@ -95,26 +106,37 @@ public static void RemoveEventHandler(IEventHandler handler, EventType mask, Han
...
                 }
-                highObjects.CopyTo(eventObjectsSnapshot, 0);
-                normalObjects.CopyTo(eventObjectsSnapshot, highObjects.Count);
-                lowObjects.CopyTo(eventObjectsSnapshot, highObjects.Count + normalObjects.Count);
-                snapshotInvalid = false;
             }
             else
             {
-                length = eventObjectsSnapshot.Length;
+                eventObjectsCopy = eventObjectsSnapshot;
```
Neverbirth: Interlocked could be used... but the CLR spec enforces this to be atomic, so it shouldn't be  a problem...

Neverbirth:
I forgot to say: as FD spends way more time reading than writing, I went with mostly a lock-free approach for this, and due to the way how the system was designed even before my changes, it would be possible that an event handler receives events after removing himself from listening. There is no code on our side right now that would suffer from this but it's something to take into account when writing code (like checking `IsDispoing || IsDisposed` in the `HandleEvent` function), and solving it would require quite drastic changes (that maybe wouldn't fully solve the issue either or would bring other problems/complex design decisions) or making everything completely locked, which could also be quite bad for performance on other threads.

PR #1859, opened 2017-12-19
[UnitTests][ASGeneratorTests] Remove duplicated code

SlavaRa:


SlavaRa:
@Neverbirth ?

Neverbirth:
I don't know about others' opinion on the subject. Personally I don't like that the common function used for some of the cases does more than needed, hiding possibly some interesting information for future use.

If @Meychi is ok with it it can be merged.

SlavaRa:
This code is simply duplicated everywhere and when you need to add something you have to add it everywhere, everywhere the same thing

elsassph:
Less boilerplate sound good.

Neverbirth:
OK.

PR #1863, opened 2017-12-23
Upgrade to NUnit 3.9.

Neverbirth:
This PR is just a preview, not meant to be merged already (I don't want @SlavaRa to kill me heh). The plan is to update all the unit tests to NUnit 3 as any new development happens for it.

Updating to NUnit 3 would bring us the possibility to parallelize the test execution too.

I checked that the tests run properly under Resharper, the NUnit console runner, and the NUnit Test Adapter (there are 2 tests that fail here: "Issue 104. typdef" and "Issue1747. Case 2", I didn't check why yet). I didn't check on the CI server, it's quite probably it won't run because of the mix of NUnit 2 and Nunit 3 tests existing right now, and because the NUnit console runner exe name is different.

While doing the upgrade I noticed several things unrelated to this update, I will put inline comments.

### DIFF
--- a/PluginCore/PluginCore/Helpers/PathHelper.cs
+++ b/PluginCore/PluginCore/Helpers/PathHelper.cs
@@ -19,10 +19,23 @@ namespace PluginCore.Helpers
         /// </summary>
         public static String BaseDir => PluginBase.MainForm.StandaloneMode ? AppDir : UserAppDir;
 
+        private static string appDir;
+
         /// <summary>
         /// Path to the main application directory
         /// </summary>
-        public static String AppDir => Path.GetDirectoryName(GetAssemblyPath(Assembly.GetExecutingAssembly()));
+        public static String AppDir
+        {
+            get
+            {
+                if (appDir == null)
+                {
+                    appDir = Path.GetDirectoryName(GetAssemblyPath(Assembly.GetExecutingAssembly()));
+                }
+
+                return appDir;
+            }
+        }
 
         /// <summary>
         /// Path to the user's application directory
--- a/PluginCore/ScintillaNet/ScintillaControl.cs
+++ b/PluginCore/ScintillaNet/ScintillaControl.cs
@@ -236,7 +236,7 @@ namespace ScintillaNet
 
         #region Scintilla Main
 
-        public ScintillaControl() : this(IntPtr.Size == 4 ? "SciLexer.dll" : "SciLexer64.dll")
+        public ScintillaControl() : this(Path.Combine(PathHelper.AppDir, IntPtr.Size == 4 ? "SciLexer.dll" : "SciLexer64.dll"))
         {
             if (Win32.ShouldUseWin32()) DragAcceptFiles(this.Handle, 1);
         }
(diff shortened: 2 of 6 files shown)


PluginCore/PluginCore/Helpers/PathHelper.cs
```diff
@@ -19,10 +19,23 @@ public class PathHelper
...
+            get
+            {
+                if (appDir == null)
+                {
+                    appDir = Path.GetDirectoryName(GetAssemblyPath(Assembly.GetExecutingAssembly()));
+                }
+
+                return appDir;
+            }
+        }
```
Neverbirth: Is it possible that the executing assembly may change at runtime or have several for the same app domain? I'm not sure, but even if it would, I think it makes sense to cache here.

PluginCore/ScintillaNet/ScintillaControl.cs
```diff
@@ -236,7 +236,7 @@ private void RemoveScrollBars(ScintillaControl sender)
 
         #region Scintilla Main
 
-        public ScintillaControl() : this(IntPtr.Size == 4 ? "SciLexer.dll" : "SciLexer64.dll")
+        public ScintillaControl() : this(Path.Combine(PathHelper.AppDir, IntPtr.Size == 4 ? "SciLexer.dll" : "SciLexer64.dll"))
```
Neverbirth: Before we were loading Scintilla based on relative path, which depends on the current running path... this can lead to multiple problems... I'm surprised we didn't run into this before? While doing the migration to NUnit 3 I stumbled into this problem... but when going backwards to NUnit 2 I was still facing this problem.

This can also make FD to fail depending on how it is run.

Tests/External/Plugins/ASCompletion.Tests/Completion/ASGeneratorTests.cs
```diff
@@ -3373,7 +3373,7 @@ internal static string HaxeImpl(ScintillaControl sci, string fileName, Generator
                     var sourceText = ReadAllTextHaxe(fileName);
                     fileName = GetFullPathHaxe(fileName);
                     fileName = Path.GetFileNameWithoutExtension(fileName).Replace('.', Path.DirectorySeparatorChar) + Path.GetExtension(fileName);
-                    fileName = Path.GetFullPath(fileName);
+                    fileName = Path.Combine(PathHelper.AppDir, fileName);
```
Neverbirth: Same problem as with Scintilla. While upgrading to NUnit 3 my startup directory was not the same as FD so this logic was not running properly.

SlavaRa:
I will check it tonight :)

SlavaRa:
From https://www.appveyor.com/docs/running-tests/#nunit-3x
```
NUnit 2.x
To run NUnit tests with real-time reporting use command:

nunit-console <assembly> [options]
or for x86 assemblies:

nunit-console-x86 <assembly> [options]
NUnit 3.x
To run NUnit tests with reporting test results to AppVeyor use command:

nunit3-console <assembly> [options] --result=myresults.xml;format=AppVeyor
```

Neverbirth:
I know, but no point in modifying that part until every test is converted to NUnit 3, and no worth the time to support both NUnit 2 and NUnit 3 tests (by the way, the VS Test Adapter doesn't work properly if mixing 2 and 3 tests).

PR #1875, opened 2017-12-31
Fixed `ASGeneterator.GetEndOfStatement` for statement with string. fixes #1874

### LINKED ISSUE #1874: [Bug][ASGenerator][GetEndOfStatement] Wrong result for expression with string

SlavaRa:


SlavaRa:
Oh

External/Plugins/ASCompletion/Completion/ASGenerator.cs
```diff
@@ -2035,9 +2037,21 @@ public static int GetEndOfStatement(int startPos, int endPos, ScintillaControl s
...
+                {
+                    if (statementEnd <= 1 || (char)sci.CharAt(statementEnd - 2) == '\\') continue;
+                    if (dQuotes == 0) dQuotes++;
+                    else dQuotes--;
+                }
+                else if (c == '\'' && dQuotes == 0)
+                {
+                    if (statementEnd <= 1 || (char)sci.CharAt(statementEnd - 2) == '\\') continue;
+                    if (sQuotes == 0) sQuotes++;
+                    else sQuotes--;
```
Neverbirth: This function already uses a Scintilla instance to know about comments why not doing the same for strings? I see a problem with it, also existing in the version without it... multiline strings. It's not handled properly here, and the bundled Scintilla doesn't support them either, but then it's delegating it to when a more proper Scintilla is added. Also, if `startPos` would happen to be in the middle of a string this wouldn't work properly.

SlavaRa: Thanks. I have to check this out.

SlavaRa:
fixed

External/Plugins/ASCompletion/Completion/ASGenerator.cs
```diff
@@ -2026,18 +2026,19 @@ public static int GetStartOfStatement(ScintillaControl sci, int statementEnd, AS
         public static int GetEndOfStatement(int startPos, int endPos, ScintillaControl sci)
         {
             int groupCount = 0;
-            int brCount = 0;
+            int brCount = 0;
+            var dQuotes = 0;
+            var sQuotes = 0;
```
Neverbirth: These variables are now unneeded.

SlavaRa: Sorry

SlavaRa: fixed

PR #1882, opened 2018-01-02
[UI] Add MemoryMessageBox

SlavaRa:
@Meychi please review this PR


PluginCore/PluginCore/Controls/MemoryMessageBox.cs
```diff
@@ -0,0 +1,124 @@
...
+        {
+            label.Text = text;
+            Text = caption;
+            return ShowDialog();
+        }
+
+        public new MemoryMessageBoxResult ShowDialog()
+        {
+            Result = MemoryMessageBoxResult.Cancel;
+            if (dontAsk.Checked)
```
elsassph: I think this is dangerous - "don't ask" should only apply to a specific operation, not any operation using this dialog.

elsassph: Good otherwise.

SlavaRa: false by default. But I think I can add the ability to hide this control.

elsassph: Ah sorry I thought it was shared across all the instances -  so that will work only if you keep the same dialog instance?

SlavaRa: Yes

SlavaRa:
@meychi, what do you think about this?

Meychi:
Look good. I think the text should be 'No to all' and not 'Not to all'. Does the message area shrink and enlarge with the text as does the MessageBox do?

Meychi:
And should we merge this after release?

PR #1884, opened 2018-01-02
Fixed ASGenerator.GetStartOfStatement for array comprehension. fixes #1883

### LINKED ISSUE #1883: [Bug][ASgenerator][GetStartOfStatement] Wrong result for array comprehension
```haxe
 [1 => 1, 2 => 2]|
```
`|` - cursor
expected result: 1
actual result: 0

SlavaRa:


External/Plugins/ASCompletion/Completion/ASGenerator.cs
```diff
@@ -1982,11 +1982,11 @@ public static int GetStartOfStatement(ScintillaControl sci, int statementEnd, AS
                 else if (c == '[' && arrCount > 0) arrCount--;
                 else if (c == ')') parCount++;
                 else if (c == '(' && parCount > 0) parCount--;
-                else if (c == '>')
+                else if (c == '>' && arrCount == 0 && parCount == 0 && braCount == 0)
```
Neverbirth: I'm really curious if this is 100% fine, as there are many cases were generics may appear between []{}(), but I cannot check at the moment, also it seems there are no tests with this?. For example `arr = [new Array<String>()]`, `function test(type:Class<Dynamic>)`, etc.

SlavaRa: 5 min

SlavaRa: All for you ^__^

SlavaRa: generics is not a problem, problem is operators `<`, `>`

SlavaRa: >For example arr = [new Array<String>()], function test(type:Class<Dynamic>), etc.

2918: `new TestCaseData(" new Vector.<int>()$(EntryPoint)"`
2927: `new TestCaseData(" new <Vector.<Object>>[new <Object>[{}]]$(EntryPoint)"`
2970: `new TestCaseData(" new Array<{name:String, params:Array<Dynamic>}>()$(EntryPoint)"`

Neverbirth: I see, but in Haxe "=>" is an operator and could appear in other places to be later processed at macro time, so "=>" should be processed specifically and not assume it's going to be between brackets.

SlavaRa: in the next PRs there will be a lot of them

PR #1885, opened 2018-01-02
[ASGenerator] 'Assign statement to var' for increment and decrement operators

SlavaRa:


External/Plugins/ASCompletion/Completion/ASGenerator.cs
```diff
@@ -1290,18 +1291,30 @@ static void AssignStatementToVar(ScintillaControl sci, ClassModel inClass, State
...
-                if (arithmeticOperators.Contains(resolve.Context.Separator))
+            var context = resolve.Context;
+            if (context != null)
+            {
+                var operators = ctx.Features.ArithmeticOperators
+                    .Select(it => it.ToString())
+                    .Concat(ctx.Features.IncrementDecrementOperators)
+                    .ToHashSet();
+                var sep = new[] {' '};
+                var isValid = new Func<ASExpr, bool>((c) => c.Separator.Contains(' ')
```
elsassph: Too unspecific name.

SlavaRa:
IsValid ?

elsassph:
Yes, also I fail to see what cases this `isValid` covers - is that viible in the tests?

SlavaRa:
Local check function, have suggestions how to call it better?

External/Plugins/AS2Context/Context.cs
```diff
@@ -123,7 +123,8 @@ public Context(AS2Settings initSettings)
             features.intrinsicKey = "intrinsic";
 
             features.functionArguments = new MemberModel("arguments", "FunctionArguments", FlagType.Variable | FlagType.LocalVar, 0);
-
+            features.ArithmeticOperators = new HashSet<char> { '+', '-', '*', '/' };
+            features.IncrementDecrementOperators = new[] {"++", "--"};
```
Neverbirth: Why `IncrementDecrementOperators`? I never saw such a classification for this, they are normally unary operators  (in the new Haxe documentation they appear together with other arithmetic operators, but in the old one they also appear as unary operators).

SlavaRa: https://docs.microsoft.com/en-us/scripting/javascript/reference/increment-and-decrement-operators-javascript
http://en.cppreference.com/w/cpp/language/operator_incdec
...

Neverbirth: Actually Microsoft is listing them together with the other arithmetic operators:

https://docs.microsoft.com/en-us/scripting/javascript/operators-javascript

w3schools and Mozilla dev space also lists them together with arithmetic operators:

https://www.w3schools.com/jsref/jsref_operators.asp
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Arithmetic_Operators

So I guess in JS they are qualified as arithmetic operators mostly.

Other links:

https://www.w3schools.com/jsref/jsref_operators.asp
https://msdn.microsoft.com/en-us/library/hetcw0tx.aspx
https://www.tutorialspoint.com/cplusplus/unary_operators_overloading.htm
https://en.wikipedia.org/wiki/Unary_operation
https://help.adobe.com/en_US/ActionScript/3.0_ProgrammingAS3/WS5b3ccc516d4fbf351e63e3d118a9b90204-7fd1.html

SlavaRa: I can not mix everything.
Otherwise, there will be a lot of confusion in the development and the amount of code in pull requests will grow tens of times, and pull requests and so are not taken for months.
Also, separation is necessary for deriving types later.

gene-pavlovsky:
Personally I'd call them `UnaryArithmeticOperators`, but why do you guys nitpick about this? I think the current name is mostly fine - unless you think there would be other unary operators added in the future. Unlikely.

PR #1894, opened 2018-01-03
[ASComplete] Fixed show calltip if previous argument uses array comprehensions. fixes #764

### LINKED ISSUE #764: [Haxe] Calltip completion breaks if previous argument uses array comprehension
``` haxe
class Main {
    static function main() {
        test([for (i in 0...10) i]
    }

    static function test(a:Array<Int>, b:Int) {

    }
}
```

Expected behavior: the calltip completion pops up after typing the comma like here:

Actual: nothing happens.

SlavaRa:


Neverbirth:
I will try to check this one today.

External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -1782,66 +1782,38 @@ private static ClassModel ResolveParameterType(int paramIndex, bool indexTypeOnl
...
         {
             int parCount = 0;
             int braCount = 0;
             int comaCount = 0;
             int arrCount = 0;
-            var genCount = 0;
-            var dquCount = 0;
-            var squCount = 0;
+            var genCount = 0;
+            sci.Colourise(0, -1);
```
Neverbirth: My understanding is that calling `Colourise(0, -1)` is expensive, and some of the functions calling `FindParameterIndex` are already calling this before (although it's not easy to find some of them by just browsing the code). @elsassph @Meychi is my comment right or am I remembering wrong/something else? If so then I guess it would be better to identify which functions calling this are not calling `Colourise` before?

Other than this I think it's ok.

elsassph: We could add some logging of calls to `Colourise` to check if we're abusing it yes.

SlavaRa:
is it okay or should I close this request?

SlavaRa:
ok, if not necessary, I will close after release :)

Neverbirth:
I missed your message. The PR is ok, but that sci.Colourise call should be checked, on the cases I checked it is not necessary because it is already called before, so making an extra call is wasting CPU from our understanding because it makes scintilla to reparse and recolourose everything. But I didn't check for every case where `FindParameterIndex` is called.

SlavaRa:
how can I verify that it was called up before?

Neverbirth:
I'm not sure if there is a built-in way. Although if my understanding is correctly we can implement it in scintillacontrol.

I would say the best is to assume inside FindParameterIndex has always been called, and make sure that it is always the case.

@elsassph am I correct? 

We could ask in the scintilla group. We should update our scintilla...

SlavaRa:
Well, I'll be waiting for a solution, but in the future, I think I can release my version of the plug-ins so that some users get the functionality while you make the decision))

Neverbirth:
Merged! Thanks. Note I will make slight modifications regarding the Colourise call.

PR #1892, opened 2018-01-03
[Haxe][ASGenerator] `Assign statement to var` for array comprehension. fixes #1766

SlavaRa:

PR #1912, opened 2018-01-05
Added operator 'modulo' to AS3Context.Features.ArithmeticOperators, HaxeContext.Features.ArithmeticOperators

SlavaRa:


Tests/External/Plugins/ASCompletion.Tests/Test Files/generated/haxe/AfterAssignStatementToVar_issue1749_6.hx
```diff
@@ -0,0 +1,6 @@
+﻿package;
+class Main {
+	public function new() {
+		var float:Float = 1 % 5;
```
Neverbirth: I don't know about AS right now, but in Haxe this is actually an `Int`. A modulo of 2 ints is an int. If it would be for example `1 % 5.0` then it would be a `Float`.

SlavaRa: Yes, I will make a deduction `Int`, after merge #1892

PR #1915, opened 2018-01-06
[ASComplete] Added ASComplete.ExpressionEndPosition. fixes #1911

### LINKED ISSUE #1911: [Bug][Haxe] Incorrect parsing of members with matching names.
expected result: `static private offset() : Int in Main`

SlavaRa:
Please use:
```csharp
var position = ASComplete.ExpressionEndPosition(sci, sci.CurrentPos);
var result = ASComplete.GetExpressionType(sci, position, false, true);
```
instead of
```csharp
var position = sci.WordEndPosition(sci.CurrentPos, true);
var result = ASComplete.GetExpressionType(sci, position);
```
because the second option does not work correctly with functions, for example `test ()`


External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -4439,6 +4439,58 @@ private static bool IsXmlType(ClassModel model)
...
+                    }
+                    else if (c <= ' ')
+                    {
+                        hadWS = true;
+                        spacesCount++;
+                    }
+                    else break;
+                }
+            }
+            return statementEnd - 1 - spacesCount;
```
elsassph: Why substracting the spaces? Don't they still count in the character position of the expression?
Shouldn't you return the word-end position in case `groupCount` doesn't reach `0`?

External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -4439,6 +4439,58 @@ private static bool IsXmlType(ClassModel model)
...
+                }
+                else if (groupCount == 0)
+                {
+                    if (characterClass.Contains(c))
+                    {
+                        if (hadWS) break;
+                    }
+                    else if (c <= ' ')
+                    {
+                        hadWS = true;
```
elsassph: In which case is a space character usefully part of an expression?

External/Plugins/ASCompletion/PluginMain.cs
```diff
@@ -1209,10 +1209,9 @@ private void OnMouseHover(ScintillaControl sci, int position)
 
             // get word at mouse position
             int style = sci.BaseStyleAt(position);
-            if (!ASComplete.IsTextStyle(style))
-                return;
-            position = sci.WordEndPosition(position, true);
-            ASResult result = ASComplete.GetExpressionType(sci, position);
+            if (!ASComplete.IsTextStyle(style)) return;
+            position = ASComplete.ExpressionEndPosition(sci, position);
```
elsassph: From my understanding, when hovering a function call, this will provide the type of the return value and not the function member itself. That would be unexpected.

Tests/External/Plugins/ASCompletion.Tests/Completion/ASCompleteTests.cs
```diff
@@ -984,6 +983,104 @@ public string ParseFile_Issue104(string name, FlagType flags)
...
+                        .Returns("a".Length);
+                    yield return new TestCaseData("a ? b : c")
+                        .Returns("a".Length);
+                    yield return new TestCaseData("a:Boolean")
+                        .Returns("a".Length);
+                    yield return new TestCaseData("a = b")
+                        .Returns("a".Length);
+                    yield return new TestCaseData("a += b")
+                        .Returns("a".Length);
+                    yield return new TestCaseData("a << b")
```
elsassph: Maybe needs a case `a+b` without spaces?

SlavaRa: 5 min

SlavaRa: fixed

External/Plugins/ASCompletion/Completion/ASComplete.cs
SlavaRa: `a += b`
groupCount is always zero

External/Plugins/ASCompletion/Completion/ASComplete.cs
SlavaRa: `test some` -> break before `s`
`test           ()` -> break after `()`

SlavaRa: We can not know in advance at what distance from the text the opening bracket will be and whether it will be at all.

SlavaRa:
@elsassph I deleted the non-obvious code. Please review it.

PR #1941, opened 2018-01-10
[Haxe][Context] Fixed autocompletion for single char String. Completion Mode: FlashDevelop

SlavaRa:

PR #1942, opened 2018-01-10
[Haxe][Context] Fixed autocompletion for single char String. Completion Mode: FlashDevelop

SlavaRa:


SlavaRa:
I broke autocompletion for single char String. #1926. Sorry :(
Solutions:
- Revert #1926
- Check and merge this fix, I believe this is the optimal solution.
- Check and merge #1941. I do not like this decision, but it retains the idea of the method(ASComplete.HandleDotCompletion:2067), although I consider this behavior to be incorrect.
Thanks.

Issue #1898, opened 2018-01-04
[Bug][Debug] Internal Debugger Exception


benjstern:
![image](https://user-images.githubusercontent.com/1266727/34554832-8c7ee2c8-f137-11e7-995d-a1a812d40a4b.png)

Nothing is sent to the output when I check the exeptions.log I get:

2018-01-04 10:03:01 AM

Internal Debugger Exception

System.NullReferenceException: Object reference not set to an instance of an object.
   at net.sf.jni4net.jni.JNIEnv.CallStaticObjectMethodPtr(Class clazz, MethodId methodIdNative, Value[] args)
   at flash.tools.debugger.Bootstrap.sessionManager()
   at FlashDebugger.FlashInterface.Start()
   at FlashDebugger.DebuggerManager.bgWorker_DoWork(Object sender, DoWorkEventArgs e)

At first run the application works and debugs fine, at subsequent runs it's shows this error message.

I'm running FlashDevelop 5.3.1.67 Development version.

Originally thought it was related to this: https://github.com/fdorg/flashdevelop/issues/1855, but upon further investigation seems to be a separate issue.

benjstern:
Same with Development version 5.3.1.67

Debugger startup error. For troubleshooting see: http://www.flashdevelop.org/wikidocs/index.php?title=F.A.Q
Error details: net.sf.jni4net.jni.JNIException: Can't load java class for flash.tools.debugger.concrete.DValue_ from classLoader sun.misc.Launcher$AppClassLoader@757aef ---> java.lang.LinkageError: flash/tools/debugger/concrete/DValue_ : Unsupported major.minor version 52.0
at net.sf.jni4net.jni.JNIEnv.ExceptionTest()
at net.sf.jni4net.jni.JNIEnv.CallObjectMethodPtr(IJvmProxy obj, MethodId methodIdNative, Value[] args)
at java.lang.ClassLoader.loadClass(String par0)
at net.sf.jni4net.utils.Registry.LoadClass(String name, ClassLoader classLoader, JNIEnv env)
--- End of inner exception stack trace ---
at net.sf.jni4net.utils.Registry.LoadClass(String name, ClassLoader classLoader, JNIEnv env)
at net.sf.jni4net.utils.Registry.RegisterClass(RegistryRecord record, ClassLoader classLoader, JNIEnv env)
at net.sf.jni4net.utils.Registry.BindJvm(RegistryRecord record, ClassLoader classLoader, JNIEnv env)
[...]

Issue #1855, opened 2017-12-15
as3 debugger not starting in 5.3.0.3


romamik:
I've just installed new flashdevelop version and get this error, when trying to start as3 debugger:

> Debugger startup error. For troubleshooting see: http://www.flashdevelop.org/wikidocs/index.php?title=F.A.Q
> Error details: net.sf.jni4net.jni.JNIException: Can't find jni4net.j-0.8.8.0.jar
> at c:\program files (x86)\flashdevelop\jni4net.j-0.8.8.0.jar
> or c:\program files (x86)\flashdevelop\bin\jni4net.j-0.8.8.0.jar
> or c:\program files (x86)\flashdevelop\lib\jni4net.j-0.8.8.0.jar
> or c:\program files (x86)\lib\jni4net.j-0.8.8.0.jar
> or C:\Program%20Files%20(x86)\FlashDevelop\Plugins\jni4net.j-0.8.8.0.jar
> or C:\Program%20Files%20(x86)\FlashDevelop\Plugins\bin\jni4net.j-0.8.8.0.jar
> or C:\Program%20Files%20(x86)\FlashDevelop\Plugins\lib\jni4net.j-0.8.8.0.jar
> or C:\Program%20Files%20(x86)\FlashDevelop\lib\jni4net.j-0.8.8.0.jar
>    в net.sf.jni4net.Bridge.FindJar()
>    в net.sf.jni4net.BridgeSetup.AddBridgeClassPath()
>    в net.sf.jni4net.BridgeSetup..ctor(Boolean autoAddBridgeClassPath)
>    в FlashDebugger.DebuggerManager.Start(Boolean alwaysStart)

Required file is present at path `C:\Program Files (x86)\FlashDevelop\Plugins\jni4net.j-0.8.8.0.jar`.
Looks like search path is url encoded while it should not. 

As a temporary solution I've reinstalled FlashDevelop to `C:\FlashDevelop` and that solved the issue for me.

SlavaRa:
Please check development build from http://flashdevelop.org/downloads/builds/

romamik:
It works as intended in development build...

pzahra:
Development version 5.3.1.52

> Debugger startup error. For troubleshooting see: http://www.flashdevelop.org/wikidocs/index.php?title=F.A.Q
[...]

PR #1956, opened 2018-01-16
Update jni4net

### LINKED ISSUE #1787: Can't find jni4net.j-0.8.8.0.jar in escaped path. Version 5.3.0.3
Escaping the path below is causing an issue finding the needed java files.  The jar file is indeed located at this path on my install of FlashDevelop. 

C:\Program%20Files%20(x86)\FlashDevelop\Plugins\jni4net.j-0.8.8.0.jar

Temporary fix:  I copied the java files below to the FlashDevelop root folder as below.
jni4net.j-0.8.8.0.jar
jni4net.n.w32.v20-0.8.8.0.dll
jni4net.n.w32.v40-0.8.8.0.dll

Copy to root folder
c:\program files (x86)\flashdevelop\

Neverbirth:
I built a new jni4net from the latest source code. It fixes #1787, and maybe solves/improves #1898 and #1855.

While doing this I also saw 2 other problems... maybe it would be better if they would be 3 separate PRs, but it would have needed a bit more of time...

1. If the data tip was never initialized it was not possible to stop the debugging (maybe there were other cases failing?). Strange nobody ever got this problem, or maybe there is some issue regarding this?
2. No reason to run the Haxe diagnostics mode if the current project is not Haxe, this also solves a crash that could happen when having a Haxe file open, and the first project to be loaded is not a Haxe one.


External/Plugins/HaXeContext/Linters/DiagnosticsLinter.cs
```diff
@@ -27,7 +27,7 @@ public void LintAsync(IEnumerable<string> files, LintCallback callback)
         {
             var context = ASContext.GetLanguageContext("haxe") as Context;
 
-            if (context == null || !CanContinue(context)) return;
+            if (context == null || !(PluginBase.CurrentProject is ProjectManager.Projects.Haxe.HaxeProject) || !CanContinue(context)) return;
```
SlavaRa: One would not try to get the context if the project is not haxe. Yes, I understand that this is a miserable optimization, but it seems so much more readable.

PR #1957, opened 2018-01-17
[ASComplete][GetExpression] Fixed autocompletion for the first line within function that returns a generic type. fixes #1954

### LINKED ISSUE #1954: [Bug] No completion if function returns <T> value.
If function return any class with type parameters, completion breaks and does not provides any completion.
Example:
```haxe
private function getStuff():Array<Stuff>
{
 ret // <- No completion provided for `return`.
}
```

SlavaRa:


Yanrishatum:
Tested, can't reproduce issue anymore.

PR #1968, opened 2018-01-22
Add more exceptions to UITools.HandleEvent to not hide CompletionList

SlavaRa:


Yanrishatum:
Tested, no longer does irritating blinking before showing Haxe completion.

PR #1977, opened 2018-01-25
[Haxe][Context] Support `using` keyword for autocomplete

SlavaRa:
It is draft version, please don't merge this PR.

### DIFF
--- a/External/Plugins/AS2Context/Context.cs
+++ b/External/Plugins/AS2Context/Context.cs
@@ -570,6 +570,14 @@ namespace AS2Context
                 if (first == '{' && last == '}') return ResolveType(features.objectKey, inFile);
                 if (first == '[' && last == ']') return ResolveType(features.arrayKey, inFile);
                 if (first == '"' || first == '\'') return ResolveType(features.stringKey, inFile);
+                var m = Regex.Match(token, "^new\\s");
+                if (m.Success)
+                {
+                    var index = token.IndexOfOrdinal("(");
+                    var startIndex = m.Index + m.Length;
+                    if (index != -1) return ResolveType(token.Substring(startIndex, index - startIndex), inFile);
+                    return ResolveType(token.Substring(startIndex), inFile);
+                }
             }
             return base.ResolveToken(token, inFile);
         }
--- a/External/Plugins/HaXeContext/Context.cs
+++ b/External/Plugins/HaXeContext/Context.cs
@@ -905,30 +905,30 @@ namespace HaXeContext
         /// <returns>A parsed class or an empty ClassModel if the class is not found</returns>
         public override ClassModel ResolveType(string cname, FileModel inFile) => ResolveType(cname, inFile, true);
 
-        ClassModel ResolveType(string cname, FileModel inFile, bool resolveUsings)
+        ClassModel ResolveType(string className, FileModel inFile, bool resolveStaticExtensions)
         {
             var result = ClassModel.VoidClass;
-            if (!string.IsNullOrEmpty(cname) && cname != features.voidKey && classPath != null)
+            if (!string.IsNullOrEmpty(className) && className != features.voidKey && classPath != null)
             {
                 // handle generic types
-                if (cname.Contains('<'))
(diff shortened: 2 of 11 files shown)


External/Plugins/AS2Context/Context.cs
```diff
@@ -570,6 +570,14 @@ public override ClassModel ResolveToken(string token, FileModel inFile)
                 if (first == '{' && last == '}') return ResolveType(features.objectKey, inFile);
                 if (first == '[' && last == ']') return ResolveType(features.arrayKey, inFile);
                 if (first == '"' || first == '\'') return ResolveType(features.stringKey, inFile);
+                var m = Regex.Match(token, "^new\\s");
```
elsassph, line 573: The convention is to pre-compile all those often used regexes as static vars.

External/Plugins/HaXeContext/Context.cs
```diff
@@ -903,65 +903,143 @@ public override bool IsImported(MemberModel member, int atLine)
...
-            if (string.IsNullOrEmpty(cname) || cname == features.voidKey || classPath == null)
-                return ClassModel.VoidClass;
+        public override ClassModel ResolveType(string cname, FileModel inFile) => ResolveType(cname, inFile, true);
 
-            // handle generic types
-            if (cname.IndexOf('<') > 0)
+        public ClassModel ResolveType(string className, FileModel inFile, bool resolveStaticExtensions)
+        {
+            var result = ClassModel.VoidClass;
+            if (!string.IsNullOrEmpty(className) && className != features.voidKey && classPath != null)
```
elsassph, line 911: I prefer guarding cases to nesting many `if`s... Moreover the guarding change makes the diff difficult to follow.

I suppose the reason is that you wanted to lookup static extensions for the resolved type in all the cases. A better approach would have been to move the entire logic of `ResolveType` into an internal function, and apply the static extension resolution to the result.

External/Plugins/HaXeContext/Context.cs
```diff
@@ -903,65 +903,143 @@ public override bool IsImported(MemberModel member, int atLine)
...
+        /// <summary>
+        /// https://haxe.org/manual/lf-static-extension.html
+        /// </summary>
+        /// <param name="target"></param>
+        /// <param name="inFile">Current file</param>
+        /// <returns></returns>
+        public MemberList ResolveStaticExtensions(ClassModel target, FileModel inFile)
+        {
+            var result = new MemberList();
+            IEnumerable<MemberModel> imports = inFile.Imports.Items;
```
elsassph, line 994: Can we use the `completionCache` to avoid re-doing the imports resolution here?

External/Plugins/HaXeContext/Context.cs
```diff
@@ -903,65 +903,143 @@ public override bool IsImported(MemberModel member, int atLine)
...
+            {
+                var import = importModels[i];
+                var type = ResolveType(import.Name, inFile, false);
+                if (type.IsVoid() || type.Members.Count == 0) continue;
+                var access = TypesAffinity(target, type);
+                var extends = target;
+                while (!extends.IsVoid())
                 {
-                    if (import.Name == cname)
+                    foreach (MemberModel member in type.Members)
```
elsassph: Is that something that potentially could be part of the `completionCache`? So at this point we only iterate over the potential static extension and select the ones compatible.

External/Plugins/HaXeContext/Context.cs
```diff
@@ -903,65 +903,143 @@ public override bool IsImported(MemberModel member, int atLine)
...
+                    extends.ResolveExtends();
+                    extends = extends.Extends;
                 }
-                if (!found && cname == "Function") return stubFunctionClass;
             }
-
-            return GetModel(package, cname, inPackage);
+            if (result.Count > 0)
+            {
+                result.Items.RemoveAll(extension =>
```
elsassph, line 1030: Can you add some comments explaining what the different phases of this function do? This last one isn't clear in particular.

PR #1978, opened 2018-01-26
[Haxe][ASFileParser] Added support for parsing of `from` and `to` keywords within header of `abstract`

SlavaRa:


External/Plugins/ASCompletion/Model/ASFileParser.cs
```diff
@@ -1895,7 +1895,7 @@ private bool EvalToken(bool evalContext, bool evalKeyword, int position)
 
             string token = curToken.Text;
             int dotIndex = token.LastIndexOf('.');
-            if (evalKeyword && (token.Length > 2))
+            if (evalKeyword && (token.Length > 2 || (haXe && token.Length >= 2)))
```
elsassph: Maybe have a variable `var minTokenLength = haXe ? 2 : 3;`?

PR #1996, opened 2018-02-06
[ASGenerator] Contextual generator shouldn't outside the class, e.g. 'generate toString()'. fixes #1995

### LINKED ISSUE #1995: [ASGenerator] Contextual generator shouldn't outside the class, e.g. `generate toString()`

SlavaRa:


External/Plugins/ASCompletion/Completion/ASGenerator.cs
```diff
@@ -441,25 +441,58 @@ protected virtual void ContextualGenerator(ScintillaControl sci, int position, A
             // TODO: Empty line, show generators list? yep
         }
 
-        protected virtual bool CanShowImplementInterfaceList(ASResult resolve)
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="sci">The Scintilla control containing the document</param>
+        /// <param name="position">Cursor position</param>
```
SlavaRa: I think that cursor position should be used from `expr`

PR #2023, opened 2018-02-14
Improvement work with optional parameters. step 1. fixes #2022

### LINKED ISSUE #2022: variable name resolves to ?name if type is function
When local function is declared with optional type with type of function being optional function variable name resolves to `?name` instead of `name` . Only happens if declared inside local function. 

Isolated reproduction:
```haxe
function onInitialize() {
    function some(?val:Bool->Void) {
      ?val
    }
}
```

SlavaRa:


Yanrishatum:
Tested, no longer can reproduce both in isolated example or project where discovered it.

PR #2024, opened 2018-02-14
Update MXML.fdt

SandersAlex:
halo - is theme
spark & mx - core


Neverbirth:
Thanks for the contribution

PR #2046, opened 2018-02-23
[Haxe][Completion] Added autocomplete after 'from' and 'to'

SlavaRa:
autocompleteafterfromto

### DIFF
--- a/External/Plugins/ASCompletion/Completion/ASComplete.cs
+++ b/External/Plugins/ASCompletion/Completion/ASComplete.cs
@@ -153,33 +153,7 @@ namespace ASCompletion.Completion
                         break;
 
                     case ' ':
-                        position--;
-                        string word = GetWordLeft(Sci, ref position);
-                        if (word.Length == 0)
-                        {
-                            char c = (char)Sci.CharAt(position);
-                            if (c == ':' && features.hasEcmaTyping)
-                                return HandleColonCompletion(Sci, "", autoHide);
-                            else if (autoHide && (c == '(' || c == ',') && !ASContext.CommonSettings.DisableCallTip)
-                                return HandleFunctionCompletion(Sci, autoHide);
-                            break;
-                        }
-                        if (word == "package" || Array.IndexOf(features.typesKeywords, word) >= 0) 
-                            return false;
-                        // new/extends/instanceof/...
-                        if (features.HasTypePreKey(word)) return HandleNewCompletion(Sci, "", autoHide, word);
-                        var beforeBody = true;
-                        var expr = CurrentResolvedContext?.Result?.Context;
-                        if (expr != null) beforeBody = expr.ContextFunction == null || expr.BeforeBody;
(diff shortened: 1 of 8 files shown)


External/Plugins/ASCompletion/Completion/ASComplete.cs
```diff
@@ -153,33 +153,7 @@ static public bool OnChar(ScintillaControl Sci, int Value, bool autoHide)
...
-                        if (expr != null) beforeBody = expr.ContextFunction == null || expr.BeforeBody;
-                        if (!beforeBody && features.codeKeywords.Contains(word)) return false;
-                        // override
-                        if (word == features.overrideKey) return ASGenerator.HandleGeneratorCompletion(Sci, autoHide, word);
-                        // import
-                        if (features.hasImports && (word == features.importKey || word == features.importKeyAlt))
-                            return HandleImportCompletion(Sci, "", autoHide);
-                        // public/internal/private/protected/static
-                        if (Array.IndexOf(features.accessKeywords, word) >= 0)
-                            return HandleDeclarationCompletion(Sci, "", autoHide);
```
SlavaRa, line 182: lines 156-182 extracted to `private bool HandleWhiteSpaceCompletion(ScintillaControl sci, int position, bool autoHide)`

External/Plugins/HaXeContext/Completion/CodeComplete.cs
```diff
@@ -48,5 +49,42 @@ public override bool IsStringInterpolationStyle(ScintillaControl sci, int positi
...
+            var positionFrom = sci.PositionFromLine(member.LineFrom);
+            for (var i = positionFrom; i < position; i++)
+            {
+                if (sci.PositionIsOnComment(position)) continue;
+                var c = (char)sci.CharAt(i);
+                if (c == '(' || c == '<') groupCount++;
+                else if (c == ')' || (c == '>' && sci.CharAt(i - 1) != '-')) groupCount--;
+                else if (c == '{' && groupCount == 0)
+                {
+                    if (i <= position) return false;
```
elsassph: This test is always true, isn't it?

SlavaRa: Nope, for example:
```csharp
yield return new TestCaseData("BeforeBody_2")
                    .SetName("class Some {|}")
                    .Returns(false);
yield return new TestCaseData("BeforeBody_4")
                    .SetName("class Some {}|")
                    .Returns(false);
```

elsassph: `return false` is correct, but I mean the loop is `for (var i = positionFrom; i < position; i++)` so `i` is always less than `position`

SlavaRa: Yes, but I do not fully understand, you see this as a mistake, or it is not immediately clear why this is done?

Issue #1812, opened 2017-11-28
Source control add via dialog then rename leads to old file still in index


gene-pavlovsky:
Create a new file via the Project panel, a nice new dialog offers to add it to source control. Answer Yes.
Rename the file in the Project panel.

Expected results: Using git terminology, the file with the original name should be unstaged, and staged with the new name.
Actual results: The file with it's original name is still staged.

Neverbirth:
@chipshort

gene-pavlovsky:
Testing the PR:

1. Created a new file "src/Foo.hx" using the Project panel, git status:
```
        new file:   src/Foo.hx
```

2. Renamed the file using project panel right-click > rename, answered NO to "update references" dialog, git status:
```
        new file:   src/Bar.hx
```

Looks good so far.

Problem 1:
3. Removed the file using project panel right-click > delete, answered OK to "file will be moved to Recycle Bin", git status:
```
        new file:   src/Quux.hx

Changes not staged for commit:
        deleted:    src/Quux.hx
```

Not so good, file should be unstaged after deletion.

Problem 2: Repeating the same test with Foo.hx->Bar.hx, I quickly answered YES to "update references" dialog. In this case "Foo.hx" was not unstaged and "Bar.hx" was not added. It doesn't happen every time so I think there is some general unreliability regarding the dialog callbacks being called or something like that.

Problem 3: 
- Create a new file "Quux.hx", don't do anything with the "Would you like to add this file to version control?" action bar. 
- Delete the file with the project panel right-click > Delete
Expected result: "recycle bin" dialog, OK, file deleted, no changes to git status.
Actual result: "Would you like to remove files from version control?", OK, error dialog: "fatal: pathspec 'Quux.hx' did not match any files" (since the file was not `git add`ed), file is not actually deleted
Additional note: When the file is just created, it's icon in the project panel is wrong - it is green, same as "up to date" files under version control, while actually it's not. And I guess this is why the "version control" dialog pops up when deleting. If you wait just a few seconds, the icon changes to gray and trying to delete the file results in expected behavior.

PR #2047, opened 2018-02-23
[SourceControl] When file was added and is moved, remove old file from index

chipshort:
As requested in #1812


SlavaRa:
looks like good

gene-pavlovsky:
I tested the PR and while it improves the behavior, the original issue is still not fully fixed.
I found two more issues while testing, and described all of them in a comment to the original issue. If desired, I can split them to separate issues.

SlavaRa:
WIP

SlavaRa:
via #3099

Issue #661, opened 2015-04-10
[haxe] Consider adding "from" and "to" to keyword or word3 group


YellowAfterlife:
These can be used in definitions of [abstract types](http://haxe.org/manual/types-abstract-implicit-casts.html), most commonly like so:
```haxe
abstract AbstractName(BaseType) from BaseType to BaseType {
    // ...
}
```

Gama11:
Probably not gonna happen, the Scintilla Lexer's keywords aren't contextual, and you wouldn't want these highlighted elsewhere in your code (similar to get / set / never in property declarations).

Meychi:
I agree with @Gama11, would be great to have a dedicated lexer for Haxe in Scintilla.

PR #2083, opened 2018-03-09
Update Haxe.xml

markknol:
* Add final keyword (since Haxe 4 preview 2)
* Add `as` (since Haxe 3.2.0) Haxe also allows the more natural `as` in place of `in` https://haxe.org/manual/type-system-import.html#since-haxe-3.2.0
* ~Add from/to, used in `abstract` https://haxe.org/manual/types-abstract-implicit-casts.html~

Not sure if this is the correct file to add it, I didnt compile the project and also saw this file https://github.com/fdorg/flashdevelop/blob/master/External/Plugins/HaXeContext/Context.cs#L123


Gama11:
https://github.com/fdorg/flashdevelop/issues/661

SlavaRa:
@markknol I probably will try implements this functionality, but I can't promise that is will be quickly.

SlavaRa:
- [x] as
- [x] final

PR #2093, opened 2018-03-24
[CodeComplete] Show only interfaces for completion after 'context.Features.ImplementsKey'

SlavaRa:


SlavaRa:
Not implemented case:
```haxe
interface IFoo {
    function foo():Void;
}
typedef TFoo = IFoo;
class Foo implements TFoo {
    public function new() {}
    public function foo() {}
}
```

PR #2107, opened 2018-04-11
Fix strange race condition with PathModel.

Neverbirth:
I ran into some strange crash when starting FlashDevelop: `PathModel.HasFile` and `PathModel.GetFile` are 2 independent functions locking over the collection of files, so it's perfectly possible that between both calls a thread modifies the collection and causes unhandled exceptions. This was a really bizarre crash, and the application state at the moment of the crash made it unable for me to come up with the reason behind.

@elsassph since this is mostly your code it would be nice if you can validate and merge.

### DIFF
--- a/External/Plugins/AS2Context/Context.cs
+++ b/External/Plugins/AS2Context/Context.cs
@@ -771,9 +771,9 @@ namespace AS2Context
             {
                 string path = Path.Combine(aPath.Path, fileName);
                 // cached file
-                if (aPath.HasFile(path))
+                FileModel nFile;
+                if (aPath.TryGetFile(path, out nFile))
                 {
-                    FileModel nFile = aPath.GetFile(path);
                     if (nFile.Context != this)
                     {
                         // not associated with this context -> refresh
@@ -785,7 +785,7 @@ namespace AS2Context
                 // non-cached existing file
                 else if (File.Exists(path))
                 {
-                    FileModel nFile = GetFileModel(path);
+                    nFile = GetFileModel(path);
                     if (nFile != null)
                     {
                         aPath.AddFile(nFile);
--- a/External/Plugins/ASCompletion/Model/PathModel.cs
+++ b/External/Plugins/ASCompletion/Model/PathModel.cs
@@ -41,14 +41,17 @@ namespace ASCompletion.Model
         /// </summary>
         static public void Compact()
         {
-            Dictionary<string, PathModel> clean = new Dictionary<string, PathModel>();
-            foreach (string key in pathes.Keys)
+            lock (pathes)
             {
-                PathModel model = pathes[key];
-                if (model.InUse) clean.Add(key, model);
-                else model.Cleanup();
+                Dictionary<string, PathModel> clean = new Dictionary<string, PathModel>();
+                foreach (string key in pathes.Keys)
+                {
+                    PathModel model = pathes[key];
+                    if (model.InUse) clean.Add(key, model);
+                    else model.Cleanup();
+                }
+                pathes = clean;
             }
-            pathes = clean;
         }
 
         /// <summary>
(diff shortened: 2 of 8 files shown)


External/Plugins/ASCompletion/Model/PathModel.cs
```diff
@@ -41,19 +41,14 @@ static public void ClearAll()
         /// </summary>
         static public void Compact()
         {
-            lock (pathes)
```
Neverbirth: This is the only part of the code locking `pathes`, so it doesn't make sense at all.

External/Plugins/ASCompletion/Model/PathModel.cs
```diff
@@ -85,7 +79,6 @@ static public PathModel GetModel(string path, IASContext context)
         public volatile bool Updating;
         public bool WasExplored;
         public bool IsTemporaryPath;
-        public DateTime LastAccess;
```
Neverbirth, line 88: These `LastAccess` and `Touch()` fields are dead code by now, so I went on and removed them.

External/Plugins/ASCompletion/Model/PathModel.cs
Neverbirth: Actually after second thought, even if the whole `pathes` thing is not fully thread-safe this code adds some safety to the `Compact` function itself because of the assignment at the end, but that's something easily solvable without fully locking.

I will try to commit a change here ASAP. Anyway, @elsassph, it would be great if you could check all this stuff in the meantime.

elsassph: Yes I used `pathes` as lock object - maybe you should wrap in a `lock (lockObject)` section.

External/Plugins/AS2Context/Context.cs
```diff
@@ -771,9 +771,9 @@ private ClassModel LocateClassFile(PathModel aPath, string fileName)
             {
                 string path = Path.Combine(aPath.Path, fileName);
                 // cached file
-                if (aPath.HasFile(path))
+                FileModel nFile;
+                if (aPath.TryGetFile(path, out nFile))
```
elsassph, line 775: Looks good to me!

External/Plugins/ASCompletion/Model/PathModel.cs
```diff
@@ -525,6 +517,20 @@ public bool HasFile(string fileName)
             }
         }
 
+        public bool TryGetFile(string fileName, out FileModel value)
+        {
+            if (!IsValid)
```
elsassph, line 525: Should it in the locked section?

External/Plugins/ASCompletion/Model/PathModel.cs
```diff
@@ -584,11 +587,6 @@ public void RemoveFile(string fileName)
             }
         }
 
-        public void Touch()
-        {
-            LastAccess = DateTime.Now;
```
elsassph, line 589: Yes the idea initially was to use the `LastAccess` info to keep recent cps in memory.

External/Plugins/ASCompletion/Model/PathModel.cs
SlavaRa: it's strange moment

Neverbirth: The rest of code doesn't include the IsValid section inside the lock, so I left the same logic.

External/Plugins/ASCompletion/Model/PathModel.cs
Neverbirth: That would be wrong, lockObject is specific to each PathModel instance, and these are static fields. I guess every pathes access should be locked, not just this one... but since I never saw any error with it I guess it's not that much of a problem?

Neverbirth: I decided to put back the locking over `pathes`. Making this part of the logic thread-safe without locking is simple, but I don't feel comfortable doing so without checking the rest of parts using `pathes`.

PR #2112, opened 2018-04-17
#2103 add support for lix and haxeshim

### LINKED ISSUE #2103: Add support for lix/haxeshim
1. FD relies on parsing the Haxe SDK `CHANGES.txt` file to determine the SDK version. This breaks when trying to add `haxeshim` as the Haxe SDK.
Proposed changes: When adding an SDK:
- run `haxe --run show-version` (haxeshim feature) to check for haxeshim
- if failed, run `haxe -version` (haxe feature) to get the version
- if failed, try to get the version from `CHANGES.txt`
Another option - check if project has a `.haxerc` file to detect haxeshim

2. FD relies on `haxelib` to get the standard library path and to resolve other libraries. This breaks if the project uses `haxeshim`/`lix` for haxe version management and library management, respectively.
Proposed changes:
- if the selected SDK is actually `haxeshim`, use features provided by it to resolve libraries
- otherwise, use `haxelib`

gene-pavlovsky:
Closes #2103.

Testing is needed!

Summary of changes:
- Haxe SDK validation checks if `haxe.exe` is in fact haxeshim, if not, uses `haxe.exe -version` to determine the version, and finally falls back to the original code that extracts the version from `CHANGES.txt`.
When using a `haxeshim` SDK:
- Re-check the haxe version on every project create/open/close event. It would probably be nice to also do it when opening then closing project properties.
- Support lix-managed libraries (completion works)
- Support for installing missing libraries via lix

Potential improvements:
- Because library A might require library B which is missing, and library B is also specified as a dependency, multiple "missing library" dialogs might be shown for the same library. This is exacerbated by linting diagnostics helper which can also trigger the "missing library" dialog. It would be nice to improve the whole "missing library" feature in the future, e.g. avoiding multiple dialogs for the same library, showing which library is missing in the dialog. Personally I would prefer to see all the missing libraries in a single dialog, with a button to install them all at once. It especially makes sense when using `lix` since a single `lix download` command should get all the dependencies. It's rather pointless to install them one by one, at least in this case.
- ? Use cache with some form of smart invalidation for lix library class paths (currently it is disabled) ?
- Fix any discovered bugs

### DIFF
--- a/External/Plugins/ProjectManager/Settings.cs
+++ b/External/Plugins/ProjectManager/Settings.cs
@@ -36,7 +36,7 @@ namespace ProjectManager
         string[] excludedFileTypes = new string[] { ".p", ".abc", ".bak", ".tmp" };
         string[] excludedDirectories = new string[] { ".svn", "_svn", ".cvs", "_cvs", "cvs", "_sgbak", ".git", ".hg", "node_modules" };
         string[] executableFileTypes = new string[] { ".exe", ".lnk", ".fla", ".flump", ".doc", ".pps", ".psd", ".png", ".jpg", ".gif", ".xls", ".docproj", ".ttf", ".otf", ".wav", ".mp3", ".ppt", ".pptx", ".docx", ".xlsx", ".ai", ".pdf", ".zip", ".rar" };
-        string[] filteredDirectoryNames = new string[] { "src", "source", "sources", "as", "as2", "as3", "actionscript", "flash", "classes", "trunk", "svn", "git", "hg", "..", "." };
+        string[] filteredDirectoryNames = new string[] { "src", "source", "sources", "as", "as2", "as3", "actionscript", "flash", "classes", "trunk", "svn", "git", "hg", "github", "gitlab", "haxelib", "library", "..", "." };
 
         HighlightType tabHighlightType = HighlightType.ExternalFiles;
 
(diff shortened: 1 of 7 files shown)


External/Plugins/ProjectManager/Settings.cs
```diff
@@ -36,7 +36,7 @@ public class ProjectManagerSettings
         string[] excludedFileTypes = new string[] { ".p", ".abc", ".bak", ".tmp" };
         string[] excludedDirectories = new string[] { ".svn", "_svn", ".cvs", "_cvs", "cvs", "_sgbak", ".git", ".hg", "node_modules" };
         string[] executableFileTypes = new string[] { ".exe", ".lnk", ".fla", ".flump", ".doc", ".pps", ".psd", ".png", ".jpg", ".gif", ".xls", ".docproj", ".ttf", ".otf", ".wav", ".mp3", ".ppt", ".pptx", ".docx", ".xlsx", ".ai", ".pdf", ".zip", ".rar" };
-        string[] filteredDirectoryNames = new string[] { "src", "source", "sources", "as", "as2", "as3", "actionscript", "flash", "classes", "trunk", "svn", "git", "hg", "..", "." };
+        string[] filteredDirectoryNames = new string[] { "src", "source", "sources", "as", "as2", "as3", "actionscript", "flash", "classes", "trunk", "svn", "git", "hg", "github", "gitlab", "haxelib", "library", "..", "." };
```
gene-pavlovsky, line 39: Existing FD users will have to update this setting in the Project Manager settings. If they don't, they will see multiple "haxelib/src" in the `References` node in the project manager panel.

gene-pavlovsky: Solved by https://github.com/fdorg/flashdevelop/pull/2120

PR #2163, opened 2018-06-23
Allow more list view shortcut keys in the keyboard shortcut dialog.

### LINKED ISSUE #2145: PgUp/PgDn don't work in the keyboard shortcuts dialog
Up and down cursors work fine. Seems that PgUp/PgDn aren't valid shortcut keys, so they should presumably just work as normal?

Thanks,

--Tom

tom-seddon:
Fix #2145.

Thanks,

--Tom

### DIFF
--- a/FlashDevelop/Dialogs/ShortcutDialog.cs
+++ b/FlashDevelop/Dialogs/ShortcutDialog.cs
@@ -414,7 +414,19 @@ namespace FlashDevelop.Dialogs
             var item = (ShortcutListItem) this.listView.SelectedItems[0];
             this.AssignNewShortcut(item, e.KeyData);
             // Don't trigger list view default shortcuts like Ctrl+Add
-            if (e.KeyData != Keys.Up && e.KeyData != Keys.Down) e.Handled = true;
+            switch (e.KeyData) {
+                case Keys.Up:
+                case Keys.Down:
+                case Keys.PageDown:
+                case Keys.PageUp:
+                case Keys.Home:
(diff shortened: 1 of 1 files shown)


FlashDevelop/Dialogs/ShortcutDialog.cs
```diff
@@ -414,7 +414,19 @@ private void ListViewKeyDown(object sender, KeyEventArgs e)
...
             this.AssignNewShortcut(item, e.KeyData);
             // Don't trigger list view default shortcuts like Ctrl+Add
-            if (e.KeyData != Keys.Up && e.KeyData != Keys.Down) e.Handled = true;
+            switch (e.KeyData) {
+                case Keys.Up:
+                case Keys.Down:
+                case Keys.PageDown:
+                case Keys.PageUp:
+                case Keys.Home:
+                case Keys.End:
```
SlavaRa, line 423:
```csharp
case Keys.Home:
case Keys.End:
```
will not be handled here.

tom-seddon: I don't follow? - when you press Home or End with focus on the list view, this handler gets called with `Keys.Home` or `Keys.End` as the `KeyData`.

Thanks,

--Tom

SlavaRa: Oh I'm sorry man. I was check other ;( I think that I can check and merge this tomorrow.

PR #2189, opened 2018-07-09
Haxe: Haxeshim multi sdk fix, custom SDK improvement

gene-pavlovsky:
Fixes an issue when using a `haxeshim` SDK in a project, while the default SDK is a regular `haxe` SDK.
The issue is caused by `GetCompilerPath` returning the default SDK path, not the current SDK path. As a result FD tries to resolve lix library dependencies using non-haxeshim `haxe`, which fails.
Actually `GetCompilerPath`'s description says "Retrieve the context's default compiler path", so perhaps the function is doing what it originally was intended to. But I don't see the point of that, what is it for? We should always try to use the currently selected SDK if possible, with a fall back to the default one (e.g. when there is no project selected).

Fixes at least two issues when using a "custom SDK" set in the project properties. In this case, there was no SDK validation performed, as a result haxeshim was not detected and didn't work, and haxe version was considered 0.0.0 so (potentially) wrong completion options were selected. This PR adds custom SDK validation and caching.

**Note**: things are very similar in AS3Context, also some parts of code there use `GetCompilerPath` and some just use `as2settings.GetDefaultSDK().Path`. I'm afraid it might also have some issues when the project SDK differs significantly from the default SDK.


External/Plugins/HaXeContext/Context.cs
```diff
@@ -702,9 +710,15 @@ public override void RemoveClassCompilerCache()
...
-        {
-            return hxsettings.InstalledSDKs?.FirstOrDefault(sdk => sdk.Path == currentSDK);
+        private InstalledSDK GetCurrentSDK() => hxsettings.InstalledSDKs?.FirstOrDefault(sdk => sdk.Path == currentSDK) ?? GetCustomSDK(currentSDK);
+
+        private InstalledSDK GetCustomSDK(string customPath)
+        {
+            InstalledSDK sdk;
+            if (customSDKCache.TryGetValue(customPath, out sdk)) return sdk;
+            sdk = createCustomSDK(customPath);
+            if (sdk != null) customSDKCache.Add(customPath, sdk);
```
elsassph, line 720: @gene-pavlovsky @SlavaRa I'm a bit worried about this static cache: shoulnd't it be cleared in some scenarios?

SlavaRa: I'll try to find time today and think about this improvement. Perhaps this is a very correct proposal.

gene-pavlovsky: @elsassph Yes you're right. I see two potential issues here:
- Switch from Project A to Project B. Both projects use the same custom SDK which is Haxeshim, but they have different compiler versions set in `.haxerc`. In this case FD will think Project B is using whatever version Project A is using, and might e.g. use wrong completion options.
- A haxe SDK at the custom SDK path is upgraded while FD is open. I guess that will also be a problem with regular installed SDKs.
Do you see other problem scenarios?

It might be better to keep the cache in `PluginMain` instead of `Context`, and add ValidateSDK call in same cases as for installed SDKs. Might be also worth to clear the cache when switching projects, I think it's ok to not clear it, just re-validate as necessary.

gene-pavlovsky: I tried to resolve these concerns in this PR:
https://github.com/fdorg/flashdevelop/pull/2206

Issue #2194, opened 2018-07-10
Source control: wrong messages and other issues


gene-pavlovsky:
It seems that there are a lot of source control dialog related bugs, many of them cause by the complexity of the various copy/move/refactor actions possible in FD and the way events are raised and processed. It would be hard to get rid of every such bug. It was discussed previously that there shouldn't even be an option to offer to create commits for actions: https://github.com/fdorg/flashdevelop/issues/1932#issuecomment-356249121
I'm obviously going back to setting the "never create a commit for moving, deleting, ..." option to true (should be `true` by default in my opinion).

Some of the issues I found:

1. I renamed a file in the project panel, and then I've got this excellent suggestion asking me if I would like to create a commit for this action, with this as a message:
`Moved src\Foo.hx$renaming$ to src\foo.hx`
Doesn't look right, does it?
It's worth to mention that this only happens when the rename involves only a case change, and the strange message comes from this line in `RenameFile.cs`: `string tmpPath = oldPath + "$renaming$";`
Same cause as #1288 

2. Another case of wrong message: Just adding a new folder `foo` via the project panel, commit dialog offers to create a commit with this message: "Moved src\New Folder to src\foo".

3. A freshly created directory `src/foo/` contains several source files (not under source control yet). Moving `src/foo/` into `/src/bar/`.
Answer yes to update references dialog.
A commit dialog shows this message: "Moved src\bar\foo\Client.hx to src\bar\foo\Client.hx" (note the old path is wrong - equal to the new path). If I answer `yes`, a similar dialog pops up for each of the source files. If I answer `no`, no more dialogs appear.

4. Move `src/bar/foo/` back to `src/` (assuming in the previous step I answered `yes` to all the commit dialogs, so all the files are now committed).
Answer yes to update references dialog.
A commit dialog shows this unexpected message: "Deleted src\bar\foo\Client.hx". If I answer `yes`, a similar dialog pops up for each of the source files. After I answer `yes` to all of them, an exception occurs in ProcessRunner, when it tries to run `git commit -m \"Deleted ...\"`. The `process.StartInfo.WorkingDirectory` is set to `"C:\\...\\src\\bar\\foo" - a directory which no longer exists since it was already moved into `src/`, therefore the "The directory name is invalid" exception from `process.Start()`.

5. Reset --hard the repo back to beginning (no more `src/foo/`, `src/bar/`).
[...]

PR #2195, opened 2018-07-10
RenameFile: Fix renaming that involves only a case change

### LINKED ISSUE #1288: Renaming class: avoid replace existing file warning if only case is changed (on Windows).

gene-pavlovsky:
Fixes #1288 and #2194 

One stone, two rabbits. Hopefully got both :)

### DIFF
--- a/External/Plugins/CodeRefactor/Commands/RenameFile.cs
+++ b/External/Plugins/CodeRefactor/Commands/RenameFile.cs
@@ -44,19 +44,20 @@ namespace CodeRefactor.Commands
                     return;
                 }
             }
+
+            string filePath = oldPath;
             // refactor failed or was refused
             if (Path.GetFileName(oldPath).Equals(newPath, StringComparison.OrdinalIgnoreCase))
             {
                 // name casing changed
-                string tmpPath = oldPath + "$renaming$";
-                File.Move(oldPath, tmpPath);
-                oldPath = tmpPath;
+                filePath += "$renaming$";
+                File.Move(oldPath, filePath);
             }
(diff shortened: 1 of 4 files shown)


External/Plugins/CodeRefactor/Commands/RenameFile.cs
```diff
@@ -44,19 +44,20 @@ protected override void ExecutionImplementation()
...
+                filePath += "$renaming$";
+                File.Move(oldPath, filePath);
             }
             if (!Path.IsPathRooted(newPath)) newPath = Path.Combine(Path.GetDirectoryName(oldPath), newPath);
 
             if (FileHelper.ConfirmOverwrite(newPath))
             {
-                FileHelper.ForceMove(oldPath, newPath);
+                FileHelper.ForceMove(filePath, newPath);
                 DocumentManager.MoveDocuments(oldPath, newPath);
```
gene-pavlovsky: The issue here was that `oldPath` contained `$renaming$`, so document manager couldn't find any matching documents, and the commit message generated by the source control dialog contained this unwanted `$renaming$`

gene-pavlovsky:
Covered the other cases.
I hope somebody could explain to me a scenario where a `Move` command might involve a rename that involves only a case change. I couldn't reproduce any of the 4 `$renaming$` cases in `Move` during my testing.

PR #2199, opened 2018-07-12
Add .gitattributes

### LINKED ISSUE #2197: Add .gitattributes file
It would be helpful to have a `.gitattributes` file that specifies line endings for files where that matters.
For example currently if the tests are not checked out using Windows line endings, they break, e.g.
`ParseFile_Comments` test:
>  Expected: "\r * Some custom comments\r "
> But was:  "\n * Some custom comments\n "

Example `.gitattributes` file which solves that problem for me:
```
* text=auto

*.cs text eol=crlf

*.sh text eol=lf
```

Of course if FD is ever developed on any other platforms than Windows, then it would be not nice to do this to `.cs` files and it would be better to actually make the tests more robust with regards to that issue.

gene-pavlovsky:
Closes #2197 

This file tells git to check out files with particular line endings, regardless of how they are stored in the repo, or the OS.

Sets the line endings to:
- CRLF for most VS files, C#, AS and Haxe source files
- LF for .sh scripts

P.S. These changes apply to newly cloned repos. To normalize line endings in an already cloned repo, according to the new `.gitattributes` file, use these commands (make sure you commit or stash your current work first):
```
git rm --cached -r . && git reset --hard
```

I followed [GitHub's instructions](https://help.github.com/articles/dealing-with-line-endings/) which resulted in 170+ modified files (normalized line endings).

Another reference: https://stackoverflow.com/a/34089670/4377192


elsassph:
Good idea!

gene-pavlovsky:
I followed [GitHub's instructions](https://help.github.com/articles/dealing-with-line-endings/) which resulted in 170+ modified files (normalized line endings).

Neverbirth:
I know that the change is huge, but from my side it's also good to merge. Also, solving conflicts for this kind of breaking change is simple as there is no more change than just the line ending.

SlavaRa (approved):

PR #2206, opened 2018-07-16
Haxe custom sdk refactor

gene-pavlovsky:
Resolves concerns raised here: https://github.com/fdorg/flashdevelop/pull/2189#pullrequestreview-136599155

Custom SDK cache was moved from Haxe Context to Haxe PluginMain, it's now a KeyValuePair instead of a full-blown dictionary since there's not much point to cache multiple custom SDKs.
All Haxeshim SDKs including a custom one are now re-validated when project is changed. Additionally, I've added a new OpenProjectProperties event to ProjectManager, Haxeshim SDKs are also re-validated on this event. So if you change the haxe version in `.haxerc` and open the project properties, the version will be updated.


External/Plugins/ProjectManager/Actions/BuildActions.cs
```diff
@@ -303,7 +303,7 @@ static public InstalledSDK MatchSDK(InstalledSDK[] sdks, string preferredSDK)
             string name = parts[parts.Length - 3];
             if (name != "")
                 foreach (InstalledSDK sdk in sdks)
-                    if (sdk.IsValid && sdk.Name == name)
+                    if (sdk.IsValid && ((name.StartsWith("Haxe Shim ") && sdk.IsHaxeShim) || sdk.Name == name))
```
SlavaRa:
```
name.StartsWith
```
->
```
name.StartsWithOrdinal
```

SlavaRa (approved):

PR #2664, opened 2019-01-28
[Haxe][FileParser] Added support for multiple inheritance

SlavaRa:
support_for_multiple_inheritance


markknol:
Does it also support the new `typedef D = A & B & {c: Int}` syntax, when using Haxe 4?

>`Type1 & Type2` is now a recognized syntax for intersection types. For the time being, it is only supported to merge structures (replacing the `{ >Type1, >Type2, }` syntax) and for type parameter constraints (replacing `T:(Type1, Type2)` which has been removed from the language).

Issue #1247, opened 2016-06-10
FlashDevelop: Trace overflow


sebthom:
After about 3000 lines in the output panel FlashDevelop stops displaying additional output and quits with "Trace overlow". Reading http://www.flashdevelop.org/community/viewtopic.php?f=6&t=5122&p=24013 I learned that this is a security issue, however for developing complex applications and esp. when doing unit testing etc. 3000 lines is not sufficient. The limit seems to be hardcoded but should be made configurable.

PR #2709, opened 2019-02-27
Support config max trace lines.

piaopolar:
Support config max trace lines.

see issues:
https://github.com/fdorg/flashdevelop/issues/1247
http://www.flashdevelop.org/community/viewtopic.php?p=25196

### DIFF
--- a/FlashDevelop/Settings/Accessors.cs
+++ b/FlashDevelop/Settings/Accessors.cs
@@ -608,6 +608,16 @@ namespace FlashDevelop.Settings
             set { this.maxRecentFiles = value; }
         }
 
+        [DefaultValue(1000)]
+        [DisplayName("Max Trace Lines")]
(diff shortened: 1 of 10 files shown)


FlashDevelop/Settings/Accessors.cs
```diff
@@ -608,6 +608,16 @@ public Int32 MaxRecentFiles
             set { this.maxRecentFiles = value; }
         }
 
+        [DefaultValue(1000)]
+        [DisplayName("Max Trace Lines")]
+        [LocalizedCategory("FlashDevelop.Category.Features")]
+        [LocalizedDescription("FlashDevelop.Description.MaxTraceLines")]
```
SlavaRa, line 614: You should add this key into localization files.

piaopolar: ok, updated.

PR #2737, opened 2019-03-19
[Haxe][Context] Avoid duplicate subscriptions in ExternalToolchain

R32:
It seems that every time you configure through the **PropertiesDialog** in ExternalToolchain, the `hxproj_ProjectUpdating` will be duplicate subscribed.

See the image below, I added `Trace.WriteLine` to reproduce it:
fd

### DIFF
--- a/External/Plugins/HaXeContext/ExternalToolchain.cs
+++ b/External/Plugins/HaXeContext/ExternalToolchain.cs
@@ -16,7 +16,7 @@ namespace HaXeContext
         static string projectPath;
         static WatcherEx watcher;
         static HaxeProject hxproj;
-        static bool skipSaveOnce;
+        static MonitorState monitorState;
         static System.Timers.Timer updater;
 
         internal static bool HandleProject(IProject project)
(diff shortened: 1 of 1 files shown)


External/Plugins/HaXeContext/ExternalToolchain.cs
```diff
@@ -175,24 +176,31 @@ internal static bool Clean(IProject project)
         /// <param name="project"></param>
         public static void Monitor(IProject project)
         {
-            if (updater == null)
+            if (!(project is HaxeProject pj))
             {
+                return;
+            }
+
+            if (updater == null) {
```
SlavaRa: ```if (updater == null) {```
->
```
if (updater == null)
{
```
or
```
if (updater is null)
{
```
please ;)

SlavaRa:
@R32 I'll check and merge if there are no problems.

R32:
I just used `rebase` with force to get it out of the wrong merge.

R32:
I hope I will not modify it anymore.

Issue #2763, opened 2019-04-18
FD is very slow when changing branches


gene-pavlovsky:
Working with a big project, if I switch to a really old branch, involving many changed files, FD freezes for quite a while - e.g. 30 seconds.
Running under MSVS, if I pause execution, every time the stack trace is at:
```
>	ASCompletion.dll!ASCompletion.Model.PathModel.DoScheduledOperations() Line 413	C#	Symbols loaded.
 	ASCompletion.dll!ASCompletion.Model.PathModel.updater_Tick(object sender, System.EventArgs e) Line 227	C#	Symbols loaded.
 	[External Code]		Annotated Frame
 	FlashDevelop.exe!FlashDevelop.MainForm.WndProc(ref System.Windows.Forms.Message m) Line 1779	C#	Symbols loaded.
 	[External Code]		Annotated Frame
 	FlashDevelop.exe!FlashDevelop.Program.RunFlashDevelopWithErrorHandling(string[] arguments, bool isFirst) Line 54	C#	Symbols loaded.
 	FlashDevelop.exe!FlashDevelop.Program.Main(string[] arguments) Line 21	C#	Symbols loaded.
```

Could this long process be somehow divided into chunks to avoid UI thread freezing, or something like that?

PR #2840, opened 2019-08-14
[Haxe][CodeComplete] Reduced the number of calls to File.Exists (#2763)

otani-koji:
When I researched the following issue with the VS2019 Performance Profiler, I found that calling `File.Exists` function takes time.
> https://github.com/fdorg/flashdevelop/issues/2763

So, I changed it to reduce this function calls by removing `foreach` expression by _toCheck.


SlavaRa:
thx

PR #2881, opened 2019-10-15
fix "Locate Active Document" in project tree view 

romamik:
Fix "Locate Active Document" in project tree view when paths differ by case (i.e. "c:\\" vs "C:\\")

In my case "Locate Active Document" was not working in hxml project with absolute paths starting with lower case "c:\".


romamik:
Oops...

PR #2883, opened 2019-10-16
[ProjectPanel] fix "Locate Active Document" in project tree view

romamik:
Fix "Locate Active Document" in project tree view when paths differ by case (i.e. "c:\" vs "C:\")

In my case "Locate Active Document" was not working in hxml project with absolute paths starting with lower case "c:".


SlavaRa:
How can I reproduce this BUG?

romamik:
[fd-test.zip](https://github.com/fdorg/flashdevelop/files/3734718/fd-test.zip)
unzip this to "c:\\"

PR #2894, opened 2019-11-05
Fixing progress bars not appearing

hachigoro:
Progress bars in the Find and Replace and the Find All References dialogs were not visible in the recent versions of FlashDevelop. Fixed that by making their declaration type the same as the initialization type, so they behave as a ProgressBarEx.


SlavaRa:
thx

PR #2896, opened 2019-11-07
Follow-up fix for invisible progress bars

hachigoro:
The previous fix did not work in all cases, leaving the progress bars not rendering. This should fix it properly, as using the ControlStyles.UserPaint style was unreliable, but catching the WM_PAINT message works well.


SlavaRa:
I have to test)

PR #2971, opened 2020-01-27
support haxe `final` keyword

### LINKED ISSUE #1925: [Haxe4] Support for `final` keyword
Haxe4 introduced a `final` keyword used in 2 places:
* Java-like, modifier to the methods. (Effect on autocomplete: Should not be shown in `override` completion.)
* Alternative for `var` keyword.

In-detail described here https://github.com/HaxeFoundation/haxe/pull/6596  
Feature already available in Haxe 4.0.0-preview.2 build.

### LINKED ISSUE #2775: [Haxe4] Auto-indent wrong after `final` keyword
Haxe code sample
```haxe
	function foo()
		if (true)
		{
			final i = 1;
			|
		}
```
Note: "|" is the caret position.

Start typing at the caret position, the code is auto-indented at the same level as `function`.
If `final i` is replaced with `var i` then indentation is correct. I guess it's related to the `final` keyword used in this context.

romamik:
Support haxe `final` keyword for variable completion and coloring. 
Resolves #2775 and partially resolves #1925 (does not support `final` as method modifier.)

### DIFF
--- a/External/Plugins/HaXeContext/Context.cs
+++ b/External/Plugins/HaXeContext/Context.cs
@@ -170,7 +170,7 @@ namespace HaXeContext
             features.importKey = "import";
             features.importKeyAlt = "using";
             features.varKey = "var";
-            //features.finalKey = "final";
+            features.finalKey = "final";
             features.overrideKey = "override";
             features.functionKey = "function";
             features.staticKey = "static";
--- a/External/Plugins/HaXeContext/Context.cs
+++ b/External/Plugins/HaXeContext/Context.cs
@@ -170,6 +170,7 @@ namespace HaXeContext
             features.importKey = "import";
             features.importKeyAlt = "using";
             features.varKey = "var";
+            //features.finalKey = "final";
             features.overrideKey = "override";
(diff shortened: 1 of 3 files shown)


SlavaRa:
@romamik should this work under haxe < 4.0.0?

romamik:
I did not test with older haxe versions. IMHO it should not break anything, but may display `final` in autocompletion list.

SlavaRa:
> I did not test with older haxe versions. IMHO it should not break anything, but may display `final` in autocompletion list.

I would not be so sure without the tests :)
This shouldn't work for versions < 4.0.0

romamik:
Do you mean to enable/disable this code depending on haxe version? Is there some example of how can this be accomplished?

romamik:
I've added version check. Now `final` is not shown in completion if haxe version is < 4. Also I've tested with haxe 3.4.7.

Not sure if the way I add/remove/check item existense in `features.codeKeywords` and `features.declKeywords` arrays is correct.

SlavaRa:
@romamik i'll check and merge today.

